PR #27, opened 2016-03-18
simplifying the app.

### LINKED ISSUE #1: Make the code better
The code isn't measurably bad, but there's a room for improvement. We care for the program to be beautifully crafted inside and out. The code needs to be more OOP, more modular, and maybe have fundamental changes to the way some objects and functions work. (Say we want to use the note buttons for other program easily. Currently, we can't do this because lack of modularity and OOP).

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/29433417-make-the-code-better?utm_campaign=plugin&utm_content=tracker%2F28290858&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F28290858&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #7: Fix animation when creating a new note
This only happens (In our tests) when scrollArea's scrollBar is visible.
So to repruduce this just create enough notes till scrollArea's scrollBar is visible, and then create a new note and you will see the problem. It's weird. It's annoying. Help us solve this.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/29433686-fix-animation-when-creating-a-new-note?utm_campaign=plugin&utm_content=tracker%2F28290858&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F28290858&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #24: Buttons icons not displayed on Linux
Icons are not showing in buttons on some Linux machines.

A collection of the machines reported that issue:

- Archlinux  64bits / Plasma 5.5
- Ubuntu 15.10, 64-bit 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/31780572-buttons-icons-not-displayed-on-linux?utm_campaign=plugin&utm_content=tracker%2F28290858&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F28290858&utm_medium=issues&utm_source=github).
</bountysource-plugin>

theshadowx:
The goal of these modifications is to make the app more efficient with minimum code and object/memory usage.

--> NoteData has its own class
     --> making members private
     --> setters / getters
     --> removing unnecessary layers
     --> isolated class : handles its own events from inside

--> Saving modification on text more efficiently
      --> no saving after each character added/deleted in the text to the database
      --> saving updated data locally
      --> saving data to database when another note is clicked or text being  put in search or quitting the app
     
--> Resizing App from the edge (some minimal issues still there but works fine)
--> Using Signal/slots instead of event filter
--> Using C++ 11 features
--> specifying a parent when creating an object so that no memory leak

This resolve #1, closes #7, closes #24

### DIFF
--- a/src/Notes.pro
+++ b/src/Notes.pro
@@ -13,9 +13,11 @@ TEMPLATE = app
 
 
 SOURCES += main.cpp\
-        mainwindow.cpp
+        mainwindow.cpp \
+    notedata.cpp
 
-HEADERS  += mainwindow.h
+HEADERS  += mainwindow.h \
+    notedata.h
 
 FORMS    += mainwindow.ui
 
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -9,7 +9,9 @@
 
 int main(int argc, char *argv[])
 {
+    QApplication::setDesktopSettingsAware(false);
     QApplication a(argc, argv);
+
     MainWindow w;
     w.show();
 
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -19,11 +19,15 @@
 MainWindow::MainWindow (QWidget *parent) :
     QMainWindow (parent),
     ui (new Ui::MainWindow),
+    m_notesDatabase(Q_NULLPTR),
+    m_trashDatabase(Q_NULLPTR),
+    m_settingsDatabase(Q_NULLPTR),
     m_noteWidgetsContainer(Q_NULLPTR),
+    m_clearButton(Q_NULLPTR),
     m_tempNote(Q_NULLPTR),
     m_currentSelectedNote(Q_NULLPTR),
     m_currentHoveredNote(Q_NULLPTR),
-    m_tempSelectedNoteBeforeSearching(Q_NULLPTR),
+    m_selectedNoteBeforeSearching(Q_NULLPTR),
     m_noteOnTopInTheLayout(Q_NULLPTR),
     m_canBeResized(false),
     m_resizeHorzTop(false),
@@ -44,7 +48,6 @@ MainWindow::MainWindow (QWidget *parent) :
     setupLine();
     setupRightFrame ();
     setupTitleBarButtons();
-    createMagnifyingGlassIcon();
     setupLineEdit();
     setupScrollArea();
     setupTextEdit();
@@ -116,7 +119,7 @@ void MainWindow::setupMainWindow ()
 */
 void MainWindow::setupKeyboardShortcuts ()
 {
-    new QShortcut(QKeySequence(Qt::CTRL + Qt::Key_N), this, SLOT(createNewNote()));
+    new QShortcut(QKeySequence(Qt::CTRL + Qt::Key_N), this, SLOT(createNewNoteWithAnimation()));
     new QShortcut(QKeySequence(Qt::CTRL + Qt::Key_Delete), this, SLOT(deleteSelectedNote()));
     new QShortcut(QKeySequence(Qt::CTRL + Qt::Key_F), ui->lineEdit, SLOT(setFocus()));
     new QShortcut(QKeySequence(Qt::CTRL + Qt::Key_E), ui->lineEdit, SLOT(clear()));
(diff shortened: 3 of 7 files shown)


nuttyartist:
You've done a terrific job by making the code much, much better.
I like the visual style and the technical changes you've done - mainly making the note a separate class, and using layouts on the notes list rather than fixed positions.

And a lot of other welcome changes and additions that will add to make the program work faster and more reliably, and for maintaining the code more easily for current and future contributors.

We really care about making the best app we can make. So we want it to look inside (code) as beautiful as the outside (GUI). This is a big step toward that goal. It also happens to close issue #1 

Yet, in order to merge this pull request, some problems that arise from your branch need to be solved:

1. Moving the splitter is very slow on a large number of notes, in contrast to the old approach
2. Why the delay at the start? (Also, the delay loading, at least on my large amount of notes looks ugly)
3. The separate lines in notes are not getting hidden
4. the scrollBars don't look the same. (Good or bad, we will need to talk about it)
5. I know there is a native clearButton for lineEdit but 1.it looks ugly. 2. I want it to look the same on different platforms
6. Pressing the arrow keys up/down in order to select notes isn't working well - look a mess when going deeply down or up.
7. Pressing 'Enter' after selecting a note with the arrow keys isn't selecting the not properly
8. When creating a new note there is no animation
9. When creating a new note and clicking on a different note the yellow highlight action is gone. (Things can't just disappear for no reason, it will make the feeling that something isn't quite right for the user)
10. When using the 'ctrl + arrow keys up/down' to navigate while a note's content is being focused on, the focus on textEdit should be gone.
11. the notes splitters should in the original master - they should align with lineEdit, and not start from the far left of the app.
12. When finishing with searching (clearing the lineEdit text) it doesn't go back to the last selected note. Or when searching and entering a note to get to it.

nuttyartist:
Great job.

Moving the splitter is still very slow on a large number of notes, but I'm going to create a different issue for that, and let this pull request be merged to benefit the master with the awesome changes you've made.

Only one thing to resolve before I merge this: the modification animation should make the note 'climb' to the top, so it gives the user a sense of what's happening. So when the note 'comes' to the top, it should be created and placed from the bottom-to-the-top - sliding up from the bottom to the top (And not the other way like the current approach).

We are getting closer and closer to version 1.0.0. When we will get there, we are going to have a very polished app. We want people to have a place to write down their thoughts and express themselves in an easy and beautiful way. We achieve that by creating a simple yet powerful and elegant note-taking app. Upon that mantra, we gonna build the next versions of Notes with features that will continue to reflect our views about making just that.

theshadowx:
For the splitter, it depends on the machine the app is running on,  I think it's better to put it as an issue. But I have idea that may make the movement of the splitter smoother, the idea is to use the [Qt Model-View pattern](http://doc.qt.io/qt-5/model-view-programming.html). so the NoteData Class will be divided on 2 classes:
 
- The model which will contain the note id, date, text and the title data, and will be derived from AbstractModelList
- a delegate class derived from QStyledItemDelegate which will be the widget containing the data

And instead of using QVBoxLayout and QScrollArea --> QListView

nuttyartist:
But won't using QListView kill are customization abilities?
Let's say we want to add an icon inside a note layout (Maybe when we'll do folders in the future). It's easily done in the current approach and many other things we may want to include in a note button , but can QListView give that flexibilty?

theshadowx:
you can customize the widget inside [QStyledItemDelegate](http://doc.qt.io/qt-5/qstyleditemdelegate.html) here is an [example](http://doc.qt.io/qt-5/qtwidgets-itemviews-stardelegate-example.html) where the cell painting is delegated to painting stars instead of a normal cell

theshadowx:
Now the insertion of note widget will move from the bottom to top

nuttyartist:
Let's continue the conversation of the splitter on a new issue.

Working well, merging it now.

peterchibunna:
I really want to be part of this development.

Issue #37, opened 2016-04-02
ScrollArea's scrollBar doesn't get updated...


nuttyartist:
ScrollArea's scrollBar doesn't get updated and hidden when number of notes goes to be less than what necessary for the srcollBar to be shown.

To reproduce this start with an empty database of notes. Open the app, create some notes till the scrollBar is shown, and then delete a note. The scrollBar is still shown. That's the problem.

PR #38, opened 2016-04-05
ScrollArea's scrollBar issue (#37) fixed

harishkrupo:
Description: Whenever a note is deleted, the scrollBar is either displayed or
hidden based on if the size of the list is smaller than scrollArea's height

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -1004,6 +1004,15 @@ void MainWindow::deleteNoteWithAnimation(NoteData *note, bool isFromUser)
 
         animation->start(QAbstractAnimation::DeleteWhenStopped);
     }
+
+    if(!m_visibleNotesList.isEmpty()){
+        int contentHeight = m_visibleNotesList[0]->height() * m_visibleNotesList.size();
+        if((contentHeight > ui->scrollArea->height()
+                && m_isScrollAreaScrollBarHidden)
+                ||(contentHeight < ui->scrollArea->height()
+                && !m_isScrollAreaScrollBarHidden))
+        setScrollAreaStyleSheet();
+    }
 }
 
 /**
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -837,6 +837,19 @@ void MainWindow::onLineEditTextChanged (const QString &keyword)
     }else{
         findNotesContain(keyword);
     }
+
+    if(!m_visibleNotesList.isEmpty()){
+        int contentHeight = m_visibleNotesList[0]->height() * m_visibleNotesList.size();
+        if((contentHeight > ui->scrollArea->height()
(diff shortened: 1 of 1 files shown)


theshadowx:
yes, it is aimed to behave like that

nuttyartist:
@theshadowx This pull request solves the issue for this branch. The scrollBar should be hidden when the height of all the notes is smaller than scrollArea, and visible when the height is bigger.

theshadowx:
oups, yes indeed, my mistake.

@harishkrupo my apologies. 

But it should be added in  : 

```
void MainWindow::deleteNote(NoteData *note, bool isFromUser)
```

not in the animation.

nuttyartist:
That's right.

@harishkrupo put it in

`void MainWindow::deleteNote(NoteData *note, bool isFromUser)
`
Like @theshadowx said.

And also, even when putting it there, there is a problem, if you create a bunch of notes and the scrollBar is showing but then you type in the search box something with no found notes, the scrollBar is still visible (When it shouldn't).

theshadowx:
I think the same lines that were added in deleteNote should be added in :
```
void MainWindow::findNotesContain(const QString& keyword)
```
here [#L1451](https://github.com/nuttyartist/notes/blob/master/src/mainwindow.cpp#L1451)

nuttyartist:
`void MainWindow::findNotesContain(const QString& keyword)
`

or in:

`void MainWindow::onLineEditTextChanged (const QString &keyword)
`

here  [#L824](https://github.com/nuttyartist/notes/blob/master/src/mainwindow.cpp#L824)
?

theshadowx:
yes, it's better the in [#L824](https://github.com/nuttyartist/notes/blob/master/src/mainwindow.cpp#L824) to manage the scrollbar in both states

nuttyartist:
Alright.

So @harishkrupo put your changes in both:

`void MainWindow::deleteNote(NoteData *note, bool isFromUser)`
`void MainWindow::onLineEditTextChanged (const QString &keyword)`

And I will merge this.

theshadowx:
@harishkrupo just that in 
```
void MainWindow::onLineEditTextChanged (const QString &keyword)
```
an **else** case should be added : 
```
if(!m_visibleNotesList.isEmpty()){
 \\ ... 
}else{
 \\ hide scrollbar
}
```

harishkrupo:
Why shouldn't it be in deleteNoteAnimation?

theshadowx:
because deleteNoteWithAnimation is not doing the deletion but just the animation, and deleteNote is the one responsible for deletion.
For example if I use deleteNote without animation and the update is in animation, the scrollbar visibility wouldn't be updated, however, if the update is in deleteNote the scrollbar visibility would be update no matter we use animation or not for deletion.

nuttyartist:
Works well. Merging,

PR #48, opened 2016-04-24
Add install target for Linux, removed stale resources file

dplanella:
- Added install target for Linux, leaving the `win32` and `macx` config as it was before. This allows running the standard `make install` command to install the files on a user's system, simplifying the Debian and snap packaging for Debian-based distros, and most probably also the packaging for rpm-based distros.
- Removed the stale `qrc_images.cpp` file. This is an autogenerated file created on each build by the `rcc` command and should not be tracked in revision control. The effect this created was that during a non-shadow build, `make` mistakenly detected that the file was already up-to-date and thus it wasn't regenerated. It then meant that an old file, without the fix for issue #24 was being used, and the icons were not being shown in the resulting binary.


theshadowx:
I think also these https://github.com/github/gitignore/blob/master/Qt.gitignore should be integrated in gitignore

nuttyartist:
Good this will solve a lot of hassle reported by users.
Also, before merging this, can the binary be generated into a separated "bin" folder?

@theshadowx isn't the current gitgnore cover everything?

theshadowx:
not all in the list needs to be added maybe just those below.
when not doing a shadow build these may be generated (I don't know the case of windows)
*.pro.user for the project configuration 
*.autosave  generated when QtCreator craches.

nuttyartist:
Oh ok, can you make a pull request?

theshadowx:
it can be added to this pull request #48 (in the commit : Removed autogenerated resources file, updated gitignore patterns)

dplanella:
Ok, I've added the suggested additional gitignore pattern.

Issue #6, opened 2015-12-28
Support unfocused window mode


nuttyartist:
When the window isn't in focus traffic light buttons should show kinda empty gray icon, currentSelectedNote highlight should turn gray and shadows should fade out. We haven't found a way to detect unfocused state to do these things, but you may know something we don't.

nuttyartist:
From my testing 

>     connect(qApp, SIGNAL(applicationStateChanged(Qt::ApplicationState)), this,&MainWindow::appStateChanged);

Should do the trick

Issue #45, opened 2016-04-22
Preventing multiple instances of Notes


theshadowx:
This feature will prevent data from corruption, so that data couldn't be removed by an instance and modified by another one.

Issue #30, opened 2016-04-02
When searching, we should be able to scroll throught a note content without being focused


nuttyartist:
When doing a search, we should be able to to scroll through the note's content without getting it focused.
It happens only when scrolling using clicking the scrollBar, scrolling with the mouse wheel works well.

PR #50, opened 2016-04-29
Fixing Issues #6 #30 #45

theshadowx:
This issues were fixes : #6 #30 #45 

Now: 
- Just one instance should be running. 
- The lineEdit (search bar) keeps Focus when searching and the scrollbar is moved by mouse. 
- Current row and Top left buttons are grayed when the application is inactive.

### DIFF
--- a/src/Notes.pro
+++ b/src/Notes.pro
@@ -4,7 +4,7 @@
 #
 #-------------------------------------------------
 
-QT       += core gui
+QT       += core gui network
 
 greaterThan(QT_MAJOR_VERSION, 4): QT += widgets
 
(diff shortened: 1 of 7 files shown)


nuttyartist:
#6 and #45 got solved but not #30.
When doing a search, selecting a long note that makes the scrollBar visible and then trying to scroll using the mouse by clicking on the scrollBar, it puts the focus on textEdit and so it clears the search session and takes you to the note, rather than just let you scroll and explore the note without being taken away.

theshadowx:
oups my mistake, I misread the issue. my mind went to the scrollbar of NoteView.  :(
Ok I'll fix it.
So what about the scrollbar of the listView?

nuttyartist:
The same, we should scroll the noteView while searching without getting taken away. But I haven't noticed any problem with that. What do you thought is the problem?

theshadowx:
when pressing the scrollbar of the noteview the lineEdit loses focus, but indeed the search mode is still actif,

nuttyartist:
Oh yea that's good you fixed that - the focus should still be on lineEdit.

nuttyartist:
Works well, merging.

PR #69, opened 2016-06-18
Integrate QSimpleUpdater with project

alex-spataru:
I have integrated the QSimpleUpdater library to your project. For the moment, I have enabled it on all platforms. To disable it on a specific platform, we can use some pre-processor macros in _main.cpp_.


theshadowx:
Instead of including the files, I think it would be better to add a submodule from your repo, so that when there is update done in the original repo, it will be easy to use it without copy/paste which may generate problems

alex-spataru:
I added the QSimpleUpdater as a submodule. However, the Travis CI build fails because it cannot add the Qt 5.5.1 repository.

src/main.cpp
```diff
@@ -5,22 +5,32 @@
 *********************************************************************************************/
 
 #include "mainwindow.h"
-#include <QApplication>
 #include "singleinstance.h"
 
+#include <QApplication>
+#include <QSimpleUpdater.h>
```
theshadowx: "QSimpleUpdater.h"

3rdParty/qxt/qxt.pri
```diff
@@ -0,0 +1,25 @@
...
+}
+
+macx {
+    SOURCES += $$PWD/qxtglobalshortcut_mac.cpp
+}
+
+unix:!macx {
+    LIBS += -lX11
+    SOURCES += $$PWD/qxtglobalshortcut_x11.cpp
+}
```
theshadowx: :+1:

theshadowx:
For travis, the problem doesn't come from your code, so just ignore it.

For appveyor, I think it's because it doesn't import the submodule

alex-spataru:
I updated the main.cpp to change:
`#include <QSimpleUpdater.h>`
to:
`#include "QSimpleUpdater.h"`

At least for Travis CI, I can modify the configuration to use Qt 5.2 (from the standard repositories). This should eliminate the build error, however, we would loose the ability to check if the project compiles with Qt 5.5 and Qt 5.6.

Unfortunately, I do not have any experience with Appveyor, so I can't give my opinion on it.

nuttyartist:
I think that in version 1.0.0 we should state the minimum requirement is Qt 5.6.
The good thing is that now we can virtually supply the most recent Qt library with our app across all major platforms - where all of them now has ways to package our app with it's dependencies. (Windows - dll's, Ubuntu - snap, Mac - self contain .app)

theshadowx:
The problem in travis is that the Qt5.6 was updated to Qt5.6.1:
the problem comes from here : https://github.com/nuttyartist/notes/blob/dev/.travis.yml#L11
```
- if [ "$QT_BASE" = "56" ]; then sudo add-apt-repository ppa:beineri/opt-qt56-trusty -y; fi
```

the line should be modified to
```
- if [ "$QT_BASE" = "56" ]; then sudo add-apt-repository ppa:beineri/opt-qt561-trusty -y; fi
```

nuttyartist:
@theshadowx I have [committed](https://github.com/nuttyartist/notes/commit/db12365bc2949817557f7e994a73c4b2d5a851e5) this now.

nuttyartist:
The error we get with AppVeyor is: 
> Cannot open include file: 'QSimpleUpdater.h': No such file or directory
And I checked and I get this error too when I try on compile it on your branch @alex-spataru.

theshadowx:
@alex-spataru @nuttyartist  I think the problem in appVeyor is that it doesn't upload the submodule that's why it doesn't find the file .h

in https://github.com/nuttyartist/notes/blob/dev/appveyor.yml#L38

this should be added:
```
cd C:\ci\notes\3rdParty\QSimpleUpdater
git submodule update --init
cd C:\ci\notes\src
```

nuttyartist:
I have committed the changes. :+1:

alex-spataru:
@theshadowx @nuttyartist Allright, I will reconfigure the submodule and try again. Thanks for the info :+1:

nuttyartist:
@alex-spataru When I'm downloading the app from your forked repo and try to compile it says: error: 

> QSimpleUpdater.h: No such file or directory
>  #include "QSimpleUpdater.h"
>                             ^

I guess because the files of the files of QSimpleUpdater are not in the folder. Is there something I should do that I missed?

alex-spataru:
@nuttyartist Can you try pulling the code from Git again? Maybe it was something that I messed up with the submodules.

nuttyartist:
@alex-spataru Yes, the problem still occurs. Just to make sure, that's the command I'm using:
> git clone https://github.com/alex-spataru/notes.git

theshadowx:
@nuttyartist here is the steps:

```
git clone https://github.com/alex-spataru/notes.git
cd notes/3rdParty/QSimpleUpdater/
git submodule update --init
```
And then it should work as usual

nuttyartist:
Oh thanks @theshadowx , now it's working.
But isn't that approach with the submodule cumbersome? Why not just include all the files? How much is the library going to change, is it so much that it will be a pain to copy and paste changes? And also what about changes to the library itself, if we will see that we need to change stuff for our own sake, how is it gonna affect the relationship with the upstream?

theshadowx:
you can do all the steps in on line

```
git clone --recursive https://github.com/alex-spataru/notes.git
```

There is a solution is to fork QSimpleUpdater and then submodule it in Notes.

nuttyartist:
@theshadowx That's more like it.
Well, I think we'll go with submodule QSimpleUpdater from alex repo as of right now. We'll change that if we'll see we need to. 

If we are to do it this way, I guess It we'll be appropriate to give more info about how to clone Notes that way in the Readme, right?

I'm merging this now.

alex-spataru:
Thanks for the merge! Just be sure to check the [UPDATES.json](https://github.com/nuttyartist/notes/blob/dev/UPDATES.json) file and specify the appropriate download links. Also, if you want to enable the integrated downloaded, just change [this line](https://github.com/nuttyartist/notes/blob/dev/src/main.cpp#L46).

PR #70, opened 2016-06-21
Add packaging scripts

alex-spataru:
I have added a packaging script to generate Debian packages easily. I have also added some scripts in the root directory to initialize the submodules.


src/Notes.pro
```diff
@@ -41,16 +41,15 @@ HEADERS  += \
...
-    isEmpty (PREFIX) {
-        PREFIX = /usr
-    }
-
-    BINDIR = $$PREFIX/bin
-    DATADIR =$$PREFIX/share
-
-    target.path = $$BINDIR
-    INSTALLS += target
+linux:!android {
```
theshadowx: Not good, you should give the user the choice to install the app wherever he/she wants.

theshadowx: Create a symlink from $$Prefix/bin/notes to /usr/bin/notes when PREFIX is not /usr

src/Notes.pro
```diff
@@ -41,16 +41,15 @@ HEADERS  += \
...
+    target.path = /usr/bin
+    icon.path = /usr/share/pixmaps
+    desktop.path = /usr/share/applications
+    icon.files += $$PWD/packaging/linux/common/notes.png
+    desktop.files += $$PWD/packaging/linux/common/notes.desktop
+
+    TARGET = notes
+    INSTALLS += target desktop icon
 }
```
theshadowx: Not good, you should give the user the choice to install the app wherever he/she wants.
Create a symlink from $$PREFIX/bin/notes to /usr/bin/notes when PREFIX is not /usr

alex-spataru: Allright, I will restore the previous Linux configuration in the `qmake` file.

alex-spataru: @theshadowx I restored the project configuration and removed the init scripts.

nuttyartist:
When trying to run the make_deb.sh I get this error:

> dpkg-checkbuilddeps: error: Unmet build dependencies: debhelper (>= 9) qtbase5-dev (>= 5.2) qtbase5-dev-tools (>= 5.2) qt5-qmake qtbase5-private-dev

I have tried on both eOS Freys (14.04) and on Ubuntu 16.04.

alex-spataru:
@nuttyartist This has something to do with the Debian control file, in which I added `qtbase5-private-dev` to the build dependencies. Did you try to install those packages before attempting to build the .deb file?

nuttyartist:
Oh ok, I installed `qtbase5-private-dev` and now the error is gone. Why did you put it there?

There is a new error that occurs when compiling:
```
fatal error: QSimpleUpdater.h: No such file or directory
```
I guess because it doesn't downloaded the submodule?

theshadowx:
yep

```
git submodule update --init
```

alex-spataru:
@nuttyartist I couldn't get Notes to compile without that package. It seems that `qxt` needs it in order for a successful compilation.

theshadowx:
yes it needed it

nuttyartist:
@alex-spataru ah ok. Can you update the file with the `git submodule update --init`?

alex-spataru:
@nuttyartist Which file should I update?

nuttyartist:
The `make_deb.sh` file I think.

alex-spataru:
@nuttyartist The sub-module only affects the `QSimpleUpdater` folder, every other file is part of the Notes repository.

nuttyartist:
When you compile the app with `qmake -qt5 ../../../../../src/Notes.pro` the `QSimpleUpdater` isn't there, so we need to download it in some way. I'm not quite experience in that.
@theshadowx What do you say?

alex-spataru:
Ok, I understand you now! I will update the script right away.

nuttyartist:
Great! Working good. Finally, an automatic script to create .deb packages. It will save a lot of hassle and will allow us to deploy faster. Merging.

PR #77, opened 2016-07-07
Add support for generating snap packages

dplanella:
This branch adds support for:

- Generating a snap package as part of a build script, automatically bumping its version based on the latest git commit id:

    ```
cd src
git submodule update --init
qmake
make snap
    ```
- Alternatively generating a snap package manually

    ```
cd linux/packaging/snap
snapcraft
    ```

Notes:
- At least on Ubuntu, the mouse pointer changes to a different one from the rest of the system when hovering over the app. I've not yet figured out what causes it, but I consirer it a minor issue that does not affect functionality.
- Unlike the v0.9.0 tag, at least on Ubuntu, the `dev` branch seems to create an entry on the indicators area that:
  - Does not close when the app is closed (i.e. remains in the indicators area)
  - Gets added as a new entry every time the app is started
  - Does not show any icons or entries
  - Prevents the app from fully closing. I.e. launch `notes` on the terminal, watch the output, close the app with the (X) button, notice the app still holding onto its output on the terminal. Ctrl + C to exit it and the indicator.

notes-notification


dplanella:
Note that (I believe unrelated to this PR), I'm getting the following error trying to run `qmake`:

```
$ qmake
Cannot read /home/dpm/dev/snaptest/notes/3rdParty/QSimpleUpdater/QSimpleUpdater.pri: No such file or directory
Project MESSAGE: This project is using private headers and will therefore be tied to this specific Qt module build version.
Project MESSAGE: Running this project against other versions of the Qt modules may crash at any arbitrary point.
Project MESSAGE: This is not a bug, but a result of using Qt internals. You have been warned!
```

it seems that the `3rdParty/QSimpleUpdater/QSimpleUpdater.pri` is not committed to the source tree? And why is the project using private headers?

alex-spataru:
@dplanella QSimpleUpdater is linked to this repository as a submodule. Use `git submodule update --init` to get the contents for the QSimpleUpdater.

dplanella:
@alex-spataru ah, thanks, that worked. That should probably be mentioned in the top-level README.md file for contributors as well, I think.

nuttyartist:
Awesome, creating the snap is working well, but I can't get the app running. I get this error when I'm trying from terminal: (When trying from the launcher the app starts and quite (because of the error))
`/snap/notes/x1/bin/desktop-launch: line 142: exec: Notes: not found`

Also, can you change the icon in `/notes/packaging/linux/snap/setup/gui/notes_icon.png` to the new icon from `notes/src/images/notes_icon.png`?

dplanella:
@nuttyartist:
- What command did you actually use to run the app from the terminal? I'd expect simply calling `notes` should work
- If you happen to be running Ubuntu, or a distro which already supports snapd, does the version from the store work for you? You can either:
```
    # Just refresh the notes app to replace the sideloaded version by the one from the Store 
    sudo snap refresh notes

    # Alternatively, manually delete the sideloaded version first and install from the store
   sudo snap remove notes
   sudo snap install notes
```

I'll update the icon, sure!

nuttyartist:
The one from the store (sudo snap install notes) works well, except for the icon and for some reason the fonts are not the rendered correctly, it (seems) to use "Ubuntu" font rather than "Liberation Sans", Is that happens on your machine too? 

To run the sideloaded version I used `notes` from the terminal.
And to build the snap I used `snapcraft`, which seems like you just deleted from the build documentation. Can't that be the cause?

dplanella:
Addressing issues bit by bit, I've pushed a change to the PR that should make the sideloaded app to work from the CLI again, will address the icon and desktop file updates next

dplanella:
Ok, to address the comments:
- The app now starts correctly after being sideloaded
- The icon and .desktop file were updated
- I'm not sure what the actual issue is with the font. I cannot tell if it's Liberation Sans, but it's certainly not the Ubuntu font, if I'm not mistaken.
- I've noticed an issue with app indicators that seems to affect the dev branch only. I can't tell if it was already there or if the snap prevents from it working correctly.

nuttyartist:
The app is now running correctly, but there's no icon at all (not in the launcher, systray, or the task bar). And yes the systray (app indicator) doesn't working - isn't showing the menu correctly. 

I checked on the version that currently on the store and it has no entry in the indicators panel at all because in that version (0.9.0) we didn't had it, it's for 1.0.0. So we do need to fix the indicator problem for 1.0.0, but for now to upload to the store the only things needs to be done is the icon and the fonts. I get the same fonts as in the screenshot you sent, and it's not supposed to be that way.

dplanella:
Hm, icon is working here, but I'll double-check on another computer. Regarding the font... where do you define it in the code?

nuttyartist:
So to sum up:

For the v0.9.0 tag that's going to the Ubuntu store, this needs to be done:
- Update icon to the new one.
- Fonts issues.

For the dev branch, we need to solve the systray icon problem where:
- The icon isn't showing.
- The menu entries aren't showing.
- Gets added as a new entry every time the app is started.

(About the other problems you mentioned relating to the app not closing - it's supposed to be that way - that clicking on 'x' is just hiding the app, and you quit the app using the indicator)

nuttyartist:
@dplanella [Here](https://github.com/nuttyartist/notes/blob/dev/src/notewidgetdelegate.cpp#L11-12) and [here](https://github.com/nuttyartist/notes/blob/dev/src/mainwindow.cpp#L431).

packaging/linux/snap/snapcraft.yaml
```diff
@@ -23,6 +23,7 @@ parts:
     build-packages:
       - qtbase5-dev
       - qtbase5-dev-tools
+      - qtbase5-private-dev
```
nuttyartist: Oh yea, that one is required, good notice!

dplanella:
Issues fixed:
- Update icon to the new one, get it to show up at all.
- Fonts issues.

nuttyartist:
All-right, the icon is showing up, and the fonts issue is fixed (:

I don't see the app in the launcher for some reason but I think the problem is on my end, does it happen to you too?
Also the indicator still has no icon and no menu entries.

dplanella:
The icon appears on the launcher for me, as well as in the Dash, but I'll try a fresh install on my other PC and report back.

The appindicator issues are still there and I'll look at them either tomorrow or early next week. Nearly there! :)

nuttyartist:
Yes truly nearly there! 

I created a deb from the dev branch and the indicator is working as it should, so I think the issues got do to something with the snap.

dplanella:
Ok, thanks for checking, will investigate the indicator issues then!

nuttyartist:
Ok, thanks!

dplanella:
After investigation, I filed some snappy bugs to address the issues:

- https://bugs.launchpad.net/ubuntu/+source/snapd/+bug/1600136
- https://bugs.launchpad.net/ubuntu/+source/snapd/+bug/1600138
- https://bugs.launchpad.net/ubuntu/+source/snapd/+bug/1600140

nuttyartist:
Alrighty.
This pull request does what it says well, it's generating a snap for us.
I will open a new issue about the problems with the indicator and we'll wait for these to be solved.
Thanks for the good work and the tedious debugging @dplanella!
Merging it is.

PR #84, opened 2016-08-08
Add QSimpleUpdater statically

alex-spataru:
I removed the QSimpleUpdater module and added all its files statically, so that the project can customize the workings of the library (as requested in Gitter).


nuttyartist:
Thanks!

Issue #13, opened 2015-12-28
Create automatic packaging for .deb, .rpm, snap and flatpack


nuttyartist:
We want to create the packages through a simple 'run that script and it will create your packages' kind of way. For .deb, .rpm and snap.

- [x] .deb script
- [x] .snap script
- [ ] .rpm script
- [ ] .flatpack script
- [x] Main Automatic Script

theshadowx:
here is an interesting project that do packaging for linux OSes

https://github.com/probonopd/AppImageKit

nuttyartist:
Oh yeah, I heard of it, it's a very cool thing.

The main drawback I can see is that after downloading the appimage people needs to allow it to be executable, which require, on most of the distros right-clicking ->properties->allow to execute (something like that) but on some distros it's different (eOS), or they need to use the command line.
Both of the options are not as end-user friendly as just clicking on a .deb file and just install.

Second thing is that using .deb, .rpm and snap,and uploading them to the official repository will allow our app to be discovered by people using software center etc...

A solution for the first claim might be to put a nice explanation on the website on the download page.
Also automatic updates are very important, and I saw they have a nice support for that one, which is good.

What will be a good indicator for choosing between all of the options is the end result of what the user will experience. If in the end, it will be easy and intuitive for him to download and install, and without a hassle get automatic updates, then that's could tell it's a good thing. It should also be easy and intuitive and automatic for us - the developers to package and iterate the app.

alex-spataru:
Well, for one of my projects I wrote scripts to generate _.deb_ and _.snap_ packages. Unfortunately, I am clueless about how to generate _.rpm_ packages.

You can find the deploy scripts [here](https://github.com/WinT-3794/QDriverStation/tree/master/etc/deploy/linux).

nuttyartist:
@dplanella is already working on a snap packaging [here](https://github.com/dplanella/notes/tree/snappy/packaging/linux/snap). Maybe you can contribute there if you see areas you can help.

We have a *debian* folder, can you write a script to package it?

The goal is to have one main script to be run, that will make all the different packages automatically by running their own scripts. So when the specific scripts will be ready we'll create the main script.

alex-spataru:
Yes, I will create a new pull request with the debian package script.

alex-spataru:
@nuttyartist Done!

nuttyartist:
@dplanella doesn't seem to progress on the .snap script.

[...]

PR #85, opened 2016-08-14
Add general packaging script

alex-spataru:
I have created a simple packaging script that creates both Snap and Debian packages (as discussed in issue #13).


nuttyartist:
Great, thanks! merging...

PR #98, opened 2016-10-22
Automatic updates windows

alex-spataru:
Hi,

I have implemented are the features that you requested in [this pull request](https://github.com/nuttyartist/notes/pull/97), including the integrated downloader. Also, I have managed to fix the issue with the HTML redirections, which allows users to download updates directly from the files stored in GitHub releases.

All Best,
Alex Spataru


nuttyartist:
Can't wait to check it out. 
For some reason, the updater isn't showing. I have putted a `qDebugger` in `void UpdaterWindow::checkForUpdates()` and it doesn't get called.
I also did `app.setApplicationVersion ("0.8.0");` for the updater to infer changes, yet, the updater doesn't get called.

alex-spataru:
I'll check again the code, it could be that the project does not link to the SSL libraries under Linux.

nuttyartist:
Cool! Now the dialog is showing. I will test it extensively in the following days and report back.

alex-spataru:
All right! Thanks!

PR #104, opened 2016-11-19
Add integrated downloader & apply stylesheets

alex-spataru:
Hi,

This PR implements the following changes:
- Add integrated downloader
- Save partial downloads to disk (not on RAM)
- Apply stylesheets to the updater dialog to make it look like the mockups in Gitter

Notes:
- The stylesheet uses pixels instead of points to define font sizes
- The stylesheet is loaded from the application resources, not from the UIC


nuttyartist:
Looking good!

- We should use the Qt::PointingHandCursor flag upon hovering the "Close" and "Update" buttons.
- Also, when pressing the buttons they kinda move a bit, they should stay still.

It needs a bit of playing around with the spacing between the elements and a bit of font change to look more like in the mockup, how do I do this? How to change the font size of the labels (changing in the updaterwindow.ui form doesn't take effect)? and the spacing?

Overall looks good, and the downloader is more responsive, but I get this error when it says "Opening downloaded file" 
`gvfs-open: file:///home/user/Downloads/notes_0.9.0_amd64.deb: error opening location: No application is registered as handling this file`

theshadowx:
@nuttyartist I think you have this error because there is no software linked to .deb, so the updater can't open the file.

In ubuntu the updater opens software library.

alex-spataru:
I have worked on the layout spacings and modified the font weight in some of the labels, here is the result:

Screenshot

nuttyartist:
Better!

@theshadowx you're right! forgot that Elementary dropped official support for .deb. 
@alex-spataru can you detect if there is an error and put a message below the timeLabel?
"There is an error while trying to open the file'

alex-spataru:
@nuttyartist Of course, I will also check if I can open a file manager pointing to the downloaded file (when there is no default application for opening the downloaded file).

nuttyartist:
@alex-spataru Oh great!

alex-spataru:
I added the code to notify the user when the download cannot be opened. 

At least for me, the cursor icon returns to "arrow" upon hovering a button (I performed tests on macOS and Windows). Tomorrow I will test the behavior of the cursor icons on Ubuntu.

nuttyartist:
@alex-spataru It shouldn't be an arrow. It should be hand-pointing, the same way it's in buttons like the newNoteButton and trashButton.

The notification works well, can you do that it opens up the folder where the downloaded file is?

nuttyartist:
@alex-spataru What's left is only opening the folder of the downloaded file if it fails to open it. And the things with the buttons. Then we can merge this.

alex-spataru:
All right, I will work on it tomorrow morning and notify you through here.

alex-spataru:
I have changed the cursors and implemented the code to open the folder where the update is saved (when the dialog cannot open the file automatically). Let me know if it works for you too!

nuttyartist:
@alex-spataru It doesn't work for me.
It says "Opening downlaoded file..." but it doesn't do anything.
In the Output window it says: "gvfs-open: file:///home/user/Downloads/notes_0.9.0_amd64.deb: error opening location: No application is registered as handling this file"

alex-spataru:
@nuttyartist I have updated the code to open the download folder when Notes cannot open the update file. Could you verify that this code works for you too?

nuttyartist:
Oh I haven't seen your message! I will check it now. @alex-spataru

nuttyartist:
Still the same error:

gvfs-open: file:///home/user/Downloads/notes_0.9.0_amd64.deb: error opening location: No application is registered as handling this file

alex-spataru:
@nuttyartist No worries! I'll check the documentation of `QDir` to see if I can find an alternative way to get the folder path from a given file name.

nuttyartist:
@alex-spataru Alrighty, thanks.

nuttyartist:
@alex-spataru I've tried to fix that too but haven't found a solution. Can you give it another shot again? Only this PR and a few little things to finish the 1.0.0 version.

src/updaterwindow.cpp
```diff
@@ -311,8 +311,17 @@ void UpdaterWindow::openDownload (const QString& file)
...
 
-        QUrl folderPath = QUrl::fromLocalFile (QDir (new_file).absolutePath());
-        QDesktopServices::openUrl (folderPath);
+        /* Obtain folder list */
+        QStringList directories = QDir (new_file).absolutePath().split ("/");
+        directories.removeLast();
+
+        /* Generate file path */
+        QString url;
+        foreach (QString string, directories)
```
theshadowx: Instead for loop : 
```
QString url = QStringLiteral("/") + directories.join('/');
```

nuttyartist:
Still not opening the folder for me, same error. Is it working for some of you?

alex-spataru:
I think I found the issue. In my Ubuntu test configuration, `QDesktopServices::openUrl()` always returns true, even if the system failed to open the file. Basically, this means that my implementation 'thinks' that the system opened the file correctly. Therefore, no corrective action is taken by the updater window.

I will investigate more about this issue later tonight.

nuttyartist:
Cool, it's working!
Really like the dialog message, it's great.

I will later check if the updater works on Windows and then merge it.

nuttyartist:
@alex-spataru Just checked this on Windows. 2 things:
- It seems to open 2 installer windows when opening the downloaded file.
- We should close Notes when opening the file.

theshadowx:
@nuttyartist  I don't think it's possible to close notes when the file is opened from another process.

I have noticed that the installer opens two folder browsers also in Ubuntu.

nuttyartist:
@theshadowx What do you mean by another process? I mean Notes should close itself automatically after it opened the file.

We should investigate that.

theshadowx:
What if the user didn't install the update from the file, should the app be closed although no update was done.

I mean by process, is that the app can't know if really it was updated, as there is no link between the installer and the app.

nuttyartist:
@theshadowx I didn't quite get you, could you elaborate again?
By the way, the installer on Windows is already detecting if the app is running and recommend to choose the option of closing it automatically before installing. But I want the user to be the least confused as possible, so I want us to take the action and close it automatically after opening the installer.

alex-spataru:
@theshadowx @nuttyartist I think that it is possible to do this using QProcess (by running it as a detached process), I will try to implement this later tonight.

theshadowx:
@alex-spataru Yes indeed, I forgot about it. :+1:

src/updaterwindow.cpp
```diff
@@ -19,6 +19,19 @@
...
+#else
+  #ifdef USE_XDG_OPEN
+    #undef USE_XDG_OPEN
+  #endif
+#endif
+
+#ifdef USE_XDG_OPEN
+  #include <QProcess>
+  static QProcess XDGOPEN_PROCESS;
+#endif
```
theshadowx: Better use camelCase to not be confused as it's not a macro

alex-spataru:
@nuttyartist @theshadowx I have added the code to exit Notes when the update file is opened, could you corroborate that it works?

nuttyartist:
@alex-spataru Sure, going to lay this day for working on Notes and preparing v1.0.0.

nuttyartist:
If we're going to ask the user to quit the app before istalling the update, it's better to open the installer only after the user clicked on "Yes". If the user clicked on "No" then the installer shouldn't open. Right now, it's asking the user and _at the same time_ open the installer.

I would suggest that the right behaviour would be to open the installer and close Notes automatically.
The user already pressed "update" with the intention to do so (he could also cancel that beforehand).

Also, we got to figure why it opens up the installer twice.

alex-spataru:
@nuttyartist Thanks for the feedback, I will remove the message box and just close Notes automatically. Then, I will work on why the installer is opened twice.

nuttyartist:
@alex-spataru Great.

nuttyartist:
@alex-spataru Hey Alex, can you solve the latest issue (installer opening twice)?

alex-spataru:
@nuttyartist I will check the issue tomorrow and notify you about any progress

nuttyartist:
@alex-spataru Alrighty.

nuttyartist:
@alex-spataru It doesn't seem to work for me.

nuttyartist:
Ok, we finally got to this point. haha
Awesome work Alex and Ali.
Merging...

PR #112, opened 2017-02-15
feature : Application updater

theshadowx:
Check if there is new version of the application.
Show update window to let the user decide whether to update the application or not.
If decides to update : 

- Download the installer
- Open the software center to install the app
- If app can't be opened, just show the download folder containing the installer.


nuttyartist:
:)

PR #117, opened 2017-02-23
Adding Collapse/Expand of note list view

theshadowx:


nuttyartist:
Very good, like that the text in the menu changes as the state changes. Just be sure to remove the "side" in "Show note list side" and "Hide note list side" and change them to "Show notes list" and "Hide notes list" and I will merge.

theshadowx:
ok

PR #120, opened 2017-03-18
Import/Export Notes

geoffschoeman:
Added the import/export notes feature.

### DIFF
--- a/src/dbmanager.cpp
+++ b/src/dbmanager.cpp
@@ -81,7 +81,7 @@ NoteData* DBManager::getNote(QString id) {
         note->setFullTitle(fullTitle);
         return note;
     }
-    return NULL;
+    return Q_NULLPTR;
 }
 
 NoteExport* DBManager::getNoteExport(QString id) {
@@ -108,7 +108,7 @@ NoteExport* DBManager::getNoteExport(QString id) {
         note->fullTitle = fullTitle;
         return note;
     }
-    return NULL;
+    return Q_NULLPTR;
 }
 
 
@@ -365,10 +365,10 @@ NoteExport DBManager::buildNote(QSqlQuery query) {
     return note;
 }
 
-void DBManager::importNote(NoteExport noteExport) {
+void DBManager::importNote(const NoteExport& noteExport) {
     NoteExport* note = getNoteExport(noteExport.id);
     bool isNew = false;
-    if (note == NULL) {
+    if (note == Q_NULLPTR) {
         // Note doesn't exist, create it
         note = new NoteExport();
         note->id = noteExport.id;
@@ -399,4 +399,5 @@ void DBManager::importNote(NoteExport noteExport) {
     } else {
         updateImportedNote(note);
     }
+    delete note;
 }
--- a/src/dbmanager.cpp
+++ b/src/dbmanager.cpp
@@ -84,32 +84,32 @@ NoteData* DBManager::getNote(QString id) {
     return Q_NULLPTR;
 }
 
-NoteExport* DBManager::getNoteExport(QString id) {
-    QSqlQuery query;
+//NoteExport* DBManager::getNoteExport(QString id) {
+//    QSqlQuery query;
 
-    int parsedId = id.split('_')[1].toInt();
-    QString queryStr = QStringLiteral("SELECT * FROM active_notes WHERE id = %1 LIMIT 1").arg(parsedId);
-    query.exec(queryStr);
+//    int parsedId = id.split('_')[1].toInt();
+//    QString queryStr = QStringLiteral("SELECT * FROM active_notes WHERE id = %1 LIMIT 1").arg(parsedId);
+//    query.exec(queryStr);
 
-    if (query.first()) {
-        NoteExport* note = new NoteExport();
-        int id =  query.value(0).toInt();
-        qint64 epochDateTimeCreation = query.value(1).toLongLong();
-        QDateTime dateTimeCreation = QDateTime::fromMSecsSinceEpoch(epochDateTimeCreation, QTimeZone::systemTimeZone());
-        qint64 epochDateTimeModification= query.value(2).toLongLong();
-        QDateTime dateTimeModification = QDateTime::fromMSecsSinceEpoch(epochDateTimeModification, QTimeZone::systemTimeZone());
-        QString content = query.value(4).toString();
-        QString fullTitle = query.value(5).toString();
+//    if (query.first()) {
+//        NoteExport* note = new NoteExport();
+//        int id =  query.value(0).toInt();
+//        qint64 epochDateTimeCreation = query.value(1).toLongLong();
+//        QDateTime dateTimeCreation = QDateTime::fromMSecsSinceEpoch(epochDateTimeCreation, QTimeZone::systemTimeZone());
+//        qint64 epochDateTimeModification= query.value(2).toLongLong();
+//        QDateTime dateTimeModification = QDateTime::fromMSecsSinceEpoch(epochDateTimeModification, QTimeZone::systemTimeZone());
+//        QString content = query.value(4).toString();
+//        QString fullTitle = query.value(5).toString();
 
-        note->id = QStringLiteral("noteID_%1").arg(id);
-        note->creationDateTime = dateTimeCreation;
-        note->lastModificationDateTime = dateTimeModification;
-        note->content = content;
-        note->fullTitle = fullTitle;
-        return note;
-    }
-    return Q_NULLPTR;
-}
+//        note->id = QStringLiteral("noteID_%1").arg(id);
+//        note->creationDateTime = dateTimeCreation;
+//        note->lastModificationDateTime = dateTimeModification;
+//        note->content = content;
+//        note->fullTitle = fullTitle;
+//        return note;
+//    }
+//    return Q_NULLPTR;
+//}
 
 
 QList<NoteData *> DBManager::getAllNotes()
@@ -170,42 +170,42 @@ bool DBManager::addNote(NoteData* note)
 
 }
 
-bool DBManager::addImportedNote(NoteExport* note) {
-    QSqlQuery query;
-    QString emptyStr;
-    qint64 epochTimeDateCreated = note->creationDateTime.toMSecsSinceEpoch();
-    qint64 epochTimeDateModified = note->lastModificationDateTime.toMSecsSinceEpoch();
-    QString content = note->content.replace("'","''").replace(QChar('\x0'), emptyStr);
-    QString fullTitle = note->fullTitle.replace("'","''").replace(QChar('\x0'), emptyStr);
-    QString queryStr = QString("INSERT INTO active_notes (creation_date, modification_date, deletion_date, content, full_title) "
-                               "VALUES (%1, %2, -1, '%3', '%4');")
-            .arg(epochTimeDateCreated)
-            .arg(epochTimeDateModified)
-            .arg(content)
-            .arg(fullTitle);
+//bool DBManager::addImportedNote(NoteExport* note) {
+//    QSqlQuery query;
+//    QString emptyStr;
+//    qint64 epochTimeDateCreated = note->creationDateTime.toMSecsSinceEpoch();
+//    qint64 epochTimeDateModified = note->lastModificationDateTime.toMSecsSinceEpoch();
+//    QString content = note->content.replace("'","''").replace(QChar('\x0'), emptyStr);
+//    QString fullTitle = note->fullTitle.replace("'","''").replace(QChar('\x0'), emptyStr);
+//    QString queryStr = QString("INSERT INTO active_notes (creation_date, modification_date, deletion_date, content, full_title) "
+//                               "VALUES (%1, %2, -1, '%3', '%4');")
+//            .arg(epochTimeDateCreated)
+//            .arg(epochTimeDateModified)
+//            .arg(content)
+//            .arg(fullTitle);
 
-    query.exec(queryStr);
-    return (query.numRowsAffected() == 1);
+//    query.exec(queryStr);
+//    return (query.numRowsAffected() == 1);
 
-}
+//}
 
-bool DBManager::updateImportedNote(NoteExport* note) {
-    QSqlQuery query;
-    QString emptyStr;
-    int id = note->id.split('_')[1].toInt();
-    qint64 epochTimeDateModified = note->lastModificationDateTime.toMSecsSinceEpoch();
-    QString content = note->content.replace("'","''").replace(QChar('\x0'), emptyStr);
-    QString fullTitle = note->fullTitle.replace("'","''").replace(QChar('\x0'), emptyStr);
-    QString queryStr = QStringLiteral("UPDATE active_notes "
-                                      "SET modification_date=%1, content='%2', full_title='%3' "
-                                      "WHERE id=%4")
-                       .arg(epochTimeDateModified)
-                       .arg(content)
-                       .arg(fullTitle)
-                       .arg(id);
-    query.exec(queryStr);
-    return (query.numRowsAffected() == 1);
-}
+//bool DBManager::updateImportedNote(NoteExport* note) {
+//    QSqlQuery query;
+//    QString emptyStr;
+//    int id = note->id.split('_')[1].toInt();
+//    qint64 epochTimeDateModified = note->lastModificationDateTime.toMSecsSinceEpoch();
+//    QString content = note->content.replace("'","''").replace(QChar('\x0'), emptyStr);
+//    QString fullTitle = note->fullTitle.replace("'","''").replace(QChar('\x0'), emptyStr);
+//    QString queryStr = QStringLiteral("UPDATE active_notes "
+//                                      "SET modification_date=%1, content='%2', full_title='%3' "
+//                                      "WHERE id=%4")
+//                       .arg(epochTimeDateModified)
+//                       .arg(content)
+//                       .arg(fullTitle)
+//                       .arg(id);
+//    query.exec(queryStr);
+//    return (query.numRowsAffected() == 1);
+//}
 
 bool DBManager::removeNote(NoteData* note)
 {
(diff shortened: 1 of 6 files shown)


theshadowx:
Please set the version to Qt5.6 for QDataStream

nuttyartist:
Hey, thanks! Looks very cool. A few things are on my mind:

- This feature is critical for it to work properly. Is there a way I could test If it backed up and migrated all my notes successfully (I have about 700 notes)?
- We should have a loader progress bar when importing the notes, it took quite some time to import my notes, and people should be able to get a clear feedback on the state of the app, so they won't have to wonder if the app has frozen or is it just loading. A simple progress bar like @theshadowx have done [here](https://github.com/nuttyartist/notes/blob/dev/src/mainwindow.cpp#L94) would be good. Also, on my machine, the os (ubuntu) even thinks the app is frozen for a few seconds till it successfully loaded. Is there a more stable way to load?
- Can you explain more about the notes.nbk, what does it contain? Also, is it contain also the settings.ini file? (I like that it has special file ending).
- I get a lot of "QSqlQuery::value: not positioned on a valid record" when imparting the notes. Is it something to worry about?

Thanks again for your implementation of this, this is a feature a lot of people have been asking, and I myself wanted, so it will surely benefit the app.

theshadowx:
> QSqlQuery::value: not positioned on a valid record

This warning shows up when there is not record in database and you try to read it.
I don't think this warning is linked to what @geoffschoeman did.

geoffschoeman:
Thanks @theshadowx, I'll change the QDataStream version to 5.6.

geoffschoeman:
@nuttyartist, I'll look into adding a progress bar and splitting the update into concurrent tasks to speed up the import process.

I treated this implementation as more of an import/export than a backup/restore, The idea was that I could give you an export of my notes and you could import them without losing your existing notes.  I also wanted to let the user back up their own notes and re-import without duplicating notes that were in both the backup file and their local DB. Because of this the import process runs a few inefficient queries to determine whether to skip, update, or create a new record. 

- If a note exists with the same title and content it is skipped
- If a note exists with the same ID and title the current note is updated
- Otherwise a new note is created

A backup/restore feature would certainly be more efficient but we would lose the ability to import multiple files. Maybe there is a need for both implementations?

theshadowx:
I think the import should check just the id and time of creation/modification: 
```
- if the id doesn't already exist add it to the table
- otherwise the id exists, check the date of creation.
       - if both have the same date of creation check the date of last modification : 
               - if the imported note last date of modification > last date of modification of the existing note : -
                     => do you want to update this note ?
               - otherwise,  
                      => a new version of this note already exists, do you want to replace it ?
        - otherwise 
                 => another note has the same id as this one:
                         do you want to replace it ?
                         do you want to add it with a new id ?
```

nuttyartist:
I didn't have time to review this thoroughly yet. I will try to find time tomorrow or in the coming days.

geoffschoeman:
I split the import task into concurrent threads with a progress bar. This should be much faster. I tested with 1000 notes and it took about 4 seconds.

I did not add the user prompt to choose whether or not to keep the newer or older note when faced with the same note and a different last updated time. I thought this may be cumbersome to prompt every time this occurs, because it could be many. I would like to look into another solution like showing blanket options to apply to all like occurrences like "Keep latest", "Keep original", or "Keep both". Or, maybe showing all conflicts in a single form and allowing the user to select notes to keep. I'm confident that with a little more thought I can come up with a better user experience.

For now I played it safe and both conflicted notes are kept. Here are the current rules:

- if a note with the same ID doesn't exist, create a new one
- if a note with the same ID exists and the creation time is different, create a new one
- if a note with the same ID, and creation time, and last updated time exists, skip import
- if a note with the same ID , and creation time exists, but last updated time is different, create a new one just to be safe

nuttyartist:
Hey @geoffschoeman,

I gave it a bit more though and while it's an interesting idea to do it that way - to do export/import rather than backup/restore, it seems like a very cumbersome and confusing approach - I'm talking from the standpoint of the user. I'm still not sure what's the best option yet. We need a simple and consistent way for letting people manage their notes files.

What's cool about the export/import approach is that it lets you **add** notes from a specific notes file, which is good cause it saves you adding notes one by one (It's hard for me to think about situations people will usually use this, do you have examples on your mind?).

On the other hand, backup/restore is simple, clear, straightforward and mostly hassle-free. You just restore all the notes from a certain file. It can be said that you can achieve the same with exporting/importing by deleting the database file and then importing, but we're talking here about simple users.

In the meantime, I'm going to try to discuss these options with users and see what they would want, in order to have a better picture.

Still, regarding your code, now the app is just crashing after a few seconds, I get "The program has unexpectedly finished.". 

About the rules, why don't look at the content of the notes? This is really what matters, isn't it? What about these rules:

- if a note with the same ID doesn't exist, create a new one
- if a note with the same ID exists and content is different, create a new one
- if a note with the same ID and content, skip

geoffschoeman:
@nuttyartist, I forgot to push a local merge, the app should no longer crash.

As far as the note comparison rules go, the main reason that I keyed off both ID and last updated time stamp was to leave room for future development possibly presenting the user with a list of conflicts and and offering the option to keep the older, newer, or both notes.

Maybe this is overkill and backup/restore is the direction to go.

theshadowx:
@nuttyartist the content of the notes is not a stable indicator as the content is not unique, where id and date of creation/modification gives the indication of the situation of the note and two notes, in normal situation, can't have the same date of creation or the same date of modification, Like in file browser.

I'm still convinced with the proposition I made about importing notes logic, as It's based on what happen in OSes for files. 

@geoffschoeman When you copy a bunch of files from folder to another, it will ask each time what to do when a file already exists with the same name, and also it proposes a checkbox to check if you want to repeat the same operation if it finds other duplicated files.

src/notedata.h
```diff
@@ -4,6 +4,15 @@
...
 
+struct NoteExport
+{
+    QString id;
+    QString fullTitle;
+    QDateTime lastModificationDateTime;
+    QDateTime creationDateTime;
+    QString content;
+};
+
```
theshadowx: why adding a stuct, when NoteData gives the same properties ?

src/mainwindow.cpp
```diff
@@ -1317,6 +1337,113 @@ void MainWindow::checkForUpdates (const bool clicked) {
...
+    QList<QFuture<void>> futures;
+    for (int i = 0; i < noteList.size(); ++i) {
+        futures <<  QtConcurrent::run(this, &MainWindow::importNote, noteList[i]);
+    }
+    for (int i = 0; i < futures.size(); ++i) {
+        while (!futures[i].isFinished()) {
+            // block until all threads complete
+        }
+    }
+}
```
theshadowx: you can use void QtConcurrent::blockingMap(Sequence &sequence, MapFunction function)

theshadowx: you can use [void QtConcurrent::blockingMap(Sequence &sequence, MapFunction function)](http://doc.qt.io/qt-5/qtconcurrent.html#blockingMap) instead of the two loops

src/dbmanager.cpp
```diff
@@ -242,3 +334,69 @@ int DBManager::getLastRowID()
...
+    QString fullTitle = query.value(5).toString();
+    note.id = QStringLiteral("noteID_%1").arg(id);
+    note.fullTitle = fullTitle;
+    note.creationDateTime = dateTimeCreation;
+    note.lastModificationDateTime = dateTimeModification;
+    note.content = content;
+    return note;
+}
+
+void DBManager::importNote(NoteExport noteExport) {
```
theshadowx: void DBManager::importNote(const NoteExport& noteExport)

src/notedata.h
geoffschoeman: @theshadowx, I used the struct to ensure that we serialize only the data necessary to reconstruct the note. NoteData is a QObject which has additional properties not related to the note itself.

src/dbmanager.cpp
```diff
@@ -242,3 +334,69 @@ int DBManager::getLastRowID()
...
+    NoteExport* note = getNoteExport(noteExport.id);
+    bool isNew = false;
+    if (note == NULL) {
```
theshadowx: better to use Q_NULLPTR instead NULL

src/dbmanager.cpp
```diff
@@ -57,6 +57,61 @@ bool DBManager::isNoteExist(NoteData* note)
...
+        QString fullTitle = query.value(5).toString();
+
+        note->id = QStringLiteral("noteID_%1").arg(id);
+        note->creationDateTime = dateTimeCreation;
+        note->lastModificationDateTime = dateTimeModification;
+        note->content = content;
+        note->fullTitle = fullTitle;
+        return note;
+    }
+    return NULL;
```
theshadowx: Q_NULLPTR

src/dbmanager.cpp
```diff
@@ -57,6 +57,61 @@ bool DBManager::isNoteExist(NoteData* note)
...
+        QString fullTitle = query.value(5).toString();
+
+        note->setId(QStringLiteral("noteID_%1").arg(id));
+        note->setCreationDateTime(dateTimeCreation);
+        note->setLastModificationDateTime(dateTimeModification);
+        note->setContent(content);
+        note->setFullTitle(fullTitle);
+        return note;
+    }
+    return NULL;
```
theshadowx: Q_NULLPTR

src/mainwindow.cpp
geoffschoeman: Thanks @theshadowx, I'll take a look.

nuttyartist:
@theshadowx Oh yea, you're right. So the rules of @geoffschoeman seem good. The only thing I don't understand is how this 
> - if a note with the same ID exists and the creation time is different, create a new one

can happen?

And again, the question of whether to go with backup/restore instead of export/import.

theshadowx:
@nuttyartist well I prefer that this feature gives the user the choice between adding or replacing.

src/dbmanager.cpp
```diff
@@ -242,3 +334,69 @@ int DBManager::getLastRowID()
...
+    QString content = query.value(4).toString();
+    QString fullTitle = query.value(5).toString();
+    note.id = QStringLiteral("noteID_%1").arg(id);
+    note.fullTitle = fullTitle;
+    note.creationDateTime = dateTimeCreation;
+    note.lastModificationDateTime = dateTimeModification;
+    note.content = content;
+    return note;
+}
+
```
theshadowx: I think it's better to use pointer to prevent copying.

src/mainwindow.cpp
theshadowx: there might be confusion between them in code.

geoffschoeman:
@nuttyartist, The ID by itself isn't really a good identifier in this case since it is auto-incremented. The first 3 notes you create will have the ID's  1, 2, and 3 respectively as will mine, but the creation date will most likely be different. The combination of both is a pretty good indication that it is the same note.

theshadowx:
@geoffschoeman indeed, more it's very unlikely (if impossible) to have two notes having the same creation time and not be pointing to the same note as, the creation time is in milliseconds.

src/notedata.h
theshadowx: There might be confusion between them in code. 

NoteData is in QObject to use Qt deletion mechanism. I think there is memory leaks in your code, there is no delete after creating noteExport dynamically.

theshadowx:
btw @geoffschoeman you are welcome the chat room https://gitter.im/nuttyartist/notes

nuttyartist:
@geoffschoeman @theshadowx Oh ok got it.

@geoffschoeman I checked again and it keeps on crashing...

src/dbmanager.cpp
```diff
@@ -57,6 +57,33 @@ bool DBManager::isNoteExist(NoteData* note)
...
 
+NoteData* DBManager::getNote(QString id) {
+    QSqlQuery query;
+
+    int parsedId = id.split('_')[1].toInt();
+    QString queryStr = QStringLiteral("SELECT * FROM active_notes WHERE id = %1 LIMIT 1").arg(parsedId);
+    query.exec(queryStr);
+
+    if (query.first()) {
+        NoteData* note = new NoteData();
```
theshadowx: this function might be used elsewhere, so I think it's better to set the parent "this", so if it's not deleted explicitly, we are sure that there would be no leak ;)

src/notedata.cpp
```diff
@@ -98,3 +99,24 @@ void NoteData::setCreationDateTime(const QDateTime&creationDateTime)
 {
     m_creationDateTime = creationDateTime;
 }
+
+QDataStream &operator<<(QDataStream &stream, const NoteData* noteData) {
+    return stream << noteData->id() << noteData->fullTitle() << noteData->creationDateTime() << noteData->lastModificationdateTime() << noteData->content();
+}
+
+QDataStream &operator>>(QDataStream &stream, NoteData* &noteData){
+    noteData = new NoteData();
```
theshadowx: here too

src/mainwindow.cpp
```diff
@@ -1317,6 +1337,102 @@ void MainWindow::checkForUpdates (const bool clicked) {
...
+    } else {
+        QFile file(fileName);
+        if (!file.open(QIODevice::WriteOnly)) {
+            QMessageBox::information(this, tr("Unable to open file"), file.errorString());
+            return;
+        }
+        QDataStream out(&file);
+        out.setVersion(QDataStream::Qt_5_6);
+        out << m_dbManager->getAllNotes();
+    }
```
theshadowx: you forgot to close the file

src/dbmanager.cpp
geoffschoeman: I was receiving an error for each note saying that the parent could not be set from a child thread. Maybe we could make this private to enforce internal use only?

theshadowx: because you're using it in importNote() and importNote() is used in importNotes() which uses QtConcurrent that puts the the imported note in another thread. That's why you get this error.

To use object in another thread it shouldn't have a parent. Well, we'll see what we can do :)

nuttyartist:
Hey guys, sorry for the late reply, had quite a busy weekend.

@geoffschoeman I tested your repo again and now I'm getting this error:
```
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
The program has unexpectedly finished.
```

theshadowx:
@nuttyartist can you show the stack --> by clicking on the run debug and repeating what you've done before.

@geoffschoeman could you include notes included in trash when exporting notes ?

nuttyartist:
@theshadowx OK. It's so weird, sometimes it works well and there's no error, and sometimes there's an error. Of course, while I'm doing the same thing with the same database. I had 2 errors but forgot to copy or screenshot them, and now it's working...

I'll try to update here on the error.

@theshadowx Do you mean they should be added to the Trash, right? Not to the regular notes list.

nuttyartist:
Ok, I now know how to replicate the error. The error stops to occur because I'm using the already-generated notes.nbk file. In order to get the error, I need to export a new one and then remove notes.db and start Notes again and to do Import - which will cause the error. Basically, when I try to import the .nbk file the first time there'es an error and when I try again after killing the process (and after deleting notes.db) it works.

Here are two different screenshots:

selection_009
selection_010

theshadowx:
I think I had something like this before.
The reason why you receive this error , is that in sqlite you can't have many queries at the same time.

http://stackoverflow.com/questions/3736984/multiple-sql-statements-in-qsqlquery-using-the-sqlite3-driver

nuttyartist:
@theshadowx mmm... ok, so if it's risky to do it multithreaded (we're risking here our users' content), can we do it linear again but in a way it will not use the cpu on full blast or something? Or maybe there is a safer way to do multi-threading?

Yea I think it's a good idea to back up Trashed notes too.

theshadowx:
it should be sequential, @geoffschoeman I know it going to take a long time for the import/export operation but it's better than the app to be crashed

nuttyartist:
I thought about achieving eating the whole cake by doing something like this:
Export/More... -> Export, Import (Adds notes from a file), Restore (Replaces all notes within a file)
What do you say?

nuttyartist:
How's it going with this @geoffschoeman?

geoffschoeman:
@nuttyartist, @theshadowx , I just checked in the separate import/restore functionality. I held off on adding the deleted notes to the expor. The deleted_notes table has a non-nullable ID field which is typically set to the active note ID of the note that is being deleted. It is difficult to guarantee that through the various create, add, delete, import, export, and restore processes that the ID will remain unique. I think that the DB constraint can safely be removed as mirroring the ID of a deleted record isn't typically helpful unless there is a solid reason that the ID cannot be reused again.

nuttyartist:
@geoffschoeman And in simple words?

geoffschoeman:
Removing the null Id constraint on the deleted messages table will allow us to restore deleted​ notes without having to worry about conflicting IDs.

theshadowx:
@geoffschoeman the imported notes ids must be re-adjusted to be added the tables.

nuttyartist:
What about importing them and giving them unique ID's?

nuttyartist:
@geoffschoeman I forgot to check your latest commit. Checked it now.

Wow, now it's super fast. What changed so much?
Looks pretty good.

Few things to note:

- It's important to add the information in the brackets: Import (Adds notes from a file), Restore (Replaces all notes within a file), otherwise the user will get confused about the function of all the options.
- I exported my notes and then imported them back, without deleting my database, and it just duplicated all of them, is that part of the process you defined?

nuttyartist:
@geoffschoeman We're getting very close to finishing all the things for version 1.0.0, can you push a bit further and check the things I discussed above? I think it's important that this feature will be in the next version.

geoffschoeman:
@nuttyartist Yes I will have some time this weekend to add the menu item descriptions.

nuttyartist:
Alrighty. I will review everything tomorrow just to check that the import does import all the notes successfully. Then I will merge.

Also, can you elaborate about that?

> - I exported my notes and then imported them back, without deleting my database, and it just duplicated all of them, is that part of the process you defined?

Thanks.

geoffschoeman:
@nuttyartist yes I intended both the import and restore to add all notes, only difference being the restore warns the user and removes the current notes before the import.

nuttyartist:
Ok, I checked everything.

Last tiny things to do:

- Can you show the warning message right after clicking on restoring, and not after choosing the file? This should put the user in more control by letting him know immediately what is going to happen. 
- Change the labels in the menus to "Import (add notes from a file)" and "Restore (replace all notes with notes from a file)", I think it should be better.

nuttyartist:
OK, everything seems to work just right.
Good work @geoffschoeman, this feature would be useful to a lot of users (just got an email from someone, out of many, on where the notes are located so he could back them up).

Merging...

PR #141, opened 2017-06-10
Fix #137

### LINKED ISSUE #137: App can't connect to db
During the first launch of the app, it tries to connect to database while the file is not being created.

theshadowx:
force to create settings folder before trying to create database file.
initialize pointer members.


nuttyartist:
Merging...

PR #147, opened 2017-07-02
add install fakeroot

theshadowx:


nuttyartist:
so this solves the problem with travis...

Issue #65, opened 2016-05-25
Make the option for the window to be Always on top


nuttyartist:
Should we do it?
If so, how we can do it without cluttering the interface? We are not going to have settings panel so we got to think on another way. 

In future releases, there will be a three-dot-button to activate some features, but I don't want to clutter it so we'll see how everything look, when I'll upload my design proposition in [wiki/Design](https://github.com/nuttyartist/notes/wiki/Design) soon.

theshadowx:
I'm thinking about something that looks like this but not with the same button design

![pin on top](http://www.techsupportalert.com/files/images/FileBox-On-Top.jpg)

nuttyartist:
I think this kind of integrated menu bar not suite Notes because it doesn't go well with the design, it is a waste of vertical space, and it can make the app clutter with many options that will ruin its simplicity.

Yet, if our app wasn't cross-platform and would only run on os's with global menu bars like Ubuntu Unity and OS X etc... then taking that approach will be good because it's part of the os, so it doesn't clutter the app and not wasting free vertical space.

But we are aiming for supporting many os's so we gonna need to look for other option.

theshadowx:
it's not the menu which I'm showing here but the pin near '-' .

which its form is modified depending on whether app is pinned or not.

nuttyartist:
Ohh. Buttons like that near the + and trash button are very valuable, the user use them frequently, and the option for Always on top is really not high enough to put it in a button, it's a feature that its place is in some sort of menu that people discover. So I suggest we put it the 3 dot button in the future and also create a keyboard shortcut for this and update the ones on the website.

Also, putting notes Always on top, a.k.a pinning them on your screen, is mostly relevant for kind of sticky notes, so a good place to put a button like you showed might be when you double click a note (or right click and choose...) and open it in a separate window, this might be a good place to show the Pin Button.

nuttyartist:
We gonna drop it because it doesn't work well.

nuttyartist:
We should really have that option. We can do it in our (View -> Always on top) menu.

nuttyartist:
Solved #181.

PR #181, opened 2017-10-13
Implemented staying always on top

torbei:
See issue #65 

Option is located in the dots menu > View > Always stay on top.

Worked with Ubuntu 17.04 Zesty. Can't test under Windows or macOS

At the moment it does not have a keyboard shortcut but I think it should have one. Which one would you prefer?

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -986,6 +986,13 @@ void MainWindow::onDotsButtonClicked()
         exportNotesFileAction->setDisabled(true);
     }
 
+    // Stay on top action
+    QAction* stayOnTopAction = viewMenu->addAction(tr("Always stay on top"));
+    stayOnTopAction->setToolTip(tr("Always keep the notes application on top of all windows"));
+    stayOnTopAction->setCheckable(true);
+    stayOnTopAction->setChecked(m_alwaysStayOnTop);
+    connect(stayOnTopAction, SIGNAL(triggered(bool)), this, SLOT(stayOnTop(bool)));
+
     mainMenu.exec(m_dotsButton->mapToGlobal(QPoint(0, m_dotsButton->height())));
 }
 
@@ -2556,4 +2563,25 @@ bool MainWindow::eventFilter (QObject *object, QEvent *event)
     }
 
     return QObject::eventFilter(object, event);
+}
+
+void MainWindow::stayOnTop(bool checked) {
+    Qt::WindowFlags flags;
+#ifdef Q_OS_LINUX
+    flags = Qt::Window | Qt::FramelessWindowHint;
+#elif _WIN32
+    flags = Qt::CustomizeWindowHint;
+#elif __APPLE__
+    flags = Qt::Window | Qt::FramelessWindowHint;
+#else
+#error "We don't support that version yet..."
+#endif
+    if (checked) {
+        flags |= Qt::WindowStaysOnTopHint;
+        m_alwaysStayOnTop = true;
+    } else {
+        m_alwaysStayOnTop = false;
+    }
+    this->setWindowFlags(flags);
+    setMainWindowVisibility(true);
 }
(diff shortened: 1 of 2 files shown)


theshadowx:
The big issue we had for this feature, is to work on Windows. 
I think we had been able to make it work on Linux before.

Could you make sure that it work for Windows.

torbei:
Ok so I setup a virtual machine with Windows 10 from here: https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/ and tried to compile the PR with Qt Creator 4.3.1

Staying on top works without any issues on Windows 10 Enterprise here as far as I can tell

torbei:
Before you merge it, what keyboard shortcut should it get?

nuttyartist:
Hi there @torbei!

Thanks for the contribution. I tested on a Mac and it seems to work well. 2 points:
- Because the window is always on top, the traffic-light-buttons icons' should keep their default-idle icon, and not change to "unfocusedButton.png". Also, the shadow state should not change.
-  For some reason, I can only see the '✓' icon only when I hover over the menu, but it should be shown also when I don't hover and highlight the menu. 

I would really appreciate this feature. These 2 points need to be fixed before merging. Also, I remember we had problems on some Linux/Windows machines so I would continue testing and update here.

nuttyartist:
By the way, regarding the keyboard shortcut, I've been thinking about 'Ctrl + k'. What do you think?

torbei:
>  regarding the keyboard shortcut, I've been thinking about 'Ctrl + k'. What do you think?

I don't see any issue there.

> These 2 points need to be fixed before merging.

Okay I will try to fix this in the near future

nuttyartist:
Ok 👍

torbei:
> - For some reason, I can only see the '✓' icon only when I hover over the menu, but it should be shown also when I don't hover and highlight the menu.

So I can't recreate that issue. Not on Ubuntu and neither in the Windows VM. Would it be possible for you to share a gifv or something from this issue?

> - Because the window is always on top, the traffic-light-buttons icons' should keep their default-idle icon, and not change to "unfocusedButton.png". Also, the shadow state should not change.

I don't know if this is a good idea. Because that behavior is standard when the application looses it's focus. Even if I use my window manager (KDE) and keep the window on top of the others it will loose the focus (and therefore change to "unfocusedButton.png") if I'm actively working in another application. The same behavior goes for other applications (e.g.  keeping the terminal or file manager window on top of the other windows). What is your opinion about this considering the previous point?

nuttyartist:
Here, the GIF is a bit slow.
always_on_top

Mmm yeah, you've got a point. Ok, so just the problem in the GIF is left. Is that occurs to you on other platforms too?

torbei:
No this does not occur on Linux or Windows for me. See the gif below for Linux. Same goes for Windows.

notes

Maybe a bug with Qt on Mac? Although I couldn't find something with a quick search.

theshadowx:
I think it's theme issue, are you using black theme ?

nuttyartist:
> I think it's theme issue, are you using black theme ?

No, I'm using the default one.

> Maybe a bug with Qt on Mac? Although I couldn't find something with a quick search.

Sounds like it.

torbei:
@nuttyartist Ok I think I found something about this issue:

torbei:
@nuttyartist OK I think I found something about this issue:

https://stackoverflow.com/questions/31259496/checked-symbol-in-checkbox-is-not-visible-unless-the-element-is-hover

Could you try if you could fix it? Because I don't have a Mac System and can't reproduce it.

nuttyartist:
@torbei What code changes should I make?

torbei:
Looks like it has something to do with the background specified here:

https://github.com/nuttyartist/notes/blob/dev/src/mainwindow.cpp#L926

I would try to remove it and look what happens.
Maybe also the background color in [L930](https://github.com/nuttyartist/notes/blob/dev/src/mainwindow.cpp#L930)

nuttyartist:
> Looks like it has something to do with the background specified here:
 https://github.com/nuttyartist/notes/blob/dev/src/mainwindow.cpp#L926

That's right. I've checked, and when removing that line it works as expected. But the thing is, we do need to have a white background, so how can we solve this?

torbei:
Could you try adding this:
`QMenu::indicator:checked
{
    image: url(:/icons/checkbox_checked);
}`

from the [Stackoverflow topic](https://stackoverflow.com/questions/31259496/checked-symbol-in-checkbox-is-not-visible-unless-the-element-is-hover) to the stylesheet after [L931](https://github.com/nuttyartist/notes/blob/dev/src/mainwindow.cpp#L931).

I don't know if this would work, because there is no answer in the stackoverflow topic.

nuttyartist:
No, it doesn't work well.
I think we should merge this and create an issue for this bug to solve in the future.

nuttyartist:
Merging... Good work, @torbei.

PR #188, opened 2018-01-13
fix bugs 168 & 169

m-o-s-t-a-f-a:
In this commit I have defined a new shortcut for toggling note lists and make traffic light visible when the note list is hidden.


nuttyartist:
Seems to work very well! Why have you decided to dynamically add the traffic button lights? And not just put them inside the form?

m-o-s-t-a-f-a:
I wanted to give an absolute position to the traffic. and if user hides or shows NoteList, it won't change. The parent for traffic lights is main window, neither MainWindow nor NoteList.
If we put it on the form we need to assign a parent widget for them. hiding parent widget (in this case notelist) will hide traffic light as well.

Another approach is having two different traffic light. one for NoteList and one for Note. I think it's meesy and hard to control the events.

m-o-s-t-a-f-a:
I also made another change. I saved the geometry when the window is maximized. when user re-open Notes it will restore its geometry again (maximized). For full-screen mode it has no effect. I have tested it on Ubuntu.

nuttyartist:
Very nice that you added the keyboard shortcut hint!

> I wanted to give an absolute position to the traffic light and if user hides or shows NoteList, it won't change. The parent for traffic lights is main window, neither MainWindow nor NoteList.
> If we put it on the form we need to assign a parent widget for them. hiding parent widget (in this case NoteList) will hide traffic light as well.
> 
> Another approach is having two different traffic lights. one for NoteList and one for Note. I think it's messy and hard to control the events.

Ok

> I also made another change. I saved the geometry when the window is maximized. when user re-open Notes it will restore its geometry again (maximized). For full-screen mode it has no effect. I have tested it on Ubuntu.

Works well.

Well, everything seems to work well. Merging...

Thanks, Mostafa, I would love to see even more from you.

Issue #193, opened 2018-01-29
Notes loose its window state after after hide/show main window


b00f:
If you change window state to full-screen (or maximized) and then hide/show main window it cause to lose its state.

PR #194, opened 2018-01-29
fix issue #193 - Notes loose its window after after hide/show main window

b00f:
This patch is a fix for issue #193


nuttyartist:
Seems to work well on Windows.

nuttyartist:
And Mac too. Merging...

Issue #202, opened 2018-03-06
Bug related highlighting search text


b00f:
![screenshot from 2018-03-06 16-04-26](https://user-images.githubusercontent.com/8073510/37020796-12dc4d7e-2158-11e8-867d-318c23684951.png)
How to produce it:
1- search for a text
2- got to text and put cursor on it
3- press Alt+tab to switch the window
4- back to notes 
5- all text are highlighted!

nuttyartist:
I don't encounter this bug. On what OS are you getting this?

b00f:
I fixed it. I could generate it in my PC (ubuntu 16.04)

PR #203, opened 2018-03-07
Fixing bug#202

b00f:


theshadowx:
please clean, why for just fixing bug #202 you need all these commits which does't have relation with the bug.

b00f:
When user switch between windows (Alt+Tab) notes won't lose the cursor position anymore.

nuttyartist:
That's weird, I haven't experienced that (on a Mac). Can you send a short video of the Notes losing the cursor position after Alt+Tab?

Also, Indded like @theshadowx  said, it seems like there are a lot of commits not related to this bug? Can you clean it up or send a new PR please?

b00f:
I think I know why you can't generate this bug. In version 1.0.0, when user search for a text and after that click on the note, the search will clear and he lose the search text.
in new version by clicking on notes you won't lose your search (highlighted text still remains yellow)
However if user switch between windows the search again get cleared and cursor backed to the beginning of the note. 
Look here:


```
    case QEvent::FocusIn:{	
-        if(object == m_textEdit){	
-	
-            m_noteView->setCurrentRowActive(true);	
-	
-            if(!m_isOperationRunning){	
-                // When clicking in a note's content while searching,	
-                // reload all the notes and go and select that note	
-                if(!m_searchEdit->text().isEmpty()){	
-                    m_selectedNoteBeforeSearchingInSource = QModelIndex();	
-	
-                    if(m_currentSelectedNoteProxy.isValid()){	
-                        QModelIndex indexInSource = m_proxyModel->mapToSource(m_currentSelectedNoteProxy);	
-                        clearSearch();	////// BUG 202
-                        m_currentSelectedNoteProxy = m_proxyModel->mapFromSource(indexInSource);	
-                        selectNote(m_currentSelectedNoteProxy);	 ////// BUG 202
-	
-                    }else{	
-                        clearSearch();	////// BUG 202
-                        createNewNote();	////// BUG 202
-                    }	
-	
-                    m_textEdit->setFocus();	
-	
-                }else if(m_proxyModel->rowCount() == 0){	
-                    createNewNote();	
-                }	
-            }	
-        }
```

theshadowx:
Hi @b00f it's not a bug, this is how the search feature was meant to be. Maybe it creates in your update an issue. This why before modifying please check how things work or just ask, so that we could discuss about whether to modify it or not.

when the user click in document, it means that the user has found what was searching for, this is why the  search is  cleared, maybe we can add a checkbox to toggle the search so that the click on the note wouldn't clear the search ? what do you thing ? @b00f @nuttyartist

nuttyartist:
@b00f Indeed like @theshadowx said this is how the search workflow was meant to be. Can you find where in your recent changes it clashes with the current code? Also, please clean the big number of commits and leave only the ones related to this #202 bug fix.

@theshadowx Yes, I always felt something wrong about this decision. I think I had good motivation behind this decision in the past, but I now know it's not right. Checkbox where? I think what we should do is get rid of this behavior and instead when user click/changes/scroll a note while in search mode, nothing will happen. Only when he clear the search field himself - by clicking the 'x' button or changing the search keywords - the notes list view will change. Of course, the time of the note should change accordingly and updated in the notes list. I will create a new issue for this.

theshadowx:
Hi @b00f , no worries, there is solution to that : 
. create branch from nuttyartist/dev
. Add your fix (you should redo what you've done to fix this issue again)
. then PR (the current PR will be closed)

Sorry for taking so much time to answer.

nuttyartist:
What @theshadowx said ^^

b00f:
Moved to this branch:
https://github.com/nuttyartist/notes/tree/highlight-search

PR #209, opened 2018-04-21
bringing back removed features:

theshadowx:
. showing blue rectangle when the searchEdit has focus.
. create a new note when there is no note in the notelist and text edit has focus on


nuttyartist:
Those features were removed because of the _highlight-search_ repo?

theshadowx:
instead of removing just the part that generating the #202 bug, Other things that were not related to it were removed.

nuttyartist:
Aha ok, good notice.

PR #213, opened 2018-04-26
Only allow dragging from top part of window

### LINKED ISSUE #173: Window should be draggable only from the top part of the window.
My opinion is that a Window should be draggable only from the top part of the Window from y:0 to lineEdit:y. This is the expected behavior on most operating systems and not from any point in the app.

Plonq:
`QMainWindow` allows dragging any empty part of the window to move
it, as well as double clicking to maximize/minimize.
This patch adds to the existing event filter for `frame`, listening
for `MouseButtonPress` and `MouseButtonDblClick` events, only
allowing them through if the click's Y coordinate is anywhere above
the search text edit field (`ui->searchEdit`).

Fixes #173

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -2596,8 +2596,7 @@ bool MainWindow::eventFilter (QObject *object, QEvent *event)
         // from the top part of the window
         if(object == ui->frame){
             QMouseEvent* mouseEvent = static_cast<QMouseEvent*>(event);
-            const int lowerBound = ui->verticalSpacer_upSearchEdit->geometry().bottom();
-            if(mouseEvent->y() > lowerBound){
+            if(mouseEvent->y() >= ui->searchEdit->y()){
                 return true;
             }
         }
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -2596,7 +2596,8 @@ bool MainWindow::eventFilter (QObject *object, QEvent *event)
         // from the top part of the window
         if(object == ui->frame){
             QMouseEvent* mouseEvent = static_cast<QMouseEvent*>(event);
-            if(mouseEvent->y() >= ui->searchEdit->y()){
+            const int lowerBound = ui->verticalSpacer_upSearchEdit->geometry().bottom();
+            if(mouseEvent->y() > lowerBound){
                 return true;
             }
         }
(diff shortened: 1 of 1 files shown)


theshadowx:
The way you did it, may create a bug --> What if the searchEdit is moved somewhere else ? this requirement won't be satisfied.

nuttyartist:
Hi @Plonq!, thanks for contributing to Notes.

@theshadowx Is the only problem is that he calculates the area of Y's that allow drag and drop to happen by the searchEdit's Y? If so, it doesn't seem like a problem to me. If something changes, we can easily tweak this to a different element.

What other approach is there? The only I can think of is to take the Y's of the traffic-light buttons and add a small constant to their Y and calculate it by the result.

Plonq:
No problem @nuttyartist :)

@theshadowx that is a good point, which I had considered. But as @nuttyartist points out, the only other alternative (without reworking the layout) is to use some sort of fixed value. I judged that this method was better as it would adapt to small changes in the layout automatically.

A more elegant solution might be to add a widget that is parent to everything but child to `frame`, which covers everything except the 'titlebar' section we want draggable.

theshadowx:
@Plonq  @nuttyartist as an alternative to add a frame and use it as title bar, which will contain traffic light, and include it into the layout with fixed height, so there would be no dependencies to other component.

Plonq:
@nuttyartist let me know if you'd prefer the more elegant solution, and I'll give it a shot.

nuttyartist:
@Plonq @theshadowx I don't like fixed variables. It should be done as dynamically as possible. 

@Plonq If your elegant solution doesn't bound you to use constant variables then you can give it a go.

Plonq:
@nuttyartist I tried out the different layout, but since the backgrounds are added to the sides of the splitter, the splitter needed to fill the entire window, so it didn't work as I had thought.
I have instead changed it so it's based on the left vertical spacer, which is better than using the position of the search edit.

nuttyartist:
Why is it better than then search box? The search box is a more permanent object (talking about being affected in future versions) than the left vertical spacer.

theshadowx:
the vertical spacer can be replaced.

nuttyartist:
@theshadowx Yeah. @Plonq Replace it back to the search box and I'll merge this. I think it's flexible enough for now.

Plonq:
Fair enough. My reasoning was:

1. The spacer is named (after the search edit), so I figured it wouldn't just be replaced.
2. The spacer is the 'source' of the fixed size (25px).

But it's up to you :)

nuttyartist:
@Plonq But there is a higher chance the spacer will change it's position or be eliminated in future releases than the search box.

That's good now. Thanks, @Plonq!

Merging...

PR #223, opened 2018-10-15
Remove threading on all db operations

vdombrovski:
This is proposed as a workaround for a silent bug present when using QT 5.11:

When called via a `QtConcurrent` thread, the query would not complete yielding:

`QSqlDatabasePrivate::database:requested database does not belong to the calling thread`

Which would result in a note potentially not being persisted, and thus effectively removed once the application closes.

Now I understand that performing I/O operations synchronously in the main thread could lead to freezes, so this is definitely not the best workaround. It would be nice if someone else could have another look at this issue.

### DIFF
--- a/src/dbmanager.cpp
+++ b/src/dbmanager.cpp
@@ -3,7 +3,6 @@
 #include <QTimeZone>
 #include <QDateTime>
 #include <QDebug>
-#include <QtConcurrent>
 
 DBManager::DBManager(const QString& path, bool doCreate, QObject *parent) : QObject(parent)
 {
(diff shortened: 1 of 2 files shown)


nuttyartist:
Hey @vdombrovski, thanks for pointing this out!

Is this bug occur only on Qt 5.11?
If so, have you identified why?

@theshadowx Can you check the PR, please?

vdombrovski:
Here, I've just put it to the test once more

Qt 5.10
```sh
$ pacman -Q qt5-base
qt5-base 5.10.0-1

$ ./notes 
Database: connection ok
# when saving a note
```

Qt 5.11
```sh
$ pacman -Q qt5-base
qt5-base 5.11.2-1

$ ./notes 
Database: connection ok
# when saving a note
QSqlDatabasePrivate::database: requested database does not belong to the calling thread.
QSqlQuery::exec: database not open
```

Also, Qt release notes confirm this behavior:

http://code.qt.io/cgit/qt/qtbase.git/tree/dist/changes-5.11.0/?h=v5.11.0#n231

nuttyartist:
@vdombrovski Thank you for investigating this issue. I'll wait for our lead developer @theshadowx to approve/suggest changes to this PR.

theshadowx:
I this the issue mentioned by this PR has been fixed in the this one https://github.com/nuttyartist/notes/pull/230

nuttyartist:
Okay. @vdombrovski report to us if you encounter this issue again, please. For now, we'll close this.

PR #230, opened 2019-12-01
Fixing database

theshadowx:
This is fixing the database according to the latest updates on Qt on sql.
The database manager is now working in its own thread dependent from the main one.
The communication between main thread and the one related to DBManager is done through signals and slots.

### DIFF
--- a/src/dbmanager.cpp
+++ b/src/dbmanager.cpp
@@ -5,59 +5,100 @@
 #include <QDebug>
 #include <QtConcurrent>
 
-DBManager::DBManager(const QString& path, bool doCreate, QObject *parent) : QObject(parent)
+/*!
+ * \brief DBManager::DBManager
+ * \param parent
+ */
+DBManager::DBManager(QObject *parent)
+    : QObject(parent)
 {
     qRegisterMetaType<QList<NoteData*> >("QList<NoteData*>");
+}
 
+/*!
+ * \brief DBManager::open
+ * \param path
+ * \param doCreate
+ */
+void DBManager::open(const QString &path, bool doCreate)
+{
+    QSqlDatabase m_db;
     m_db = QSqlDatabase::addDatabase("QSQLITE");
-    m_db.setDatabaseName(path);
 
-    if (!m_db.open()){
+    m_db.setDatabaseName(path);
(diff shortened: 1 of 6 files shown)


nuttyartist:
Hey!

I'm having problems cloning your repo. I'm using this command:

`git clone --single-branch --branch fixing_database https://github.com/theshadowx/notes.git`

But when I try to compile using Qt I get:

```
/Users/ruby/Downloads/notes/src/mainwindow.h:26: error: 'QAutostart' file not found
#include <QAutostart>
         ^~~~~~~~~~~~
```

theshadowx:
you should add --recursive 

Btw I can't access to Slack my account to note group was desactivated

nuttyartist:
Sorry, for that! Slack should work now.

Everything seems to work well.

2 notes:
1. At first, I got compiling errors because of `qautostart.cpp`. It's needed to add `#include <QProcess>` and replace variable name from`dir` to `appDir`.
2. Just tiny thing I found is when I click the 3 dots button I get in the debug this message: `syntax error: Expected end of line but found class name. (-2741)`. Didn't see any difference visually tho.

Should you let me know what we gonna do about (1) update it upstream or in our repo? And I'll merge this.

theshadowx:
weird that you get this issues, I didn't have them.
Which Qt have you used and what OS ?

nuttyartist:
Just updated to the latest Qt version 5.13.1. Using Mac OS Mojave. The second problem is minor/not important. But the first one causes the program not to compile (on my machine at least). Which makes sense cause code I mentioned from qautostart is wrong.

theshadowx:
which .pro file did you load ?

nuttyartist:
Both in root and src.

theshadowx:
The errors you mentioned are not related to the pull request.

autostart is not supported in MacOS, I will fix that issue after this is been accepted.
Also as we imported that repo we can't modify it.

the second issue is related to osascript.

I'll disable autostart for mac.

nuttyartist:
Maybe we can import it into our repo and then change it?

theshadowx:
I will take a look at that later

nuttyartist:
Ok. 

Good work (; Merging...

PR #231, opened 2019-12-07
Notes gh workflow

theshadowx:
Adding Ubuntu and MacOS workflows


nuttyartist:
Can you explain workflow for what? Travis? App-veyor? Something else?

theshadowx:
This one looks like appveyor/travis but it's from GitHub itself, so it might be better as it's included in GitHub platform

nuttyartist:
Gotcha! Merging...

PR #232, opened 2019-12-13
disabling autostart for macos

theshadowx:
workaround for autostart in macos


nuttyartist:
Sorry for the delay. Is disabling Qautostart will disable the automatic updates?

theshadowx:
autostart and autoupdate are two different independent features

nuttyartist:
Okay, merging... (;

Issue #81, opened 2016-07-25
Highlight searched words inside the note editor when searching. [15$]


nuttyartist:
We should highlight the searched words inside the editor when doing a search.
We may also want to go right to the first searched word found in the editor on each selected note.

After the PR #198, two things remaining:

- [x] Decide on a yellow shade. EDIT: the current one is good.
- [x] Go right to the first searched word found in the editor on each selected note. The found string should be displayed in the middle of the editor.

nuttyartist:
After the PR #198, two things remaining:

- [ ] Decide on a yellow shade.
- [ ] Go right to the first searched word found in the editor on each selected note.

nuttyartist:
@b00f Is there a way you can make the editor go to the first highlighted word for each selected note when searching if it's not already visible?

b00f:
I couldn't find a way to move scroll bar. I don't know yet the answer.

nuttyartist:
I know you can change the scrollbar value using QScrollBar setValue. Now we need to find a way to get the Y position of a line (that the highlighted text is found in). Then we'll need a way to check if this position is visible to the user. If not we should set the scrollbar value to the position, else, there's nothing to do since it's visible.

b00f:
I did one day work on it. I couldn't get the scroll bar position. I am not sure if there is a way or not.
I am afraid to tell but I can't work on it again.
Maybe you can check the google for a good solution.

nuttyartist:
Done.

nuttyartist:
@Waqar144 you can claim your bounty here: https://www.bountysource.com/issues/36332307-highlight-searched-words-inside-the-note-editor-when-searching-15

Waqar144:
> @Waqar144 you can claim your bounty here: https://www.bountysource.com/issues/36332307-highlight-searched-words-inside-the-note-editor-when-searching-15

Thanks :)

Issue #174, opened 2017-09-05
Improvement for searching experience


nuttyartist:
Currently, It's fightening to search something and then press enter to get to the note, because you afraid you will press enter again and mistakingly change the note and it will go up the list. 

A solution would be that when searching a note and pressing enter it will get you to the location of the note but the focus would be on the list. Pressing enter again will focus the note.

nuttyartist:
Not sure if it's a good idea actually.

PR #236, opened 2019-12-19
Implement #81

Waqar144:
#81

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -601,6 +601,7 @@ void MainWindow::setupSearchEdit()
 void MainWindow::setupTextEdit()
 {
     QString ss = QString("QTextEdit {background-image: url(:images/textEdit_background_pattern.png); padding-left: %1px; padding-right: %2px; padding-bottom:2px;} "
+                         "QTextEdit{selection-background-color: rgb(65, 154, 223);}"
                          "QScrollBar::handle:vertical:hover { background: rgb(170, 170, 171); } "
                          "QScrollBar::handle:vertical:pressed { background: rgb(149, 149, 149); } "
                          "QScrollBar::handle:vertical { border-radius: 4px; background: rgb(188, 188, 188); min-height: 20px; }  "
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -1412,7 +1412,11 @@ void MainWindow::selectNoteUp()
             m_currentSelectedNoteProxy = aboveIndex;
             showNoteInEditor(m_currentSelectedNoteProxy);
         }
-        m_noteView->setFocus();
+        if (!m_searchEdit->text().isEmpty()) {
+            m_searchEdit->setFocus();
+        } else {
+            m_noteView->setFocus();
+        }
     }
 }
 
@@ -1436,7 +1440,12 @@ void MainWindow::selectNoteDown()
                 showNoteInEditor(m_currentSelectedNoteProxy);
             }
         }
-        m_noteView->setFocus();
+        //if the searchEdit is not empty, set the focus to it
+        if (!m_searchEdit->text().isEmpty()) {
+            m_searchEdit->setFocus();
+        } else {
+            m_noteView->setFocus();
+        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


nuttyartist:
Regarding #81, Your implementation highlights the next found phrase on every "Return" press. While it's a really good addition, it doesn't yet show the first phrase found if it's not in the view (you need to scroll to find it or press enter). Could you add that it automatically scrolls down to the found phrase? It should work also when a user searches a phrase and then continue looking in other notes (by scrolling with the arrow keys), the found phrases should be shown if not visible.

Regarding #174, you didn't implement what suggested in the issue, rather you eliminated notes to be focused when pressing enter (from my trials).

Waqar144:
> Regarding #81, Your implementation highlights the next found phrase on every "Return" press.

Yes, I thought that was how it should be :thinking:

While it's a really good addition, it doesn't yet show the first phrase found if it's not in the view (you need to scroll to find it or press enter). Could you add that it automatically scrolls down to the found phrase?

Hmm.. I think its possible

It should work also when a user searches a phrase and then continue looking in other notes (by scrolling with the arrow keys), the found phrases should be shown if not visible.

You mean the search should be carried to the next note(on arrow key press)?

> Regarding #174, you didn't implement what suggested in the issue, rather you eliminated notes to be focused when pressing enter (from my trials).

Ah, Sorry. My bad. I guess I misunderstood that one.

Having a shortcut as enter prevents the searchedit to function correctly on 'Enter' key press.

nuttyartist:
Okay, awesome! Now it's scrolling to the first result. Just a little thing before merging:

1. Replace the gray color (representing focused find) with rgb(65, 154, 223). Keep the yellow color for unfocused find.
2. While searching and selecting different notes with the arrow keys, it works very well by highlighting everything, but pressing the enter button doesn't lead to highlighting the next found phrase and going to its place (only when lineEdit is in focus). Can you implement that?

Waqar144:
> Okay, awesome! Now it's scrolling to the first result. Just a little thing before merging:
> 
>     1. Replace the gray color (representing focused find) with rgb(65, 154, 223). Keep the yellow color for unfocused find.
> 
>     2. While searching and selecting different notes with the arrow keys, it works very well by highlighting everything, but pressing the enter button doesn't lead to highlighting the next found phrase and going to its place (only when lineEdit is in focus). Can you implement that?

Done, I think.

nuttyartist:
Great! Works well. Just a tiny bit of change, can you change the focused color to rgb(63, 99, 139)?

Sorry for being indecisive.

Waqar144:
sure no problems!

nuttyartist:
Seems to be straightforward enough to merge. @Waqar144 Thank you!

src/mainwindow.cpp
```diff
@@ -313,8 +313,8 @@ void MainWindow::setupKeyboardShortcuts()
     new QShortcut(QKeySequence(Qt::Key_Up), this, SLOT(selectNoteUp()));
     new QShortcut(QKeySequence(Qt::CTRL + Qt::Key_Enter), this, SLOT(setFocusOnText()));
     new QShortcut(QKeySequence(Qt::CTRL + Qt::Key_Return), this, SLOT(setFocusOnText()));
-    new QShortcut(QKeySequence(Qt::Key_Enter), this, SLOT(setFocusOnText()));
-    new QShortcut(QKeySequence(Qt::Key_Return), this, SLOT(setFocusOnText()));
+    //new QShortcut(QKeySequence(Qt::Key_Enter), this, SLOT(setFocusOnText()));
+    //new QShortcut(QKeySequence(Qt::Key_Return), this, SLOT(setFocusOnText()));
```
theshadowx: The shortcut shouldn't have been commented for this PR

nuttyartist: Oh, true! Missed that one. @theshadowx Can you push a commit for that?

Waqar144: Without it, 'Enter' key press won't work in the line-edit. Everytime you press enter, instead of searching it will go inside the text edit.

nuttyartist:
@Waqar144 It seems like because of this PR, when the user hit Enter it doesn't focus on the note's content. Can you create a new PR which fixes this? I guess it needs to be conditioned on whether a search is active or not - if it's not active: focus on the content. If it's active: go to the next found keyword. If the search is active but the user still wants to focus on a note's content, it should be using CTRL+Enter.

Waqar144:
> @Waqar144 It seems like because of this PR, when the user hit Enter it doesn't focus on the note's content. Can you create a new PR which fixes this? I guess it needs to be conditioned on whether a search is active or not - if it's not active: focus on the content. If it's active: go to the next found keyword. If the search is active but the user still wants to focus on a note's content, it should be using CTRL+Enter.

Done in #240

PR #237, opened 2019-12-19
Markdown highlighting support

Waqar144:
testing branch


nuttyartist:
I really like it! Since it almost doesn't impact our previous version - most notes would look plain text, and can *keep* on being plain text, it's a very easy case to merge.

@theshadowx feedback?

bjorn:
Did you consider adding https://github.com/pbek/qmarkdowntextedit as a git submodule instead of copying it all? It's already set up to be usable as a submodule since it's [used as a submodule by QOwnNotes](https://github.com/pbek/QOwnNotes/tree/develop/src/libraries).

nuttyartist:
@bjorn Someone can create a new PR for this and I'd like to test that. What are the differences to this PR? Is it worth adding the submodule rather than using what Qt as already built-in?

Waqar144:
> Did you consider adding https://github.com/pbek/qmarkdowntextedit as a git submodule instead of copying it all? It's already set up to be usable as a submodule since it's [used as a submodule by QOwnNotes](https://github.com/pbek/QOwnNotes/tree/develop/src/libraries).

Hi,
@bjorn 
Thanks. Yes, I did. I just set this up for quick testing so that its straighforward to test. If @nuttyartist  decides thinks its okay, I will add `qmarkdowntexedit` as a submodule.

@nuttyartist the advantage is that it makes it very simple to update the library. QMarkdownTextedit has changed quite a lot since the time i created this branch. Also, being the dev behind `qmarkdowntextedit`, i would personally prefer it as a git submodule because we are frequently changing/improving/updating things over there

bjorn:
> What are the differences to this PR? Is it worth adding the submodule rather than using what Qt has already built-in?

@nuttyartist My question was not about using something Qt had built-in, but about the way the source code of `qmarkdowntextedit` is included in this repository. One way is to make a copy (like done in this PR) but alternatively its repository can be referenced as a git submodule. Referencing the original repository makes it more visible where the code comes from and easier to update the reference to newer versions later. It does affect your git workflow a little, but it's a good feature to be  familiar with anyway.

I didn't realize @Waqar144 is actually a maintainer of qmarkdowntextedit. :-)

Waqar144:
> I didn't realize @Waqar144 is actually a maintainer of qmarkdowntextedit. :-)

And a lot people don't realize that you are one of the first authors of our dear QtCreator #respect ;)

nuttyartist:
@Waqar144, @bjorn Sounds good to me to have it as a submodule. @Waqar144 You can go about changing it.

It's great to have people that care about awesome open-source software here :)

Waqar144:
Added `qmarkdowntextedit` as submodule

nuttyartist:
Good work, @Waqar144. Merging...

Issue #234, opened 2019-12-18
Note don't get saved


Waqar144:
Notes won't get saved if they have content like this:
```
"hello %1 hello"
```

theshadowx:
Which version of Notes are you using and which Qt version is used

Waqar144:
I built it myself(dev branch) with Qt 5.13.2

theshadowx:
Thanks for reporting this issue

nuttyartist:
That's weird. For me it saves the note always differently, after a re-run it becomes: "hello hello %1 hello hello".

Is it because the special sign fucks up with SQL?

Waqar144:
> That's weird. For me it saves the note always differently, after a re-run it becomes: "hello hello %1 hello hello".
> 
> Is it because the special sign messes up with SQL?

Yes, I think the query doesn't get escaped properly

theshadowx:
I don't think it's sql, but rather Qt ,%1 is placement that is replaced by using .arg().

I haven't yet debugged the issue.

nuttyartist:
We need to further investigate these situations (based on @bjorn [comment](https://github.com/nuttyartist/notes/pull/280#issuecomment-688913866)):

* Migrating a note from the old format to the new database format that contains "%1" (either active note or trashed note)
* Deleting a note after putting "%1" in it.

All other query parameters are numbers, which do not have this problem.

zjeffer:
> Deleting a note after putting "%1" in it.

This doesn't trigger any bugs for me, might already be fixed.

> Migrating a note from the old format to the new database format that contains "%1" (either active note or trashed note)

I don't know how to do this, @nuttyartist maybe you can try?

bjorn:
@zjeffer My comment was based on the fact that bfdca20a8d230f1a7352960881d455cca52c57fa only fixed the issue in one place, but left for example `addNote` using the same broken approach for setting the query parameters, at the time:

https://github.com/nuttyartist/notes/blob/bfdca20a8d230f1a7352960881d455cca52c57fa/src/dbmanager.cpp#L197-L210

Note the chained `arg` calls, which can mess up once there are `%1` occurrences in the content, which will get replaced by `fullTitle` subsequently. However, this code was replaced in #302 (6d7f2a6dc7b051d3430fb9c1c6ae395a27b00676 and 4383bef654a406a5436dbafc7661c9781816992c, specifically), so I think this issue can be closed now.

[...]

PR #238, opened 2020-01-11
Fix notes don't get saved if they have % in them

Waqar144:
#234

### DIFF
--- a/src/dbmanager.cpp
+++ b/src/dbmanager.cpp
@@ -3,6 +3,7 @@
 #include <QTimeZone>
 #include <QDateTime>
 #include <QDebug>
+#include <QSqlError>
 #include <QtConcurrent>
 
 /*!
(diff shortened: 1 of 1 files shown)


nuttyartist:
Seems to work!

Since @theshadowx worked on the database, I'm waiting for his approval.

src/dbmanager.cpp
```diff
@@ -277,20 +278,22 @@ bool DBManager::updateNote(NoteData* note)
     int id = note->id();
     qint64 epochTimeDateModified = note->lastModificationdateTime().toMSecsSinceEpoch();
     QString content = note->content()
-            .replace("'","''")
+            .replace(QLatin1String("'"), QLatin1String("''"))
             .replace(QChar('\x0'), emptyStr);
     QString fullTitle = note->fullTitle()
-            .replace("'","''")
+            .replace(QLatin1String("'"), QLatin1String("''"))
```
bjorn: Oh btw, I think this replacement should be removed now. `bindValue` should already do any needed quoting, so now this will be duplicating single quotes.

src/dbmanager.cpp
```diff
@@ -277,20 +278,22 @@ bool DBManager::updateNote(NoteData* note)
...
             .replace(QChar('\x0'), emptyStr);
```
bjorn: Not sure why we need this? Was it also related to it breaking the SQL query, so possibly resolved with `bindValue`?

src/dbmanager.cpp
theshadowx: Is there anything said about this in documentation ?

Waqar144:
> Did you check whether there are any other SQL queries that should be transitioned to use `bindValue`?

Yes there are some more. Personally, I think all queries that Insert/Update should be prepared. It's always a good idea to be defensive with respect to data.

src/dbmanager.cpp
Waqar144: Indeed. That's what bindValue is doing so no point in keep these. It does lead to duplication of single quotes every time the note is saved.

src/dbmanager.cpp
Waqar144: Not sure about it myself.
Removing it doesn't cause any difference in the output. But maybe there's some specific case against which this needs to be tested or was put in, in the first place by @theshadowx

nuttyartist:
Let me know when it's ready.

src/dbmanager.cpp
bjorn: > Is there anything said about this in documentation ?

I couldn't immediate find it in Qt's documentation, but bound values are a common feature in SQL databases so Qt is probably just forwarding these functions to the underlying SQL API. It could even avoid the need for escaping entirely since the values are no longer part of the query. So it's good for security as well as performance.

Waqar144: From the `QSqlQuery` docs:

> QSqlQuery supports prepared query execution and the binding of parameter values to placeholders. Some databases don't support these features, so for those, Qt emulates the required functionality. For example, the Oracle and ODBC drivers have proper prepared query support, and Qt makes use of it; but for databases that don't have this support, Qt implements the feature itself, e.g. by replacing placeholders with actual values when a query is executed

And SQLite does support bindValues
https://sqlite.org/c3ref/bind_blob.html

nuttyartist:
Seems to work well, merging...

Issue #239, opened 2020-01-12
Not focusing on note's content when a user press Enter


nuttyartist:
It seems like because of #236 PR, when the user hit Enter it doesn't focus on the note's content. I guess it needs to be conditioned on whether a search is active or not - if it's not active: focus on the content. If it's active: go to the next found keyword. If the search is active but the user still wants to focus on a note's content, it should be using CTRL+Enter.

PR #240, opened 2020-01-13
Search fixes

Waqar144:
- Rewrote `highlightSearch` to use the available API
- Focus on textEdit on return key press

#239

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -2817,34 +2817,21 @@ void MainWindow::highlightSearch() const
 {
     QString searchString = m_searchEdit->text();
 
-    if(!searchString.isEmpty()){
-        m_textEdit->blockSignals(true);
+    if(searchString.isEmpty())
+        return;
 
-        QTextDocument *document = m_textEdit->document();
+    m_textEdit->moveCursor(QTextCursor::Start);
 
-        QTextCursor highlightCursor(document);
-        QTextCursor cursor(document);
+    QList<QTextEdit::ExtraSelection> extraSelections;
+    while(m_textEdit->find(searchString)) {
+        QTextEdit::ExtraSelection extra = QTextEdit::ExtraSelection();
+        extra.format.setBackground(Qt::yellow);
+        extra.cursor = m_textEdit->textCursor();
+        extraSelections.append(extra);
+    }
 
-        cursor.beginEditBlock();
-
-        QTextCharFormat colorFormat(highlightCursor.charFormat());
-        colorFormat.setBackground(Qt::yellow);
-
-        QTextCursor pos = document->find(searchString, 0);
-        if (!pos.isNull())
-            m_textEdit->setTextCursor(pos);
-
-        while (!highlightCursor.isNull() && !highlightCursor.atEnd()) {
-            highlightCursor = document->find(searchString, highlightCursor);
-
-            if (!highlightCursor.isNull()) {
-
-                highlightCursor.mergeCharFormat(colorFormat);
-            }
-        }
-
-        cursor.endEditBlock();
-
-        m_textEdit->blockSignals(false);
+    if (!extraSelections.isEmpty()) {
+        m_textEdit->setTextCursor(extraSelections.first().cursor);
+        m_textEdit->setExtraSelections(extraSelections);
     }
 }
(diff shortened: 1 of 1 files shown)


src/mainwindow.cpp
```diff
@@ -2817,34 +2827,21 @@ void MainWindow::highlightSearch() const
...
-        }
+    m_textEdit->moveCursor(QTextCursor::Start);
 
-        cursor.endEditBlock();
+    QList<QTextEdit::ExtraSelection> extraSelections;
+    while(m_textEdit->find(searchString)) {
+        QTextEdit::ExtraSelection extra = QTextEdit::ExtraSelection();
+        extra.format.setBackground(Qt::yellow);
+        extra.cursor = m_textEdit->textCursor();
+        extraSelections.append(extra);
```
bjorn: The format is always the same, so we can share it:

```c++
QTextCharFormat highlightFormat;
highlightFormat.setBackground(Qt::yellow);

QList<QTextEdit::ExtraSelection> extraSelections;
while(m_textEdit->find(searchString))
    extraSelections.append({ highlightFormat, m_textEdit->textCursor() });

**Comment by nuttyartist** [comment]
Two things:

1. Ctrl-enter doesn't seem to work on my computer (Mac) while searching, it doesn't focus and show the cursor on the selected note. Maybe it's a problem specific to Mac or my computer.
2. It doesn't relate to this PR, but maybe you can solve it here. It seems like the selected note color doesn't change when a user enter into the note. It should change to m_ActiveColor(218, 233, 239) from notewidgetdelegate.cpp. Can you solve it?

File: src/mainwindow.cpp [hunk 2]
```diff
@@ -2743,6 +2743,20 @@ bool MainWindow::eventFilter(QObject *object, QEvent *event)
...
+        auto *keyEvent = static_cast<QKeyEvent *>(event);
+        if (keyEvent->key() == Qt::Key_Return && m_searchEdit->text().isEmpty()) {
+            setFocusOnText();
+            return true;
+        } else if (keyEvent->key() == Qt::Key_Return &&
+                   keyEvent->modifiers().testFlag(Qt::ControlModifier)) {
+            setFocusOnText();
+            return true;
+        }
+        return QMainWindow::eventFilter(object, event);
```
bjorn: Why is this line here? I guess it could just be removed because there is already `return QObject::eventFilter(object, event);` below (which could also use `QMainWindow`, but `QMainWindow` and `QWidget` don't override this virtual anyway).

src/mainwindow.cpp
Waqar144: Done

Waqar144:
> Two things:
> 
>     1. Ctrl-enter doesn't seem to work on my computer (Mac) while searching, it doesn't focus and show the cursor on the selected note. Maybe it's a problem specific to Mac or my computer.
> 
>     2. It doesn't relate to this PR, but maybe you can solve it here. It seems like the selected note color doesn't change when a user enter into the note. It should change to m_ActiveColor(218, 233, 239) from notewidgetdelegate.cpp. Can you solve it?

Both done.

src/mainwindow.cpp
Waqar144: Ah, yes. I noticed it later. I will just convert into `break;` 
thanks

nuttyartist:
Some problems:

1. Ctrl-Enter while searching doesn't focus on textEdit and put the cursor there for me, it
2. When searching and focusing on textEdit using the mouse (currently selected note content), it doesn't change the color to m_ActiveColor for some reason. But it does when using Ctrl-Enter (although, as stated in 1, it doesn't put the cursor there and focus).

Waqar144:
Ctrl + Enter is working for me on Linux 🤔

Waqar144:
The mousePressEvent doesn't get caught in eventFilter for some reason. So yeah, that part doesn't work, with / without searching (clicking into textEdit).

nuttyartist:
Hmm... Do you have an idea why?

And about Ctrl+Enter, it works also for me but doesn't put the cursor on textEdit while searching, only when not searching. But it still might be a Mac problem.

nuttyartist:
Oh ok, it put the cursor next to the found text, just noticed. Works well! Merging...

Thanks @Waqar144!

nuttyartist:
This PR appears to create a serious bug where pressing Enter doesn't work in a notes' content - therefore I can't create new lines. I created a new issue for this #244

Waqar144:
> Hmm... Do you have an idea why?

Maybe because there is already an override for `mousePressEvent` in mainwindow.cpp. If we subclass QTextEdit and handle the mouse press event there, it might work. And that is something that is eventually going to have to happen, for Rich Text/ text editing commands etc.

PR #241, opened 2020-01-14
Use :focus selector in search edit stylesheet

bjorn:
Avoids some duplicated CSS code.

I've removed the "frameWidth" line because it was not having any effect anyway due to "padding-right" being duplicated and because a focus change would override the whole stylesheet. It was already there in the initial commit, so I'm not sure why it was there.


nuttyartist:
Gonna review this later/tomorrow (:

nuttyartist:
👍 Merging...

PR #242, opened 2020-01-14
Travis CI: Removed the failing 'trusty' build

bjorn:
Since the addition of qmarkdowntextedit, APIs are used that were added in Qt 5.3. Unless we're interested in restoring compatibility with Qt 5.2 (probably possible by changing qmarkdowntextedit), I would suggest we simply drop this Trusty build, who's end of support was reached in April 2019.

I've also added a Qt version check, along with a Qt 5.3 build to check whether Notes really does compile against that Qt version. Of course, we may want to raise the minimum Qt version further, say to Qt 5.5, which is the version in the current oldest supported Ubuntu (xenial).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,11 +3,11 @@ dist: trusty
 language: cpp
 
 env:
+  - QT_BASE=53
   - QT_BASE=55
   - QT_BASE=56
   - QT_BASE=57
   - QT_BASE=58
-  - QT_BASE=trusty
   - QT_BASE=xenial
   - QT_BASE=yakkety
   - QT_BASE=zesty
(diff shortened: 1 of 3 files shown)


Waqar144:
Even Qt 5.3 is pretty much ancient by now. We only support it in qmarkdownedit because of Debian 8 and older fedora versions and Debian 8 is going EOL in June this year. Qt 5.2 in the meanwhile will cause many issues. So dropping it seems like a good idea.

nuttyartist:
> Of course, we may want to raise the minimum Qt version further, say to Qt 5.5, which is the version in the current oldest supported Ubuntu (xenial).

I think we should do that.

nuttyartist:
You need to make some changes?

bjorn:
There are currently no further changes necessary. I'd say we can wait with raising the minimum version to Qt 5.5 and drop the 5.3 build when there is some code that needs the newer version.

nuttyartist:
So we're good to merge?

dimeldo:
Merging... (:

nuttyartist:
Merging...

PR #243, opened 2020-01-14
AppVeyor: Updated Qt 5.7 builds to Qt 5.9

bjorn:
Fixes failing AppVeyor build because Qt 5.7 is not available for msvc2015_64. In addition 5.9 would be preferable as an LTS release.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -9,9 +9,9 @@ environment:
     qt: 5.6
     compiler: msvc2015
 
-  - name: win32_57
+  - name: win32_59
     platformEnv: amd64_x86
-    qt: 5.7
+    qt: 5.9
     compiler: msvc2015
 
   # MSVC x64
@@ -20,9 +20,9 @@ environment:
     qt: 5.6
     compiler: msvc2015_64
 
-  - name: win64_57
+  - name: win64_59
     platformEnv: amd64
-    qt: 5.7
+    qt: 5.9
     compiler: msvc2015_64
 
   # MinGW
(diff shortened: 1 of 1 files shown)


nuttyartist:
>  In addition 5.9 would be preferable as an LTS release.

What do you mean? So we'll support from only from 18.04?

Waqar144:
> What do you mean? So we'll support from only from 18.04?

Nah, he means LTS release of Qt.
Version 5.9 was an LTS release.

bjorn:
Right, Qt 5.6, 5.9 and 5.12 were LTS releases and received a much longer period of support and bugfixes. By doing releases against those versions, users may benefit from a more stable experience.

Actually the builds should probably be updated to Qt 5.9 and 5.12, dropping 5.6 which is no longer supported (we can also keep 5.6, since it's the last Qt version that still supported Windows XP). But with this change I just wanted to make the AppVeyor CI build to succeed again.

nuttyartist:
@bjorn Gotcha, yeah you bump it to Qt 5.9. We'd still support Windows 7, right?

@bjorn and @Waqar144 You're more than welcome to join our Slack chat with this link: https://join.slack.com/t/notes-app-team/shared_invite/enQtOTExMDA4MTY0ODA3LTk1MDViYzRlNTI5ZTAxZjMwYzY5ZDI4ZjkxYjIyNDQ0NGFlNDZjMDNjNGI3MjRkODMwOTRjODIwYTc4MmM0ZWY

theshadowx:
It's not about the version of Qt, it's about the defaut Qt version that comes with Ubuntu. As we don't generate ubuntu install package that needs dependencies that doesn't exist in the mainstream repository

nuttyartist:
@theshadowx What do you mean? What are the implication of that? And do you support these PR changes?

theshadowx:
My mistake, I thought this was for travis

nuttyartist:
Okay, merging...

Thanks, @bjorn!

Issue #244, opened 2020-01-28
Can't create new lines because of PR #240


nuttyartist:
The PR appears to create a serious bug where pressing Enter doesn't work in a notes' content - therefore I can't create new lines.

@Waqar144

PR #245, opened 2020-01-28
Fix return key not working #244

Waqar144:
#244

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -2753,7 +2753,7 @@ bool MainWindow::eventFilter(QObject *object, QEvent *event)
         auto *keyEvent = static_cast<QKeyEvent *>(event);
         if (keyEvent->key() == Qt::Key_Return && m_searchEdit->text().isEmpty()) {
             setFocusOnText();
-            return true;
+			break;
         } else if (keyEvent->key() == Qt::Key_Return &&
                    keyEvent->modifiers().testFlag(Qt::ControlModifier)) {
             setFocusOnText();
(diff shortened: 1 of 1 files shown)


src/mainwindow.cpp
```diff
@@ -2753,7 +2753,7 @@ bool MainWindow::eventFilter(QObject *object, QEvent *event)
         auto *keyEvent = static_cast<QKeyEvent *>(event);
         if (keyEvent->key() == Qt::Key_Return && m_searchEdit->text().isEmpty()) {
             setFocusOnText();
-            return true;
+			break;
```
bjorn: Just suggesting to keep the indentation consistent:
```suggestion
            break;
```

Waqar144: Alright, thanks.

Waqar144: How about we just not put a break in here? There's a `break` after the 'if` condition which will do the same.

Waqar144:
Also fixed row color not changing when clicking on `textEdit`. I must admit that the `if` ladder here is rather perplexing:

https://github.com/nuttyartist/notes/blob/c7a8687abd2c73996b05a17f35f0d43b11805ee5/src/mainwindow.cpp#L2670-L2674

nuttyartist:
> @nuttyartist
> @bjorn
> Also fixed row color not changing when clicking on `textEdit`. I must admit that the `if` ladder here is rather perplexing:
> 
> https://github.com/nuttyartist/notes/blob/c7a8687abd2c73996b05a17f35f0d43b11805ee5/src/mainwindow.cpp#L2670-L2674

Hahaha

Cool, merging!

Issue #248, opened 2020-02-03
Can't locate Travis generated packages, for 1.5.0 release


nuttyartist:
For some reason, I can't find the links to where Travis upload the generated packages. For example for the AppImage package: https://travis-ci.org/nuttyartist/notes/jobs/643738967

@Waqar144, @theshadowx, @bjorn Can you help?

I want to release version 1.5.0 today/tomorrow.

theshadowx:
If I recall correctly, the package will be generated only when dev branch is merged the the master one. Need to be checked !

nuttyartist:
Is it a requirement of Travis? Can we change that?

bjorn:
> If I recall correctly, the package will be generated only when dev branch is merged the the master one. Need to be checked !

[According to this section](https://github.com/nuttyartist/notes/blob/bdc6dbf17a2c9a7ea36fee6956ef4b7c617a5729/.travis.yml#L79-L103) the builds are always uploaded, just with different names depending on whether it's building the `master` branch or another branch.

However, after successful upload transfer.sh is supposed to output the URL, and instead the log reads "Could not save metadata" three times. This appears to be a problem at Transfer.sh, see the error in their server-side code here:

https://github.com/dutchcoders/transfer.sh/blob/b90bd685e75bb77563e3749dcf0729a8aabb7bb3/server/handlers.go#L322-L324

I'm not sure what triggers this, but I guess [setting up an alternative deployment method](https://docs.travis-ci.com/user/deployment) would be a solution if you want to release fast. For example, why not [upload directly to GitHub](https://docs.travis-ci.com/user/deployment/releases/)?

nuttyartist:
@bjorn Will it upload directly to Github releases on every commit? I used to see these repos with an alarming number of release versions. If we can avoid this clutter, the better. Or maybe you're not talking about this? There's also this "Github Workflow" thing, but I don't know how straightforward it is to move from Travis to this.

bjorn:
> Will it upload directly to Github releases on every commit?

Only if you tell it to, but generally you'll set up this deploy rule to deploy only for tags. That's what this part in the deploy section does:
```yml
  on:
    tags: true
```

> There's also this "Github Workflow" thing, but I don't know how straightforward it is to move from Travis to this.

[...]

PR #249, opened 2020-02-05
Added deployment to GitHub on tags

bjorn:
See issue #248.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -105,3 +105,18 @@ after_success:
 notifications:
   email: false
 
+deploy:
+  provider: releases
+  skip_cleanup: true
+  draft: true
+  tag_name: "$TRAVIS_TAG"
+  api_key:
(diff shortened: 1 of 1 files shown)


bjorn:
So, after merging this, the builds done on Travis CI based on the `v1.5.0` tag _should_ be directly uploaded to GitHub. I've set `draft: true` so that the release will not be immediately published, so that we can also test this somewhat silently.

Let me know when you want me to push this and a tag to test the process. We can just delete the tag again when something goes wrong.

nuttyartist:
Let's do it!

bjorn:
Ok, I merged, tagged `v1.5.0`, had to [fix a typo](f1bc03fcb161fb1d39e4dfa549a87e64754e934c), and retagged. Let's see what happens now...

bjorn:
Alright, next problem is that the `.deb` is still always called `notes_1.0.0_amd64.deb` regardless of the target version of Ubuntu, whoops.

bjorn:
Now packages should be getting renamed (346841dd99a5fdf955d89c8714b04d08730de5a7).

nuttyartist:
So it didn't work and you removed the tag? I just saw your new PR, gonna respond there.

Issue #195, opened 2018-02-03
Automatic startup


nuttyartist:
We should implement that Notes would run when the system is running on startup. This way people could jot down their thoughts (ideally, when working) just by pressing the "Super Key" + "N".

bjorn:
So, this was actually implemented in 3978e47d245ff84bbb708fe78722a09617cac99b, but the issue didn't close because that commit referenced #193 instead of #195.

But, later on autostart was disabled for macOS in #232, though without stating the problem it was fixing. @theshadowx, can you shed some light on this?

theshadowx:
@bjorn The QAutostart is not supported in mac, as mentioned in readme: https://github.com/b00f/qautostart

The reason is that the compilation fails as QProcess doesn't work in mac.

bjorn:
@theshadowx Ah, I was confused because the code did seem to handle macOS. But, it can't be that in general `QProcess` is not working on Mac. I'm using it in Tiled as well to run custom commands set up by the user and it does also work on macOS.

theshadowx:
I have forked QAutostart and added in qautostart.cpp
```
include <QProcess> 
```

It did compile https://github.com/theshadowx/notes/runs/471152271?check_suite_focus=true. I don't Mac machine so I don't know if the functionality is working.

bjorn:
> I have forked QAutostart and added in qautostart.cpp

I don't really understand, `#include <QProcess>` [is already there](https://github.com/b00f/qautostart/blob/3bcab269a6a8ba4a4824d73bbf8d9b87dd62cd18/src/qautostart.cpp#L29
). I'll see if I can find some time to test the functionality.

theshadowx:
Indeed, maybe ```include <QProcess>``` should be at the end. As when I just commented ifdef guard and compile with github ci, it failed :

nuttyartist:
I can test it if needed, @theshadowx. Let me know the artifact link.

theshadowx:
Use this branch from my repo https://github.com/theshadowx/notes/tree/qprocess_mac

nuttyartist:
Didn't find any artifacts: https://github.com/theshadowx/notes/actions/runs/45817715

theshadowx:
> Didn't find any artifacts: https://github.com/theshadowx/notes/actions/runs/45817715

You should compile it in your machine

nuttyartist:
Aw oops, haha sorry. Will do and report later.

bjorn:
[...]

PR #258, opened 2020-02-27
Enabled autostart option on macOS

bjorn:
This change makes it compile, but the commands executed by QAutostart on macOS do not seem to work, though they appear close to something that should work.

I can look deeper into it and test things later.

See issue #195.

### DIFF
--- a/src/Notes.pro
+++ b/src/Notes.pro
@@ -24,10 +24,7 @@ greaterThan (QT_MAJOR_VERSION, 4): QT += widgets
 include ($$PWD/../3rdParty/qxt/qxt.pri)
 include ($$PWD/../3rdParty/QSimpleUpdater/QSimpleUpdater.pri)
 include ($$PWD/../3rdParty/qmarkdowntextedit/qmarkdowntextedit.pri)
-
-!macx {
 include ($$PWD/../3rdParty/qautostart/src/qautostart.pri)
-}
 
 SOURCES += \
     $$PWD/main.cpp\
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -1047,7 +1047,6 @@ void MainWindow::onDotsButtonClicked()
     QAction* checkForUpdatesAction = mainMenu.addAction(tr("Check For Updates"));
     connect (checkForUpdatesAction, &QAction::triggered, this, &MainWindow::checkForUpdates);
 
-#if !defined (Q_OS_MACOS)
     // Autostart
     QAction* autostartAction = mainMenu.addAction(tr("Start automatically"));
     connect (autostartAction, &QAction::triggered, this, [&]() {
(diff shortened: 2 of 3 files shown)


nuttyartist:
Yes, it doesn't seem to add Notes to Login Items.

bjorn:
I've played around with this some more on my MacMini and got it to work with a few changes. I've opened a pull request at https://github.com/b00f/qautostart/pull/1.

Unfortunately, one thing that doesn't appear to work with Notes is the `hidden: true` option. I tried to use this because it is not possible to pass parameters like `--autostart` to login items. But Notes is showing its main window on startup anyway. I wonder if we're doing something non-standard that may cause this not to work, or whether the Hidden option has issues with Qt-based applications in general.

theshadowx:
@bjorn I think it may be caused by override of the mainwindow behavior:

https://github.com/nuttyartist/notes/blob/dev/src/mainwindow.cpp#L223

We are setting the window frameless, in order to customize the window trafic buttons. I really don't like it because it makes the not taking advantage of the window manager features of the DE.

bjorn:
@theshadowx I can do some testing when I have some time again. For now I guess we could just support setting autostart this way and leave the visible window on startup as a known issue. At least it looks like we'll get [some improvements for frameless windows](https://www.qt.io/blog/custom-window-decorations) soon (though on macOS, they currently only support starting a move operation).

bjorn:
@nuttyartist My fix for macOS has been merged in the qautostart repository. Please try if it works now.

theshadowx:
> @theshadowx I can do some testing when I have some time again. For now I guess we could just support setting autostart this way and leave the visible window on startup as a known issue. At least it looks like we'll get [some improvements for frameless windows](https://www.qt.io/blog/custom-window-decorations) soon (though on macOS, they currently only support starting a move operation).

Cool, I saw that article,  when we use the 5.15, we will remove the chunk of code that dealing with mainwindow moving with cursor.

nuttyartist:
> At least it looks like we'll get some improvements for frameless windows soon (though on macOS, they currently only support starting a move operation).

That's awesome news! Finally... I wish they'll be fast with doing the same with native shadows.

> Cool, I saw that article, when we use the 5.15, we will remove the chunk of code that dealing with mainwindow moving with cursor.

Sounds good!

@bjorn Thanks for keeping on digging this problem. I'll test it again very soon. What exactly is the problem? That Notes is showing up visible on startup rather than hidden?

bjorn:
> @bjorn Thanks for keeping on digging this problem. I'll test it again very soon. What exactly is the problem? That Notes is showing up visible on startup rather than hidden?

Yes, that's the remaining problem. Other than that, at least it can autostart now.

nuttyartist:
Working! Please merge. Thank you @bjorn!

nuttyartist:
It also seems to default automatically to hide on startup:

Screen Shot 2020-03-11 at 7 07 59 PM

bjorn:
> It also seems to default to automatically hide on startup:

Yes, I patched up QAutostart so that it would automatically set the hidden flag. Though as I wrote above, this flag is currently not actually suppressing the window on startup for me.

nuttyartist:
Oh ok, gotcha.

Issue #257, opened 2020-02-27
Problem generating snap package


nuttyartist:
As mentioned in #248, there still some problems generating snap packages using Travis.

nuttyartist:
@theshadowx @bjorn We still don't have snap artifacts for v1.5.0?

theshadowx:
I have been able to make snap works correctly, I will upadre the .pro file also as snap folder will be moved to the root folder. Here is the branch specific to the snap modifications. It's not up to date, I'll update it this evening:
https://github.com/theshadowx/notes/tree/snapcraft

nuttyartist:
Awesome, then we'll continue to release v1.5.0.

theshadowx:
I think this should be closed

nuttyartist:
Okay, is it fixed? I can't see a link for the snap in https://travis-ci.org/github/nuttyartist/notes/jobs/662820085, but I guess it's because the release isn't tagged yet?

theshadowx:
The package is generated at the bottom : https://transfer.sh/e8xr4/notes_1.0.0-git5ca3205_amd64.snap

bjorn:
Unfortunately the link is already giving 404 error. However, indeed the original issue described in https://github.com/nuttyartist/notes/issues/248#issuecomment-588522356 that prompted the opening of this issue was fixed by #259, so we can close this.

Issue #185, opened 2017-11-02
Update snap to 1.0.0


Ads20000:
The snap in the store (`snap info notes`) is `0.9.0~gitb6e3b34 (4)` not `1.0.0`. Please can you push the snap to the store so that those that have it are automatically updated?

With the snap, just use (with `snapcraft` installed (`sudo snap install snapcraft --candidate`)):
`snapcraft push notes_1.0.0_amd64.snap`
`snapcraft release notes 5 edge`
test with `sudo snap install notes --edge` (check the version in `edge` is correct with `snap info notes`)
if it works, then use `snapcraft release notes 5 stable`

[Source](https://docs.snapcraft.io/build-snaps/publish)

You may also want to get automatic builds going with build.snapcraft.io :)

theshadowx:
Hi @Ads20000 
We want to update it, but we don't have the credentials to allow us to do it.

@dplanella, please could you do it, or pass the account (usrname/pw) to @nuttyartist. thanks

bjorn:
Still not uploaded now that we have the credentials? Should be a quick one to push since the snap is already done.

nuttyartist:
You're right! I just updated the page with new info, but I don't have a Linux machine ready (seems like i need one) Can I give you my user credentials over email and you'll push it? 

Also, some really awesome metrics that I only now noticed. Since we don't track analytics of active users it's cool to see that we have more than 15K weekly active users (only with snap on ubuntu). And the number is growing.

![Screen Shot 2020-02-27 at 12 21 00 PM](https://user-images.githubusercontent.com/16375940/75435464-bc0bb000-595b-11ea-8f41-94877ff4c6c2.png)

![Screen Shot 2020-02-27 at 12 21 13 PM](https://user-images.githubusercontent.com/16375940/75435473-bf9f3700-595b-11ea-90ae-cb35b17d96a0.png)

theshadowx:
The snap folder in packages should be moved to the root folder so that snapcraft can take it into considerarion

nuttyartist:
Can I send you over my snapstore account credentials?

theshadowx:
Also please wait till the snapcraft is updated. The version 1.0.0 uses sqlite database which won't work with the current snapcraft script

theshadowx:
If you want

nuttyartist:
Oh ok, so we shall wait for 1.5.0?

theshadowx:
Ya

bjorn:
[...]

PR #259, opened 2020-03-13
Fixing Snapcraft

theshadowx:
Using QSlite Database.
Using Host Theme.

 #257 #185


theshadowx:
I will fix the issue concerning travis

snap/snapcraft.yaml
```diff
@@ -0,0 +1,90 @@
+name: notes
+# Note on version: $UPSTREAM_VERSION~.git$GIT_COMMIT_ID
+summary: Note taking application, write down your thoughts.
+description: |
+  Notes is a tool for the things you need to write down off your
+  brain. It is your place of expressing yourself. Notes is an open
+  source and cross-platform project, and you're welcome to be part
+  of it and contribute. We hope you will love it as much as we do.
+version: latest
+version-script: GIT_BRANCH=$(git rev-parse --abbrev-ref HEAD);
```
bjorn: Note that `version-script` is [deprecated](https://snapcraft.io/docs/deprecation-notices/dn10), so while you're updating the `snapcraft.yml` file you may want to use the new method for setting the version.

For Tiled I ended up with:

https://github.com/bjorn/tiled/blob/9cc4bcd6322cd2d7b73a2de1f8a594749bd29837/snap/snapcraft.yaml#L37-L39

theshadowx: I wasn t aware of that, your repo is one of my references ;) I'll update that.

bjorn: Hehe, though in general you have to be a bit careful using it as reference. Most of the snap packaging stuff is just guesswork or copy/paste... Please let me know if you find anything that could be improved. :-)

bjorn:
> I will fix the issue related to travis

It seems the Travis build may be broken because of the version change in `Notes.pro`. Since it is anyway unrelated to getting the snap to build I would suggest to take that change out of this pull request and do it separately. There are a lot more places that will need to be updated regarding the version (see bb64fe95d820b5d07379d5b2fc238c4030b57118).

theshadowx:
@bjorn sure, I'm reverting it to 1.0.0 to be updated later.

nuttyartist:
@theshadowx Feel free to merge it when it's ready.

nuttyartist:
@bjorn So we can change the version after this PR?

bjorn:
> @bjorn So we can change the version after this PR?

Sure, let's open another PR for updating the version and fix any issues this causes there.

PR #261, opened 2020-03-18
Bumping app version to 1.5.0

bjorn:


bjorn:
The snap is still called `notes_1.0.0_amd64.snap` because it depends on the release being tagged. Unfortunately, the uploads again show "Could not save metadata", which indicates a problem at transfer.sh so right now I can't test the snap or Debian packages.

theshadowx:
Deb dependecies should be updated

bjorn:
@theshadowx I'm not really familiar with Debian packaging. If you think it should be done as part of the version bump, please feel free to push relevant commits here.

nuttyartist:
@bjorn @theshadowx Can we progress on this?

bjorn:
I'm afraid I don't have enough experience with Debian packaging and too busy to help out currently. Also it is unclear to me how the needed Deb updates relate to this pull request. I think it could just be merged independently.

nuttyartist:
Excuse me for not responding for a long time. Yes, let's merge this.

nuttyartist:
So, after this merge we should have all the artifacts generated properly (except Debian package which we need to update its dependencies)?

PR #263, opened 2020-03-24
Use window manager frame on Linux

### LINKED ISSUE #262: "Always on top" not working in Ubuntu
This feature has been added post 1.0.0
In Ubuntu 18.04, "Always on top" is not working,  even when the option is checked, the gui still gets hidden by others.

bjorn:
Also, hide the custom "Always stay on top" option, which did not work on GNOME and possibly other window managers. The window manager's native option to keep a window on top can now be used.

Closes #262

This is how it looks on Ubuntu:

**Before**
notes-no-frame
**After**
notes-native-frame

The screenshots also shows some DPI issues, probably due to hardcoded sizes that are not adapting to the font size. I guess it would be nice to fix that for 1.5 as well.

I also noticed the "Hide Notes" action in the tray icon is not working for me. But that is also unrelated to this patch.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -34,9 +34,12 @@ MainWindow::MainWindow(QWidget *parent) :
     m_autoSaveTimer(new QTimer(this)),
     m_settingsDatabase(Q_NULLPTR),
     m_clearButton(Q_NULLPTR),
+#ifndef Q_OS_LINUX
     m_greenMaximizeButton(Q_NULLPTR),
     m_redCloseButton(Q_NULLPTR),
     m_yellowMinimizeButton(Q_NULLPTR),
+    m_trafficLightLayout(Q_NULLPTR),
+#endif
     m_newNoteButton(Q_NULLPTR),
     m_trashButton(Q_NULLPTR),
     m_dotsButton(Q_NULLPTR),
@@ -48,7 +51,6 @@ MainWindow::MainWindow(QWidget *parent) :
     m_restoreAction(new QAction(tr("&Hide Notes"), this)),
     m_quitAction(new QAction(tr("&Quit"), this)),
     m_trayIconMenu(new QMenu(this)),
-    m_trafficLightLayout(Q_NULLPTR),
     m_noteView(Q_NULLPTR),
     m_noteModel(new NoteModel(this)),
     m_deletedNotesModel(new NoteModel(this)),
@@ -157,6 +159,7 @@ void MainWindow::setMainWindowVisibility(bool state)
  */
 void MainWindow::paintEvent(QPaintEvent* event)
 {
+#ifndef Q_OS_LINUX
     QPainter painter(this);
     painter.save();
 
@@ -174,6 +177,7 @@ void MainWindow::paintEvent(QPaintEvent* event)
     dropShadow(painter, ShadowType::Radial, ShadowSide::BottomLeft );
 
     painter.restore();
+#endif
     QMainWindow::paintEvent(event);
 }
 
@@ -215,24 +219,23 @@ MainWindow::~MainWindow()
  */
 void MainWindow::setupMainWindow()
 {
-#ifdef Q_OS_LINUX
-    this->setWindowFlags(Qt::Window | Qt::FramelessWindowHint);
-    this->setAttribute(Qt::WA_TranslucentBackground);
-#elif _WIN32
+#ifdef _WIN32
     this->setWindowFlags(Qt::CustomizeWindowHint);
-#elif __APPLE__
+#elif defined(__APPLE__)
     this->setWindowFlags(Qt::Window | Qt::FramelessWindowHint);
     this->setAttribute(Qt::WA_TranslucentBackground);
-#else
+#elif !defined(Q_OS_LINUX)
 #error "We don't support that version yet..."
 #endif
 
+#ifndef Q_OS_LINUX
     m_greenMaximizeButton = new QPushButton(this);
     m_redCloseButton = new QPushButton(this);
     m_yellowMinimizeButton = new QPushButton(this);
     m_trafficLightLayout.addWidget(m_redCloseButton);
     m_trafficLightLayout.addWidget(m_yellowMinimizeButton);
     m_trafficLightLayout.addWidget(m_greenMaximizeButton);
+#endif
 
 #ifdef _WIN32
     m_trafficLightLayout.setSpacing(0);
@@ -413,9 +416,11 @@ void MainWindow::setupTitleBarButtons()
                  "  padding: 0px; "
                  "}");
 
+#ifndef Q_OS_LINUX
     m_redCloseButton->setStyleSheet(ss);
     m_yellowMinimizeButton->setStyleSheet(ss);
     m_greenMaximizeButton->setStyleSheet(ss);
+#endif
 
 #ifdef _WIN32
     m_redCloseButton->setIcon(QIcon(QStringLiteral(":images/windows_close_regular.png")));
@@ -427,9 +432,11 @@ void MainWindow::setupTitleBarButtons()
     m_greenMaximizeButton->setIconSize(QSize(28, 16));
 #endif
 
+#ifndef Q_OS_LINUX
     m_redCloseButton->installEventFilter(this);
     m_yellowMinimizeButton->installEventFilter(this);
     m_greenMaximizeButton->installEventFilter(this);
+#endif
 }
 
 /*!
@@ -442,6 +449,7 @@ void MainWindow::setupSignalsSlots()
     // actions
     // connect(rightToLeftActionion, &QAction::triggered, this, );
     //connect(checkForUpdatesAction, &QAction::triggered, this, );
+#ifndef Q_OS_LINUX
     // green button
     connect(m_greenMaximizeButton, &QPushButton::pressed, this, &MainWindow::onGreenMaximizeButtonPressed);
     connect(m_greenMaximizeButton, &QPushButton::clicked, this, &MainWindow::onGreenMaximizeButtonClicked);
@@ -451,6 +459,7 @@ void MainWindow::setupSignalsSlots()
     // yellow button
     connect(m_yellowMinimizeButton, &QPushButton::pressed, this, &MainWindow::onYellowMinimizeButtonPressed);
     connect(m_yellowMinimizeButton, &QPushButton::clicked, this, &MainWindow::onYellowMinimizeButtonClicked);
+#endif
     // new note button
     connect(m_newNoteButton, &QPushButton::pressed, this, &MainWindow::onNewNoteButtonPressed);
     connect(m_newNoteButton, &QPushButton::clicked, this, &MainWindow::onNewNoteButtonClicked);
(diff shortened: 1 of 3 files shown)


theshadowx:
It's better to leave this after 1.5.0

bjorn:
> It's better to leave this after 1.5.0

Why? Do you think it's better to leave the broken "Always stay on top" action in for 1.5, or do you have another suggestion about what to do for 1.5?

theshadowx:
@bjorn we just remove it from menu item for now
And in post 1.5.0 we will tacle the gui issues
What do you think ?

bjorn:
> @bjorn we just remove it from menu item for now
> And in post 1.5.0 we will tacle the gui issues
> What do you think ?

Well, you suggested we use the window manager frame instead to still have the keep-on-top state available and I'd prefer this to just hiding the action.

I don't care that much either way though. Maybe @nuttyartist can help make the call for 1.5.

nuttyartist:
Hey all :)

@bjor, What DE did you use while taking these screenshots?

I'm siding with @theshadowx. The native option ruins the aesthetics of Notes. I know though, that if we are to enable native decoration after 1.5, it will not look pretty either, but I would prefer to do some user research to see what Linux users think and to test on different DE's to check how it looks. Right now, my intuition is that people download our app, mainly on Linux, because they see someone finally cared for design and not merely about functionality. So let's wait with this till after 1.5.0.

BTW, excuse my delay in responses lately. I just finished a marathon on a new thing I'm building, so now I will have more time to dedicate for Notes.

theshadowx:
The dowloads from github shows that Notes is a also used in windows.
This is for only the version 1.0.0

nuttyartist:
> The dowloads from github shows that Notes is a also used in windows.
This is for only the version 1.0.0

Yes, Indeed. Most of our users are Windows users (because of the OS share in Google searches).

bjorn:
> The native option ruins the aesthetics of Notes. I know though, that if we are to enable native decoration after 1.5 it will not look pretty either, but I would prefer to do some user research to see what Linux users think and to test on different DE's to check how it looks. Right now, my intuition is that people download our app mainly on Linux because they see someone finally cared for design and not merely about functionality.

It's a matter of opinion. As a Linux user, I like the consistency with other applications regarding the window frame, the drop shadow (the custom one looks wrong) and the position of the buttons.

What about I try to find some time to turn this into a user option? That should also make it easier to do user research. :-)

Waqar144:
> It's a matter of opinion. As a Linux user, I like the consistency with other applications regarding the window frame, the drop shadow (the custom one looks wrong) and the position of the buttons.

Me as well. Linux already provides a lot of options for making your windows look like whatever you want.

theshadowx:
> It's a matter of opinion. As a Linux user, I like the consistency with other applications regarding the window frame

absolutely !

> the drop shadow (the custom one looks wrong) and the position of the buttons.

Could you elaborate ?

Waqar144:
> Could you elaborate ?

I assume he means "it looks out of place" and not like other windows.

nuttyartist:
> What about I try to find some time to turn this into a user option? That should also make it easier to do user research. :-)

I'm totally in for this idea. Sounds good to me. Make the current custom decoration default but have an option for native window frame for Linux users using the -> View->Native window frame. I made an issue #264.

nuttyartist:
@bjorn Can you implement #264, our last issue for v1.5?

bjorn:
@nuttyartist I can try to find some time for it, but not this week.

nuttyartist:
@bjorn Got some progress?

bjorn:
@nuttyartist I really wanted to get Tiled 1.4 out first and yesterday I finally finished that release. I'll see if I can get around to this soon now.

nuttyartist:
You have done it well! Just two things:

1. Seems like there are conflicting files. Can you resolve it?
2. When a user chooses to use the native window frame, Can you close the gap that was originally for the 3 traffic lights? Talking about the red part in the following image:

Screen Shot 2020-06-22 at 1 49 14 PM

nuttyartist:
I also appreciate that you make sure to save the option the user chose for next time a user starts the program. 👍

nuttyartist:
@bjorn Any progress?

bjorn:
@nuttyartist I've resolved the conflicts, fixed a small issue I noticed on Linux and removed the empty space above the search box when the native frame is enabled.

I've only removed the space above the search box, because removing it also above the note text field would result in inconsistent margins around the controls at the top (where they would be rather closer to the top than to the sides).

Note that before merging, this feature also still needs to be tested on Windows.

nuttyartist:
Awesome, @bjorn! I will make sure to test it tomorrow and then merge (: 👍

nuttyartist:
Okay, so you successfully resolved previous issues, but now the textEditor needs go up a bit up. The newNoteButton, editorDateLabel, trashButton, and dotsButton needs to align with lineEdit.
Screen Shot 2020-07-01 at 8 55 49 AM

nuttyartist:
@bjorn Let me know when you finish this.

bjorn:
@nuttyartist As I wrote before:

> I've only removed the space above the search box, because removing it also above the note text field would result in inconsistent margins around the controls at the top (where they would be rather closer to the top than to the sides).

Let me know if you think the space should still be removed.

nuttyartist:
I'm sorry, I haven't noticed! Do you have a picture of that? Or a source I can test?

bjorn:
Comparison:

notes1 notes2

I'm a little undecided, but the one on the left seems cleaner to me. On the right, the margins to the left/right of the header do not match with the margin at the top.

nuttyartist:
It's definitely should be the right picture, where the elements are alinged with lineEdit. 

> On the right, the margins to the left/right of the header do not match with the margin at the top.

This is because of the native window. A solution might be, when native window is on, to create a spacer that will move down the entire elements a bit. So everything would still align properly.

bjorn:
> This is because of the native window. A solution might be, when native window is on, to create a spacer that will move down the entire elements a bit. So everything would still align properly.

So, to me, a possible solution is the version on the left. Another option might be to reduce the spacing on the left/right of the editor, such that it matches with the reduced space at the top.

When we use a spacer to move everything down, wouldn't we get the same situation we had initially at https://github.com/nuttyartist/notes/pull/263#issuecomment-647443116?

nuttyartist:
> So, to me, a possible solution is the version on the left. Another option might be to reduce the spacing on the left/right of the editor, such that it matches with the reduced space at the top.

I can't settle on unaligned elements. This is bad design.

> When we use a spacer to move everything down, wouldn't we get the same situation we had initially at #263 (comment)?

True. Do you think adding just a little bit would look better?

bjorn:
> I can't settle on unaligned elements. This is bad design.

What about the suggestion to reduce the spacers left/right of the text editor to make things align again? Like this:

notes1

nuttyartist:
> What about the suggestion to reduce the spacers left/right of the text editor to make things align again? Like this:

Can you create a branch for this so I can test on my machine? And another one without reducing the margins?

bjorn:
> Can you please create a branch for this so I can test on my machine? And another one without reducing the margins?

I've pushed the version on the right in https://github.com/nuttyartist/notes/pull/263#issuecomment-657258113 to this PR, and I've created a branch in my repo called [`consistent-spacing-with-native-frame`](https://github.com/bjorn/notes/tree/consistent-spacing-with-native-frame) that also reduces the space on left and right of the text edit.

bjorn:
I just tested this on Windows and turning off the native frame is not working correctly:

notes1 

It only works after restart.

Also, I noticed the window buttons are always on the left. That is really weird on Windows.

nuttyartist:
> I've pushed the version on the right in #263 (comment) to this PR, and I've created a branch in my repo called consistent-spacing-with-native-frame that also reduces the space on left and right of the text edit.

Okay let's go with 'consistent-spacing-with-native-frame'. One problem, on 'consistent-spacing-with-native-frame', the textEditor didn't revert to the right margins after turning off the native window frame.

> I just tested this on Windows and turning off the native frame is not working correctly:
>
> notes1 
>
> It only works after restart.

Do you mean getting out of the native window doesn't happen instantly on Windows and only after restart? That's really weird. If this problem is consistent on other machines we should show a prompt asking the user if he wants to do it.

> Also, I noticed the window buttons are always on the left. That is really weird on Windows.

When native-window is on, can you hide these buttons like you do on linux/mac?

bjorn:
> Okay let's go with 'consistent-spacing-with-native-frame'. One problem, on 'consistent-spacing-with-native-frame', the textEditor didn't revert to the right margins after turning off the native window frame.

Yeah, I had explained that part in the commit message of dfc8d62d728b86b6f95482de34f228caba81823f. :-)

> Do you mean getting out of the native window doesn't happen instantly on Windows and only after restart? That's really weird. If this problem is consistent on other machines we should show a prompt asking the user if he wants to do it.

What happens it what I showed on the screenshot, where our own buttons become visible but the native frame does not disappear. This bug only happens on Windows. I'll see if I can fix it.

> When native-window is on, can you hide these buttons like you do on linux/mac?

Yes, that already happens.

nuttyartist:
> Yeah, I had explained that part in the commit message of dfc8d62. :-)

I don't understand, why isn't as simple as changing the stylesheet when unchecking the native window frame option?

> What happens is what I showed on the screenshot, where our own buttons become visible but the native frame does not disappear. This bug only happens on Windows. I'll see if I can fix it.

👍 

> Yes, that already happens. Edit: To be clear, I didn't mean the buttons were "always on", I meant to say that our buttons were "on the left", whereas on Windows these buttons are on the right side for all other applications.

Ah, ok. Yes, we should move it to the right. I'm going to open an issue for this.

bjorn:
> I don't understand, why isn't as simple as changing the stylesheet when unchecking the native window frame option?

I never said it wasn't simple, I just pointed out it still needs to be done. It needs some code reorganization because that stylesheet is set in a function that can only be called once. I didn't have time for this yet.

nuttyartist:
Ok 👍

nuttyartist:
Hey @bjorn, any progress?

nuttyartist:
@bjorn I just tested on Windows and can see what you mean about the native window frame not disappearing before doing a system restart. If you think it's an upstream bug or something that's hard to figure out, maybe we can just show a dialog after choosing to show native window frame. Something like: "After you do this action, In order to revert back to custom decoration, you will need to restart your system. Are you sure you want to do this? Yes/No.". What do you think?

bjorn:
> If you think it's an upstream bug or something that's hard to figure out, maybe we can just show a dialog after choosing to show native window frame. Something like: "After you do this action, In order to revert back to custom decoration, you will need to restart your system. Are you sure you want to do this? Yes/No.". What do you think?

Yeah, I still need to look into this but if we don't find a solution then such a dialog would work.

I've been rather busy so I had no time yet to continue this patch.

nuttyartist:
That's alright. I would love to release v1.5 till the end of the month, so if we can wrap this up by then it would be great.

nuttyartist:
@bjorn Can you add this code for the dialog so we will merge this?

```c++
QMessageBox msgBox;
msgBox.setText(tr("Warning: After performing this action, to revert to custom decoration you will need to restart your system."));
msgBox.setInformativeText(tr("Would you like to continue?"));
msgBox.setStandardButtons(QMessageBox::Yes | QMessageBox::No);
msgBox.setDefaultButton(QMessageBox::No);
if  (msgBox.exec() != QMessageBox::Yes) {
    return;
}
```

nuttyartist:
Okay, merging this now. I'll add the dialog code afterward.

bjorn:
Alright, that's fine. Will see if I can still get around to the margin changes...

bjorn:
@nuttyartist That part:

> > I don't understand, why isn't as simple as changing the stylesheet when unchecking the native window frame option?
> 
> I never said it wasn't simple, I just pointed out it still needs to be done. It needs some code reorganization because that stylesheet is set in a function that can only be called once. I didn't have time for this yet.

nuttyartist:
Oh, okay!

PR #270, opened 2020-06-21
Changed some string literals to translatable strings

bjorn:
If translation is enabled at some point, these strings should be translatable.

Also changed Ctrl+J shortcut from being shown explicitly in the text to being rendered by the menu.


nuttyartist:
Cool, merging... :)

Issue #183, opened 2017-10-21
Checked symbol in checkbox is not visible unless the element is hover on Mac


nuttyartist:
On Mac, I see the '✓' icon only when I hover over the menu, but it should be shown also when I don't hover and highlight the menu.  More info about the bug and partial solutions can be found here #181.
![31583183-8a545010-b19e-11e7-852a-d4b576625cb7](https://user-images.githubusercontent.com/16375940/31850261-082e219c-b658-11e7-9a61-279157a08845.gif)

bjorn:
Just noting that this is related to the following stylesheet, according to comments at #181.

https://github.com/nuttyartist/notes/blob/858f617d5cb769fd0564688a7d01a0cb5a066769/src/mainwindow.cpp#L929-L932

nuttyartist:
Yes, these few lines are problematic. Can you please create a PR where you comment out these lines so we let the pop up be drawn natively?

https://github.com/nuttyartist/notes/blob/858f617d5cb769fd0564688a7d01a0cb5a066769/src/mainwindow.cpp#L928-L936

nuttyartist:
Solved #273 by reverting to drawing native pop-up.

PR #273, opened 2020-07-16
Aesthetic changes and better popup for 3 dots button. Fixes #177, #183

### LINKED ISSUE #177: Aesthetic changes
List of aesthetic changes to be made (I'll work on that and send a PR):

- [x] Change buttons colors to reflect better the states they're in (hovered, pressed, released).
- [x] Revert back to old editor background (?)
- [x] Add Qt::WA_MacShowFocusRect, 0 to lineEdit (this disable the not-in-style focus rectangle in lineEdie on MAC).
- [x] Custom textEdit and lineEdit color when selecting text. 
- [x] Don't change font size when clicking a note? EDIT: Decided to keep the way it is - thinning the note text when selected, to make the selection more clear.

nuttyartist:
Solves #177 and #183

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -546,6 +546,8 @@ void MainWindow::setupSearchEdit()
 {
     QLineEdit* searchEdit = m_searchEdit;
 
+    searchEdit->setAttribute(Qt::WA_MacShowFocusRect, 0);
+
     searchEdit->setStyleSheet(QStringLiteral("QLineEdit{ "
                                              "  padding-left: 21px;"
                                              "  padding-right: 19px;"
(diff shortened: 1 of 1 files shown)


src/mainwindow.cpp
```diff
@@ -1011,15 +1014,15 @@ void MainWindow::onDotsButtonClicked()
...
+    /*mainMenu.setStyleSheet(QStringLiteral(
                                "QMenu { "
                                "  background-color: rgb(255, 255, 255); "
                                "  border: 1px solid #C7C7C7; "
                                "  }"
                                "QMenu::item:selected { "
                                "  background: 1px solid #308CC6; "
                                "}")
-                           );
+                           );*/
```
bjorn, line 1025: Yes, I think this is also an improvement on Linux where the popup looked a bit broken with this stylesheet as well:

image

The checkmark is too small, and it looks better without stylesheet:

image

It will depend on the platform style, but I think that's a good thing.

I would only suggest that you remove these lines, instead of commenting them out.

nuttyartist: Will do!

nuttyartist:
Merging...

PR #280, opened 2020-09-06
Merge dev into master

nuttyartist:
Merge dev branch into master after finishing with v1.5.0.


Waqar144:
`DBManager` still needs to be updated to use correctly escaped queries

nuttyartist:
@Waqar144 Haven't we solved this already?

Can you give a problematic example?

Waqar144:
> @Waqar144 Haven't we solved this already?

Yeah, just for one query. This can happen with any of the other queries.

> Can you give a problematic example?

This issue is an example, maybe there are other issues that haven't surfaced yet.

https://github.com/nuttyartist/notes/issues/234

nuttyartist:
Let's leave this for a future version or update. I closed the issue you mentioned because the example you provided there got fixed in this [commit](https://github.com/nuttyartist/notes/commit/bfdca20a8d230f1a7352960881d455cca52c57fa). If you have other examples, please comment there so we could know what's not working.

Waqar144:
> Let's leave this for a future version or update. I closed the issue you mentioned because the example you provided there got fixed in this [commit](https://github.com/nuttyartist/notes/commit/bfdca20a8d230f1a7352960881d455cca52c57fa). If you have other examples, please comment there so we could know what's not working (and I'll open it).

Alright, but keep this in mind that this leaves space for bugs openly when we know this *can and should* be fixed. You never know what the user writes in their notes and an issue like this could potentially lead to data loss very easily :)

nuttyartist:
I agree! Do you have any examples of escaped queries that Notes can't handle properly?

bjorn:
I'm not sure if I understand the workflow here. What is this merge for? What is the "master" branch for?

nuttyartist:
master is the branch for stable releases.

Also, I think we need to merge to master the updated `UPDATES.json` file so other people can get automatic updates.

Waqar144:
> I agree! Do you have any examples of escaped queries that Notes can't handle properly?

Not currently because I haven't dug into it. Maybe any note with `%` chars etc can cause bugs

nuttyartist:
> Not currently because I haven't dug into it. Maybe any note with % chars etc can cause bugs

I can't seem to make a query that would not be saved properly. I think commit  https://github.com/nuttyartist/notes/commit/bfdca20a8d230f1a7352960881d455cca52c57fa solved that. But if anybody thinks they found something, they can report that on the issue and we'll take care of that on our next release or update.

bjorn:
> I can't seem to make a query that would not be saved properly. I think commit [bfdca20](https://github.com/nuttyartist/notes/commit/bfdca20a8d230f1a7352960881d455cca52c57fa) solved that. But if anybody thinks they found something, they can report that on the issue and we'll take care of that on our next release or update.

So, looking at the code, errors will likely occur when:

* Migrating a note from the old format to the new database format that contains "%1" (either active note or trashed note)
* Deleting a note after putting "%1" in it.

All other query parameters are numbers, which do not have this problem.

Issue #292, opened 2021-08-11
Help needed implementing Themes (ie, Dark mode) on notesList model/view


nuttyartist:
Having introduced themes in #295, I'm having a problem updating the color of the notes list (noteView) when the user chooses a different theme (ie, Dark). I guess I'm not sure what's the proper way of updating the delegate/model.

I didn't touch C++ for a while so I'm a bit rusty.

Can someone help with this?
@akud82, @bjorn , @theshadowx, @Waqar144

akud82:
It's important to emit the dataChanged() signal. 
Delegate doesn't know that you change color and you need to emit event to force repaint.

nuttyartist:
I tried this (in mainwindow.cpp): 

```emit m_noteModel->dataChanged(m_proxyModel->index(0,0), m_proxyModel->index(m_proxyModel->rowCount()-1, 0));```

But it doesn't work. I gotta say I have not much clue about the Model/View methodology.

nuttyartist:
Done in #302.

PR #295, opened 2021-09-13
Add Editor Settings window, some changes to editor, and more

nuttyartist:
This PR adds a pop-up Editor Settings window that allows:

- Choosing different fonts (from predefined ones) of 3 types: serif, sans-serif, and mono.
- Increasing/decreasing editor font size.
- Setting the width of text by no. of characters (increasing/decreasing) or choosing Full Width for the text to accommodate the entire width of the editor (excluding padding).
- 3 themes: Light, Dark, and Sepia.

And some changes to the editor:
- Automatic scaling of font sizes based on text editor size (when a user increase/decrease font size he increases/decreases the base font size from which the large-window-font-size and the small-window-font-size are determined)
- Hide frameRight when writing, recover when hovering.
- Padding the editor so the text is in the middle of it.  

This PR also:
- Fix previous MacOS issues with scroll bars by using native ones (only on MacOS, as the native scrollbar on Windows still look ugly, for some reason, and on most Linux distros I tested, the resulted scrollbar wasn't looking good as well).
- Use Qt-Nice-Frameless-Window repo to solve our window issues on Mac and Windows, and a more native look (some problems are still present on some MacOS versions, a new issue will be opened).
- Add About box with credits for devs and fonts.

Some problems remain (I will open new issues for them):

1. Need help changing the theme on notesList background color and text color. See #292 
2. Needs new traffic lights that look better on dark mode on Linux.
3. As mentioned, Custom Frameless Window (Qt-Nice-Frameless-Window) doesn't show well on some versions of Mac.


nuttyartist:
Merging, so devs can develop v2.0.0 with the latest changes. If we encounter problems, we'll solve them on the way.

Issue #299, opened 2022-04-02
Debian missing qt/sqlite driver, notes dont save or load


someodd:
First off, I love this software. Absolutely perfect for spur-of-the-moment notes. Great job!

I've noticed this problem on/off where I'll do a fresh install of Debian, install the `deb` and it won't save any notes/won't load the `notes.db` I transferred/migrated from an older system. I had to do another transfer just yesterday and I decided to launch Notes from the terminal and noticed it complaining about not having loaded/access to any qt/sqlite drivers and that it wasn't connecting to the database. Part of the error was `QSqlDatabase: QSQLITE driver not loaded`. After a little bit of research I found out I could install the driver it was complaining about with ` sudo apt install libqt5sql5-sqlite`. Now it works fine, it connects to `notes.db`, with all of my notes migrated.

I am using Debian testing (bookworm/sid). I installed using `sudo apt install ./notes_1.5.0_amd64-bionic.deb` (as well as having tried `dpkg`).

nuttyartist:
Hi! Thanks for your kind words. I'm glad you solved your issue. What OS version you were using?

Thanks for reporting the issue.

theshadowx:
might be missing in the dependencies list : 
https://github.com/nuttyartist/notes/blob/dev/packaging/linux/debian/control#L13
`Depends: libqt5widgets5 (>= 5.2), libqt5gui5 (>= 5.2), libqt5network5 (>= 5.2), libqt5sql5 (>= 5.2), libqt5core5a (>= 5.2), sqlite3, ${misc:Depends}, ${shlibs:Depends}`

someodd:
@nuttyartist I am using Debian unstable (sid/bookworm).

Thank you @theshadowx for pointing that out. I went ahead and added the dependency to the place you mentioned (I created a PR: #300). `libqt5sql5-sqlite` isn't actually a dependency of [libqt5sql5](https://packages.debian.org/bullseye/libqt5sql5), just recommended by it.

nuttyartist:
@theshadowx Thanks for helping out.

PR #300, opened 2022-04-05
Add deb depend to fix db access

someodd:
Add a dependency to the deb package, which without, results in Notes being unable to communicate to the SQLite database.

Fixes issue #299.

### DIFF
--- a/packaging/linux/debian/control
+++ b/packaging/linux/debian/control
@@ -10,7 +10,7 @@ Homepage: http://www.get-notes.com/
 
 Package: notes
 Architecture: amd64
-Depends: libqt5widgets5 (>= 5.2), libqt5gui5 (>= 5.2), libqt5network5 (>= 5.2), libqt5sql5 (>= 5.2), libqt5core5a (>= 5.2), sqlite3, libqt5sql5-sqlite (>= 5.2), ${misc:Depends}, ${shlibs:Depends}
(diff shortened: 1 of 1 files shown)


nuttyartist:
Is this still necessary? @bjorn @guihkx

guihkx:
I'm not familiar with deb packaging, but this change makes sense to me.

Also, do you officially provide any .deb packages? I couldn't find them...

packaging/linux/debian/control
```diff
@@ -10,7 +10,7 @@ Homepage: http://www.get-notes.com/
 
 Package: notes
 Architecture: amd64
-Depends: libqt5widgets5 (>= 5.2), libqt5gui5 (>= 5.2), libqt5network5 (>= 5.2), libqt5sql5 (>= 5.2), libqt5core5a (>= 5.2), sqlite3, ${misc:Depends}, ${shlibs:Depends}
+Depends: libqt5widgets5 (>= 5.2), libqt5gui5 (>= 5.2), libqt5network5 (>= 5.2), libqt5sql5 (>= 5.2), libqt5core5a (>= 5.2), sqlite3, libqt5sql5-sqlite (>= 5.2), ${misc:Depends}, ${shlibs:Depends}
```
guihkx, line 13: Nitpick: This is not a blocker, but after PR #390, Notes will not compile on any Qt version older than 5.9.5.

nuttyartist:
@someodd Can you reflect the changes to a version higher than 5.9.5?

zjeffer:
Updated to 5.9 and tested the make_deb.sh script inside an ubuntu devcontainer, merging.

PR #345, opened 2022-09-17
Update qautostart submodule

guihkx:
Introduces a fix for Windows and adds some general improvements for Linux.


nuttyartist:
Hi! I'm getting this compilation error (Linux):

Screenshot from 2022-09-17 18 31 17

guihkx:
Oh, right. `Qt::endl` was only introduced in 5.14.

I've sent a [pull request](https://github.com/b00f/qautostart/pull/4) to ensure it compiles with older Qt versions. Once it gets accepted, I'll update the submodule again.

guihkx:
Should work now, please retest when possible.

If this gets accept, I can drop the patch I'm using to create a Flatpak package of Notes [here](https://github.com/guihkx/flatpak-app-notes). It's pretty much ready and working as-is, but being able to drop the [huge patch](https://github.com/guihkx/flatpak-app-notes/blob/master/flatpak-autostart.patch) I'm using for the next release of Notes would be nice!

Unfortunately, even though the Flatpak manifest is pretty much ready to be [submitted for review](https://github.com/flathub/flathub/blob/master/CONTRIBUTING.md) on Flathub, I don't have the time to be a proper maintainer. :(

So if you @nuttyartist (or anyone else) gets interested, please, feel free to fork my repository at any time and submit it for review on Flathub.

Thanks.

nuttyartist:
Compiles well now. I'll merge this. I'll consider being a maintainer. Are you going to upload the flatpack package after this PR?

guihkx:
I'd rather wait for the next release of Notes so that I can drop that patch I'm using. There's also some TODO in the README that I'd like to address before sending the app for review on Flathub... But yeah, I can send it for review but I won't be able to maintain it. Shouldn't be hard for you to do it though, the Flatpak manifest is pretty simple.

nuttyartist:
Okay. Which TODO are you talking about?

guihkx:
Here: https://github.com/guihkx/flatpak-app-notes#todo

bjorn:
Regarding the app ID, since Notes has its own website at https://www.get-notes.com/, I think `com.get-notes.Notes` might be better.

guihkx:
Initially I tried using that, but unfortunately `flatpak-builder` doesn't allow it :(

```
error: 'com.get-notes.Notes' is not a valid application name: Only last name segment can contain -
```

bjorn:
Ok, I wasn't aware of that restriction. :/

guihkx:
Yeah I wasn't either, and until just now I wasn't aware there was [documentation](https://docs.flatpak.org/en/latest/conventions.html#application-ids) regarding this:

>`org.example-site.Foo`
>This ID is not valid according to the DBus specification. You can use `org.example_site.Foo` instead.

So we could use `com.get_notes.Notes` for the app id...

Thoughts?

nuttyartist:
> Yeah I wasn't aware either, and until just now I didn't know there was [documentation](https://docs.flatpak.org/en/latest/conventions.html#application-ids) regarding this:
> 
> > `org.example-site.Foo`
> > This ID is not valid according to the DBus specification. You can use `org.example_site.Foo` instead.
> 
> So we could use `com.get_notes.Notes` for the app id...
> 
> Thoughts?

I'll go with `com.github.nuttyartist.notes` so other apps of mine can go under one umbrella.

guihkx:
Okay cool! One last question, are you okay with `com.github.nuttyartist.notes` or would you prefer `notes` to be capitalized? Like `com.github.nuttyartist.Notes`?

nuttyartist:
What's the standard? I think lowercase should be better. Thanks!

PR #350, opened 2022-09-17
Update Ubuntu versions used in GitHub Actions

mbrukman:
* Ubuntu 16.04 was removed on Sep 20, 2021, as per https://github.com/actions/runner-images/issues/3287
* Currently-available versions are 18.04 (deprecated), 20.04, and 22.04 as per https://github.com/actions/runner-images#available-images

Added a comment pointing to the list of currently-available Ubuntu versions in a comment in the workflow config to make it easire to find the full list and update it in the future.


nuttyartist:
👍 Merging...

PR #355, opened 2022-09-18
Add keyboard shortcuts for text formatting: bold, italics, strikethrough (take 2)

AdrianTM:
Started a new pull request, I think this is from a cleaner state.

I think I addressed your request. It might need some polish in some edge cases but I think it should work fine as is.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -459,9 +459,9 @@ void MainWindow::setupKeyboardShortcuts()
     new QShortcut(QKeySequence(Qt::CTRL + Qt::SHIFT + Qt::Key_S), this, SLOT(onStyleEditorButtonClicked()));
     new QShortcut(QKeySequence(Qt::CTRL + Qt::SHIFT + Qt::Key_J), this, SLOT(toggleNodeTree()));
     new QShortcut(QKeySequence(Qt::CTRL + Qt::Key_A), this, SLOT(selectAllNotesInList()));
-    new QShortcut(QKeySequence(Qt::CTRL + Qt::Key_B), this, [this](){applyFormat("**");});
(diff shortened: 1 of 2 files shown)


nuttyartist:
Great job! It works great now. 2 Edge cases I noticed:

1. If the cursor is between two words, but on an empty space, it would operate the word before it where it should just add the operation.
short 1
2. When the cursor is on the left side of a word "|Here" the `revert` operation works well, but not when the cursor is on the right side like "Here|"
short 2

Can you fix these two and then I'll merge this?

AdrianTM:
I fixed the second inconsistency (at least partially)

I cannot fix in an easy way the first inconsistency, it's a Qt issue when you use select(QTextCursor::WordUnderCursor); Qt returns the word before the blankspace, even if you don't have the cursor on it. I would have to implement my own WordUnderCursor to go around that issue.

This is also related to another inconsistency: when you have something like **word** it returns "word" if you put the cursor between the ** and w and it returns ** if you put the cursor between d and ** There's yet another inconsistency, it matters if you are with the cursor at the end of the word and there's a space after it, or there's nothing about it.

It's a bit maddening, but I think right now it should work fine for most of the use cases. Please give a try in case I introduced any new bug, and decide if you can live with the inconsistencies, this as much as I do for now.

nuttyartist:
Okay, I will file an issue about those edge cases. 

> This is also related to another inconsistency: when you have something like word it returns "word" if you put the cursor between the ** and w and it returns ** if you put the cursor between d and ** There's yet another inconsistency, it matters if you are with the cursor at the end of the word and there's a space after it, or there's nothing about it.

Can you please show a video about what you mean? I couldn't understand.

Many thanks for your contribution! I'm looking forward to more from you!

PR #364, opened 2022-09-20
Fix GitHub Actions links to point to this project

mbrukman:
Earlier change incorrectly used GitHub Docs sample URLs and updated the badge URL, but not the target URL to point to this project's actions page.

Since this is not changing code, we can [skip ci].


bjorn:
Thanks! :-)

nuttyartist:
Thanks (:

And @bjorn thank you as well.

PR #365, opened 2022-09-20
Disable fast-fail on Ubuntu in GitHub Actions

mbrukman:
The current failure may be due to Ubuntu 22.04 missing packages that we expect to see on 18.04 and 20.04, so it may need a different set of packages, but the current approach is cancelling all builds as soon as the 22.04 build fails, so we're not learning everything we need to know.

### DIFF
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -7,10 +7,11 @@ jobs:
 
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
         # https://github.com/actions/runner-images#available-images
         os: [ubuntu-18.04, ubuntu-20.04, ubuntu-22.04]
-    
+
     steps:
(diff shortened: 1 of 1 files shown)


mbrukman:
@nuttyartist — looks like the build is OK on Ubuntu 18.04 and 20.04, but fails on 22.04, so it should be disabled until that can be fixed. Let me know if you want me to remove it as part of this PR (either in the same commit, or in a separate commit), or as a separate PR entirely.

Also, the macOS build is failing, but that's a separate issue.

bjorn (approved):
Looks like a good idea to me.

nuttyartist:
@bjorn Should we merge this and add the changes you mentioned on a different PR?

bjorn:
Yes, I think the fixing of the build is better done in another PR.

nuttyartist:
Okay, merging...

PR #366, opened 2022-09-20
fix: specify the easing curve correctly

guihkx:
This fixes a derp I introduced with #340.

Thanks @bjorn for spotting this!

### DIFF
--- a/src/notelistdelegate.cpp
+++ b/src/notelistdelegate.cpp
@@ -55,7 +55,7 @@ NoteListDelegate::NoteListDelegate(NoteListView *view, TagPool *tagPool, QObject
     m_timeLine = new QTimeLine(300, this);
     m_timeLine->setFrameRange(0,m_maxFrame);
     m_timeLine->setUpdateInterval(10);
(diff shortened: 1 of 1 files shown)


bjorn (approved):
That looks better. :)

bjorn:
Hmm, not sure why GitHub added `Co-authored-by: guihkx <guihkx@users.noreply.github.com>`...

guihkx:
IIRC, that line is added automatically when you squash a PR using Github.

bjorn:
> IIRC, that line is added automatically when you squash a PR using Github.

Yeah, but usually only when there is a mismatch between the email used on the commit and the email set on the GitHub account. Actually you seem to be `guih.rox@gmail.com` to GitHub, but you seem to be committing with `guihkx@users.noreply.github.com`, so that's probably it.

nuttyartist:
> This fixes a derp I introduced with #340.
> 
> Thanks @bjorn for spotting this!

@bjorn Good catch. Thank you both!

PR #367, opened 2022-09-22
Initial GitHub Actions integration

guihkx:
You can grab the generated artifacts from here: https://github.com/guihkx/notes/actions/runs/3102208505

So far I have successfully tested them on the following platforms:

- [x] Windows 7 64-bit
- [x] Windows 10 64-bit
- [x] Arch Linux 64-bit
- [x] macOS (don't have a Mac)

I chose some opinionated defaults while creating this workflow, but we can definitely tweak it further, according to your feedback! These are "my defaults":

- Build using the oldest version possible of the OS.
- Build using the newest Qt 5 version available.
- Build only a single binary for each OS.
- Build 32-bit binaries for Windows. I'm not particularly a fan of this one, and I just did it because the latest official release includes a 32-bit build as well. I mean, I don't think the percentage of people using a 32-bit OS is high, so I'm inclined to only build 64-bit binaries for all OSes...

Let me know your thoughts.


bjorn:
Really well written workflow! I would just suggest to keep the workflows testing macOS Homebrew compile as well as a job for compiling on latest Ubuntu against the system packages. Even if they don't generate packages, it is important to test that these kind of builds keep working.

guihkx:
Hey guys, thanks for the feedback. I've been working on this on a different branch, just so it doesn't pollute your guys' email with every push I make... 😅 

I have implemented (most) of your suggestions, but I still like to get further feedback if possible.

As of now, I have moved into three distinct workflows, instead of a single one:

- [`linux.yml`](https://github.com/guihkx/notes/actions/runs/3117196083/workflow)
- [`macos.yml`](https://github.com/guihkx/notes/actions/runs/3117196086/workflow)
- [`windows.yml`](https://github.com/guihkx/notes/actions/runs/3117196085/workflow)

The artifacts are available here:

- Linux: https://github.com/guihkx/notes/actions/runs/3117196083
- macOS: https://github.com/guihkx/notes/actions/runs/3117196086
- Windows: https://github.com/guihkx/notes/actions/runs/3117196085

It'd be great if you guys tested them (especially the macOS one, because I have no access to a Mac).

The app signing part will have to be implemented in a future PR, though. I just want to make sure I at least got the `CFBundleName` and `NSRequiresAquaSystemAppearance` stuff right and, most importantly, that the app can actually run this time... 😄 

Anyway, these are the pros of the current workflow approach, in my opinion:

- More flexibility: If you have a problematic workflow, GitHub allows you to disable it individually, while still keeping the others running.
- More organized (yet a bit more complex and "duplicated");
- Less "walking on egg shells": With the single workflow I initially introduced, I went through much pain just to make sure we'd have build commands that would work in all three OSes. For example, every time I thought of using `sed`, I had to ensure I was not using a switch that the macOS version didn't support. But even when I thought I'd found the "perfect" command for all OSes, on Windows it would not work for whatever reason. I then realized this was very silly and it'd make much more sense to have three distinct overflows with commands specific to that OS. Otherwise, an innocent contributor could simply change a single command (or parameter) and accidentally break the whole workflow for no good reason.

But even then, I'm not sure I'm *100%* happy with the current workflows either:

image

Isn't 14 jobs for each commit too excessive? 

Speaking of which, should we provide `debug` builds for all OSes? Or should we provide `release` builds only? Currently, I'm doing both. And I think this is an important question, because storage and bandwidth on GitHub are not free, which can be quite problematic considering that both `debug` and `release` builds of Windows (plus the installer) sum up ~400 MB:

image

Maybe something we could do is to only provide downloadable debug artifacts when someone commits to a specific branch... Or something like that. We could also only build the unified Windows installer when a new tag was pushed... We could also compress the artifacts better, but the downside is having to [extract them twice](https://github.com/actions/upload-artifact/issues/39)..

Anyway, there are just some random thoughts. Let me know what you guys think overall.

Thanks!

nuttyartist:
What a great job! The macOS artifact works well with the new information added.plist successfully.
14 jobs sound excessive, indeed. 

- Regarding debugging artifacts, If there's a way that only when pushing into some "debugging" branch then it will make sense.
- Under the Windows build https://github.com/guihkx/notes/actions/runs/3117196085, why do the released versions weigh 53MB and 63MB?  Are they both unified? The unified installer that I uploaded weighs 56MB.
- Also what is the "NotesSetup_2.0.0-g0a40085-Windows-release" that weight 33MB?
- So we have `Linux`, `Windows`, and `macOS` workflows. What is the `Build + Artifacts` workflow? Is it just uploading the artifacts the previous workflows created and put them in one place?

Thank you for your great work.

guihkx:
> Under the Windows build https://github.com/guihkx/notes/actions/runs/3117196085, why do the released versions weigh 53MB and 63MB? Are they both unified? The unified installer that I uploaded weighs 56MB.

Right, let me explain what each one is:

- `Notes_2.0.0-*-x64-Windows-Qt5.15.2-release`: This one is a portable `release` build (x64-only), which means you can basically download it, unzip it anywhere and just run `Notes.exe`, without having to actually run an installer and maybe mess up the current Notes installation you have. I thought this could be useful to quickly test a build, but it can be definitely removed if you wish.

- `Notes_2.0.0-*-x64-Windows-Qt5.15.2-debug`: Same as above, but it also has debug symbols.

- ` Notes_2.0.0-*-x86-Windows-Qt5.15.2-release`: Same as the first one, except this is a x86-only `release` build. 

- ` Notes_2.0.0-*-x86-Windows-Qt5.15.2-debug`: Same as above, but with debug symbols included.

And now onto your follow up question:

>Also what is the "NotesSetup_2.0.0-g0a40085-Windows-release" that weight 33MB?

This one is the installer that bundles both `release` builds (x86 and x64) into one, but only if they both succeed:

image

The installer is built using Inno Setup 6 using a [slightly modified version](https://github.com/guihkx/notes/blob/0a4008571a07fc9c7203f8581e0afee11f7c2427/packaging/windows/Notes_Inno_Script_Github.iss) of the `.iss` script you posted here. I mainly added new options just to make it look prettier on the Add/Remove Programs screen. Some comparison screenshots:

Before (official 2.0.0 installer):

image

After:

Screenshot from 2022-09-24 20-51-22

However, I pushed this `.iss` script as `Notes_Inno_Script_Github.iss`, just to avoid breaking the AppVeyor CI that still relies on `Notes_Inno_Script.iss`.

I also included the license file, and the installer displays it:

image

But my installer currently installs to `C:\Program Files\Notes` and yours to `C:\Program Files\Notes 2.0.0`, but I wonder if you really intended that, and if so, what will you change the install path again with newer 2.x.x releases?

> So we have `Linux`, `Windows`, and `macOS` workflows. What is the `Build + Artifacts` workflow? Is it just uploading the artifacts the previous workflows created and put them in one place?

Right, so the `Linux`, `Windows` and `macOS` workflows come from this new branch I'm working on, called `ci-test`. The `Build + Artifacts` workflow comes from the `ci` branch, which is being used in this pull request. I will move the workflows from the `ci-test` branch to `ci` as soon as you're okay with the workflows. I just started testing on a different branch, because like I said, I was `push`ing too much into that branch and it would absolutely annoy you guys and pollute this PR.

guihkx:
I just realized I missed this question:

>The unified installer that I uploaded weighs 56MB.

That's because your installer is bundling both `vcredist_x64.exe` and `vcredist_x86.exe` (and not really installing them), while mine just directly copies only the required DLLs from `$env:VCToolsRedistDir` ([source](https://github.com/guihkx/notes/actions/runs/3117196085/workflow#L66)).

nuttyartist:
Thank you for all the answers! I think keeping the portable versions on Windows is ok. Also, I like the changes you did with .iss to make it look right. And C:\Program Files\Notes seems to be the right path.

Is adding the license needed? I thought the license is only for those who want to use the open-source code, not as a EULA for end users. And since we don't track anything, we don't really need a privacy policy/EULA for Notes?

Great work! I'll merge after discussing the license.

guihkx:
Okay, I just added the changes from the other branch I was working on. Please take a look at the new commits as well.

>Is adding the license needed? I thought the license is only for those who want to use the open-source code, not as a EULA for end users. And since we don't track anything, we don't really need a privacy policy/EULA for Notes?

I think you're actually right. I was going to use Firefox as an example of a popular open-source program that displayed their license on their installer, but I just installed Firefox and they have don't that anymore. lol

I removed it as well.

I also disabled `debug` builds on artifacts by default, and with that I was able to lower the total number of jobs down to 7. It's easy to add them when needed, though. I've added comments on each workflow explaining how to do that.

Here's something important: I also disabled the Windows installer by default, but you can force it to run just by adding  `[ci build installer]` on the body of your most recent. I'm not sure this is the ideal solution, but it can be tweaked/improved in future PRs.

Finally, I think this should be good enough to merge as an initial integration for GitHub Actions, but let me know if you have any other requests.

nuttyartist:
Great, but I don't see a problem creating a unified installer for every commit, especially since it uses the already created artifacts, so it doesn't take much time.

guihkx:
Done!

nuttyartist:
Awesome! Merging this. Thank you so much. Awesome work @guihkx.

guihkx:
Don't mention it :)

By the way, you can now delete [this unused workflow run](https://github.com/nuttyartist/notes/actions/workflows/ci.yml), just so it doesn't confuse people.

nuttyartist:
@guihkx You mean clicking on "Disable workflow"?

guihkx:
Not exactly. I mean deleting that single workflow run, i.e.:

image

nuttyartist:
Done, thanks (:

> Not exactly. I mean deleting that single workflow run from "Build + Artifacts", i.e.:
> 
> image

Done, thanks (:

PR #371, opened 2022-09-24
Fix Ubuntu 22.04 package names

mbrukman:
The Qt 5 package names are different in Ubuntu 22.04 from prior releases (18.04 and 20.04), so we need to have a different package installation command in CI.

This is a follow-up to PR https://github.com/nuttyartist/notes/pull/365 which disabled fast-fail so we can see all Ubuntu failures independently per-version.

PR #372, opened 2022-09-24
Separate out each macOS version on GitHub Actions

mbrukman:
Instead of running on `macos-latest` which is a moving target, specify all currently-available versions, and disable fast-fail, so that we can see which versions are passing vs. failing independently, and can adjust the installation commands for each one separately.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -1,20 +1,23 @@
-name: Default Qt on macos
+name: macOS
 
 on: [push, pull_request]
 
 jobs:
   build:
-
     runs-on: ${{ matrix.os }}
     strategy:
+      fast-fail: false
       matrix:
-        os: [macos-latest]
+        # https://github.com/actions/runner-images#available-images
+        os: ['macos-10.15', 'macos-11', 'macos-12']
     
     steps:
-    - uses: actions/checkout@v1
+    - name: Checkout repo
+      uses: actions/checkout@v1
       with:
         submodules: recursive
-    - name: install dependencies
+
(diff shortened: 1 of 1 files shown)


mbrukman:
@bjorn, @nuttyartist — I don't see the macOS jobs running at all for this PR; maybe you have to enable it to run for me? I split the macOS build into 3 separate ones, one per version, so it probably requires your approval (I see nothing in the UI, since I'm not an administrator).

The Ubuntu 22.04 is failing here because the fix is in PR https://github.com/nuttyartist/notes/pull/371.

nuttyartist:
We should probably wait after merging @guihkx's PR #367?

bjorn:
Actually I don't think we need to check the build on many versions of macOS. This is not the job that builds the packages, so I think just using `macos-latest` is fine. Same for the Ubuntu workflow. These workflows just make sure that if somebody clones the repository and tries to compile, they'll likely succeed, and these people are likely on up-to-date systems.

nuttyartist:
Is there a reason to suspect the build will fail on older versions even if it's successful on the latest version of macOS? If so, maybe it's good to check those?

bjorn:
> Is there a reason to suspect the build will fail on older versions even if it's successful on the latest version of macOS? If so, maybe it's good to check those?

There is, but I don't think we need to waste so many resources checking for those rare cases, because developers will generally be on the latest version anyway.

nuttyartist:
Fair enough, if we ever decide differently we can always change this.

Thanks, @mbrukman.

PR #375, opened 2022-09-25
Add shortcuts for different heading levels

AdrianTM:
Set up different heading levels with Ctrl+1..6, Ctrl+0 resets heading Ctrl+- and Ctrl+= decrease and increase heading level


nuttyartist:
Great idea! 2 things to note:

1. When keeping on using the shortcut after using it more than once, the text goes one line down. It should change on the same line.
short
2. The shortcuts for `Ctrl +/-` are too general for this action, I would say they make more sense for increasing/decreasing font size. So maybe you change this to `Ctrl-Shift-+/-` and `Ctrl-Shift-0`

Thanks!

AdrianTM:
Thanks, I changed Ctrl-Shift-+/- for some reason Ctrl-Shift-0 doesn't work on my machine it might be conflicting with a general shortcut, it also makes sense to keep that consistent with Ctrl-1, 2, 3...

I fixed the problem you showed in the video. 

That force-push is because I pulled a commit by mistake from master, so I removed it and force-pushed in my branch, it should apply cleanly on your dev branch.

nuttyartist:
Works well now. Thank you! Merging...

PR #376, opened 2022-09-25
Add F10 for dotmenu 

AdrianTM:
Add F10 shortcut to open dot menu; add keyboard accelerators for menu (you can navigate by pressing F10 and then the underlined letter corresponding to the menu entry -- basic functionality in Qt apps)

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -438,6 +438,7 @@ void MainWindow::setupTrayIcon()
  */
 void MainWindow::setupKeyboardShortcuts()
 {
+    new QShortcut(QKeySequence(Qt::Key_F10), this, SLOT(onDotsButtonClicked()));
     new QShortcut(QKeySequence(Qt::CTRL + Qt::Key_N), this, SLOT(onNewNoteButtonClicked()));
     new QShortcut(QKeySequence(Qt::CTRL + Qt::Key_D), this, SLOT(deleteSelectedNote()));
     new QShortcut(QKeySequence(Qt::CTRL + Qt::Key_F), m_searchEdit, SLOT(setFocus()));
@@ -1419,8 +1420,8 @@ void MainWindow::onDotsButtonClicked()
     m_dotsButton->setIcon(QIcon(QStringLiteral(":/images/3dots_Regular.png")));
 
     QMenu mainMenu;
-    QMenu* viewMenu = mainMenu.addMenu(tr("View"));
-    QMenu* importExportNotesMenu = mainMenu.addMenu(tr("Import/Export Notes"));
+    QMenu* viewMenu = mainMenu.addMenu(tr("&View"));
+    QMenu* importExportNotesMenu = mainMenu.addMenu(tr("&Import/Export Notes"));
     importExportNotesMenu->setToolTipsVisible(true);
     viewMenu->setToolTipsVisible(true);
     mainMenu.setToolTipsVisible(true);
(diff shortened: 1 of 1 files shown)


nuttyartist:
I like this! The only thing missing for me is the option to toggle the menu - once it is opened, I should be able to close it with the same shortcut key.

Also, I can't see the underline on my macOS

Screen Shot 2022-09-26 at 1 51 56 PM

AdrianTM:
Pressing F10 closes the menu.

I found this regarding keyboard shortcuts on MacOS: 
>  On Mac, shortcuts are disabled by default. Call qt_set_sequence_auto_mnemonic() to enable them. However, because mnemonic shortcuts do not fit in with Aqua’s guidelines, Qt will not show the shortcut character underlined.

https://doc.qt.io/qtforpython-5/PySide2/QtWidgets/QShortcut.html

If you want can you please try adding that qt_set_sequence_auto_mnemonic() -- I don't have a Mac to test on. But even if the shortcuts would work they would still not be shown underlined.

src/mainwindow.cpp
```diff
@@ -1480,8 +1485,8 @@ void MainWindow::onDotsButtonClicked()
     }
 
     // Enable or Disable markdown
-    QString markDownLabel = m_noteEditorLogic->markdownEnabled() ? tr("Disable Markdown")
-                                                                 : tr("Enable Markdown");
+    QString markDownLabel = m_noteEditorLogic->markdownEnabled() ? tr("&Disable Markdown")
+                                                                 : tr("&Enable Markdown");
```
bjorn, line 1489: I think it might be better to put the mnemonic on M of Markdown. In general it should better to not need a separate key for enable/disable or show/hide.

src/mainwindow.cpp
```diff
@@ -1497,24 +1502,24 @@ void MainWindow::onDotsButtonClicked()
...
 
-    QAction* changeDBPathAction = mainMenu.addAction(tr("Change database path"));
+    QAction* changeDBPathAction = mainMenu.addAction(tr("&Change database path"));
     connect (changeDBPathAction, &QAction::triggered, this, [=]() {
         auto btn = QMessageBox::question(this, "Are you sure you want to change the database path?",
                                          "Are you sure you want to change the database path?"
                                          );
         if (btn == QMessageBox::Yes) {
-            auto newDbPath = QFileDialog::getSaveFileName(this, "New Database path", "notes.db");
+            auto newDbPath = QFileDialog::getSaveFileName(this, "&New Database path", "notes.db");
```
bjorn, line 1522: This is a dialog title, it should not have a mnemonic.

src/mainwindow.cpp
```diff
@@ -1568,8 +1573,8 @@ void MainWindow::onDotsButtonClicked()
 
 #ifndef __APPLE__
     // Use native frame action
-    QAction* useNativeFrameAction = viewMenu->addAction(tr("Use native window frame"));
-    useNativeFrameAction->setToolTip(tr("Use the window frame provided by the window manager"));
+    QAction* useNativeFrameAction = viewMenu->addAction(tr("&Use native window frame"));
+    useNativeFrameAction->setToolTip(tr("Use the &window frame provided by the window manager"));
```
bjorn, line 1577: I don't think it makes sense to add a mnemonic to a tool tip.

nuttyartist:
Okay, we'll leave it as it is on a Mac for now. You're only left to address @bjorn comments before I merge this.

AdrianTM:
Thanks, great comments @bjorn I fixed those and also changed a couple of other Show/Hide menu items to use only one shortcut/mnemonic.

nuttyartist:
Looks good, merging... Thanks a lot!

PR #378, opened 2022-09-27
[feature] add FreeBSD support

mizhka:
This commit adds support of FreeBSD platform.

The change is simple, it adds usage of Q_OS_FREEBSD where Q_OS_LINUX is used.


bjorn (approved):
I wonder if we should use `Q_OS_UNIX` more instead, or in most cases we can probably rely on only special-casing Windows and macOS instead. But otherwise this change looks fine.

nuttyartist:
Thanks! Merging...

PR #379, opened 2022-09-27
[cosmetic] adds window icon

mizhka:
This commit adds window icon for application. Qt adds icon into top-left corner and in taskbar.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -401,6 +401,7 @@ void MainWindow::setupMainWindow()
     ui->searchEdit->setFocus();
     QTimer::singleShot(16, ui->searchEdit, &QTextEdit::clearFocus);
 #endif
+    this->setWindowIcon(QIcon(QStringLiteral(":images/notes_icon.ico")));
 }
 
 /*!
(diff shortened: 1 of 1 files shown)


src/mainwindow.cpp
```diff
@@ -401,6 +401,7 @@ void MainWindow::setupMainWindow()
     ui->searchEdit->setFocus();
     QTimer::singleShot(16, ui->searchEdit, &QTextEdit::clearFocus);
 #endif
+    this->setWindowIcon(QIcon(QStringLiteral(":images/notes_icon.ico")));
```
bjorn, line 404:
```suggestion
    setWindowIcon(QIcon(QStringLiteral(":images/notes_icon.ico")));
```
I realize we have a lot of needless `this->` in the file already, but at least I'd prefer to remove them instead of add more. :)

bjorn (approved):
I didn't test it, but seems like a good idea indeed.

nuttyartist:
Is this change mainly affect Linux?

mizhka:
At least Linux/BSD.

nuttyartist:
Okay, merging...

PR #381, opened 2022-10-01
Update GitHub Actions badges in README

mbrukman:
PR https://github.com/nuttyartist/notes/pull/367 replaced `ubuntu.yml` with `linux.yml` and created a new `windows.yml` so now we need to update the badges in the `README.md` appropriately to refer to the new workflows.

Since we're not modifying code, we can [skip ci] for this change.


nuttyartist:
Good, thanks!

PR #382, opened 2022-10-01
fix #346

### LINKED ISSUE #346: QMessageBoxes cause the application to segfault
I'm using bspwm on Arch Linux. Whenever any QMessageBox is created, it will segfault. The first time it crashed, was in the [Updater::setUpdateAvailable](https://github.com/alex-spataru/QSimpleUpdater/blob/62567eee6631b64e09f365b7518b8744773d9b95/src/Updater.cpp#L428) method frmo the QSimpleUpdater repo. I altered the code so it wouldn't create a messagebox, and then it crashed [here](https://github.com/nuttyartist/notes/blob/dev/src/listviewlogic.cpp#L547). 

I installed notes [from the aur](https://aur.archlinux.org/packages/notes) and also tried building it myself. Both methods crash, but when I try the AppImage, everything works fine. This might be unrelated, but I did notice that the window has the macOS buttons, when it shouldn't display buttons at all on a tiling window manager like bspwm:

image

GDB backtrace: [gdb.txt](https://github.com/nuttyartist/notes/files/9591481/gdb.txt)

zjeffer:
Fixes #346 by removing the line that sets `setDesktopSettingsAware` to false.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -11,7 +11,6 @@
 
 int main(int argc, char *argv[])
 {
-    QApplication::setDesktopSettingsAware(false);
     QApplication app (argc, argv);
     // Set application information
     app.setApplicationName ("Notes");


nuttyartist:
Looks good on macOS for now. I'll test on Linux and Windows in the coming days.

Also, we have some failing checks in Appveyor and Linux, maybe because the branch is behind some commits.

zjeffer:
Should be fixed now. @nuttyartist can you approve the workflows?

nuttyartist:
Approved. I'll test this on Linux and Windows today.

nuttyartist:
Looks good on Linux and Windows. Merging... Thanks @zjeffer!

PR #386, opened 2022-10-01
QT6 migration

### LINKED ISSUE #356: Converting codebase to Qt 6
macOS m1 version:12.6

compile error in 3rd party code:
```
In file included from ../3rdParty/qxt/qxtglobalshortcut.cpp:1:
In file included from ../3rdParty/qxt/qxtglobalshortcut.h:34:
../3rdParty/qxt/qxtglobal.h:200:19: error: allocating an object of abstract class type 'QxtGlobalShortcutPrivate'
        pvt = new PVT;
```

so is it ready for the compiling of new version qt?

zjeffer:
Will fix #356. 

# TODO

- [x] Fix all build errors
- [x] Fix all warnings
- [x] Remove all unnecessary `QT_VERSION_CHECK`s
- [x] Test `notes` on all platforms
    - [x] Linux (I tested ubuntu18 with Qt5.9.5, ArchLinux with Qt5.15.6, ArchLinux with Qt6.4.0)
    - [x] macOS
    - [x] Windows 10
    - [x] Windows 11 (tested with Qt 6.4.0)
- [x] Fix font weight issue 
- [x] Make backwards compatible down to Qt 5.9
- [x] Set QDataStream version to 5.6 for maximum compatibility
- [x] Fix resizing issue: the left pane seems to grow instead of the right pane when resizing.
- [x] QT6 builds with GitHub actions


src/updaterwindow.cpp
```diff
@@ -289,10 +288,10 @@ void UpdaterWindow::startDownload(const QUrl& url)
     }
 
     /* Start download */
-    m_startTime = QDateTime::currentDateTime().toTime_t();
+    m_startTime = QDateTime::currentDateTime().toSecsSinceEpoch();
```
zjeffer, line 291: https://doc.qt.io/qt-5/qdatetime-obsolete.html#toTime_t

src/notelistview.cpp
```diff
@@ -929,14 +929,17 @@ QPixmap NoteListViewPrivate::renderToPixmap(const QModelIndexList &indexes, QRec
         option.rect = paintPairs.at(j).rect.translated(-r->topLeft());
         const QModelIndex &current = paintPairs.at(j).index;
 #endif
+        Q_Q(const QAbstractItemView);
         adjustViewOptionsForIndex(&option, current);
-        delegateForIndex(current)->paint(&painter, option, current);
+        q->itemDelegateForIndex(current)->paint(&painter, option, current);
```
zjeffer, line 934: Still not sure what this actually does, but this fixes the build. Still needs to be thoroughly tested.

src/notelistview.cpp
```diff
@@ -929,14 +929,17 @@ QPixmap NoteListViewPrivate::renderToPixmap(const QModelIndexList &indexes, QRec
...
     return pixmap;
 }
 
 QStyleOptionViewItem NoteListViewPrivate::viewOptionsV1() const
 {
     Q_Q(const NoteListView);
-    return q->viewOptions();
+    QStyleOptionViewItem* option = new QStyleOptionViewItem();
+    q->initViewItemOption(option);
+    return *option;
```
zjeffer, line 944: Don't know what this code does but it builds :)

3rdParty/qxt/qxtglobalshortcut_p.h
```diff
@@ -71,7 +71,7 @@ class QxtGlobalShortcutPrivate
     static bool eventFilter(void* message);
 #   endif // Q_OS_MAC
 #   else
-    virtual bool nativeEventFilter(const QByteArray &eventType, void *message, long *result);
+    virtual bool nativeEventFilter(const QByteArray &eventType, void *message, long long *result);
```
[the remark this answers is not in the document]
bjorn: I would suggest to write [qintptr](https://doc.qt.io/qt-6/qtglobal.html#qintptr-typedef), as used in the `nativeEventFilter` declaration.

src/dbmanager.cpp
```diff
@@ -930,13 +925,8 @@ QList<NodeData> DBManager::readOldNBK(const QString fileName)
...
         file.open(QIODevice::ReadOnly);
         QDataStream in(&file);
-#if QT_VERSION >= QT_VERSION_CHECK(5, 6, 0)
-        in.setVersion(QDataStream::Qt_5_6);
-#elif QT_VERSION >= QT_VERSION_CHECK(5, 4, 0)
-        in.setVersion(QDataStream::Qt_5_4);
-#elif QT_VERSION >= QT_VERSION_CHECK(5, 2, 0)
-        in.setVersion(QDataStream::Qt_5_2);
-#endif
+        in.setVersion(QDataStream::Qt_6_2);
```
zjeffer, line 928: I chose Qt_6_2 here because that is the LTS version.

zjeffer:
@nuttyartist @bjorn PR is complete & commits are squashed. The code needs to be reviewed and the build needs to be tested on macOS & Windows.

src/notelistview.cpp
bjorn: But it allocates a `QStyleOptionViewItem` on the heap, which is copied before the function returns and then leaked. This temporary object should be placed on the stack:
```c++
    QStyleOptionViewItem option;
    q->initViewItemOption(option);
    return option;
```

src/styleeditorwindow.cpp
```diff
@@ -97,7 +97,7 @@ StyleEditorWindow::StyleEditorWindow(QWidget *parent) :
          childButton->installEventFilter(this);
     }
 
-    new QShortcut(QKeySequence(Qt::CTRL + Qt::SHIFT + Qt::Key_S), this, SLOT(toggleWindowVisibility()));
+    new QShortcut(QKeySequence(Qt::CTRL, Qt::SHIFT, Qt::Key_S), this, SLOT(toggleWindowVisibility()));
```
bjorn, line 100: These do not mean the same thing. First there was a shortcut that you could trigger by holding Ctrl and Shift while pressing S, whereas now you have a shortcut that it triggered by pressing Ctrl, then Shift, then S, in that order (though I'm not sure such a shortcut actually works).

The change should have been to use `Qt::CTRL | Qt::SHIFT | Qt::Key_S` (which fortunately also works in Qt 5).

src/taglistdelegate.cpp
```diff
@@ -13,9 +13,9 @@ TagListDelegate::TagListDelegate(QObject *parent) :
     m_displayFont(QStringLiteral("Roboto")),
     #endif
     #ifdef __APPLE__
-    m_titleFont(m_displayFont, 13, 65),
+	m_titleFont(m_displayFont, 13, QFont::DemiBold),
     #else
-    m_titleFont(m_displayFont, 10, 60),
+	m_titleFont(m_displayFont, 10, QFont::DemiBold),
```
bjorn, line 18: This change makes the `#ifdef` pointless. Someone (probably @nuttyartist) would need to judge whether we still need a slightly more bold title font on macOS.

(this change repeats itself many times, which probably calls for a shared function)

3rdParty/qxt/qxtglobalshortcut.cpp
```diff
@@ -35,9 +35,6 @@
 
 #ifndef Q_OS_MAC
 int QxtGlobalShortcutPrivate::ref = 0;
-#   if QT_VERSION < QT_VERSION_CHECK(5,0,0)
-QAbstractEventDispatcher::EventFilter QxtGlobalShortcutPrivate::prevEventFilter = 0;
-#   endif
```
bjorn, line 40: I'm not sure if you made these changes or took them from the source, but I would suggest that if possible, we update qxt from https://github.com/hluk/CopyQ/tree/master/qxt (at least, this was the original source - 0a2b49526dcb1e95c805c4efeb97172d05049fc6) in a separate PR.

bjorn (commented):
I've left a few comments.

src/taglistdelegate.cpp
zjeffer: The font size is still different, which is why the `#ifdef` is still there. I fully agree with the shared function idea, though. @nuttyartist What is your opinion on this?

src/mainwindow.cpp
```diff
@@ -440,46 +439,46 @@ void MainWindow::setupTrayIcon()
 void MainWindow::setupKeyboardShortcuts()
 {
     new QShortcut(QKeySequence(Qt::Key_F10), this, SLOT(onDotsButtonClicked()));
```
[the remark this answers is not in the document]
zjeffer: TODO: 

- [ ] change these to (Qt::Key1 | Qt::Key2 | ...) instead of multiple arguments

nuttyartist:
So, will we have one codebase for both Qt6 and Qt5?

zjeffer:
> So, will we have one codebase for both Qt6 and Qt5?

In my opinion, that would be best. The changes between Qt 5 and 6 are not that large, judging by the relatively small amount of work I've had to do for this PR.

src/taglistdelegate.cpp
zjeffer: It looks like many classes initialize the same font options in their constructor. Would it be a better idea to let them inherit from some kind of "Widget with default font" class, so every class that inherits from this gets the same font options (instead of creating a function).

Particularly the `*delegatedEditor`s use exactly the same code, but there are many more classes that use very similar code for fonts.

bjorn: @zjeffer In general, inheritance is not a good option for solving such a problem because we can only inherit from a single class. So you can't inherit from both the "widget with default font" and a "button widget", for example.

Another problem with inheritance is that it would still mean every widget carries its own `QString m_displayFont` and `QFont m_titleFont` members, which is just a waste of computation every time such a widget is instantiated and a waste of memory. The `m_displayFont` isn't even used at all anymore after the `QFont` has been created.

In my opinion it's better to create for example a singleton that will initialize and provide access to the fonts to be used in the application:

```c++
static QString displayFont()
{
#ifdef __APPLE__
    return QFont(QStringLiteral("SF Pro Text")).exactMatch() ? QStringLiteral("SF Pro Text") : QStringLiteral("Roboto");
#elif _WIN32
    return QFont(QStringLiteral("Segoe UI")).exactMatch() ? QStringLiteral("Segoe UI") : QStringLiteral("Roboto");
#else
    return QStringLiteral("Roboto");
#endif
}

class Fonts {
public:
    static Fonts &instance()
    {
        static Fonts fonts;
        return fonts;
    }

    const QFont &titleFont() const { return m_titleFont; }

private:
    Fonts() :
#ifdef __APPLE__
        m_titleFont(displayFont(), 13, 65)
#else
        m_titleFont(displayFont(), 10, 60)
#endif
    {}

    QFont m_titleFont;
};
```
And then you can access the font like  `m_label->setFont(Fonts::instance().titleFont())`, for example.

(however, I would encourage you to open separate PRs for such a change, instead of trying to fit all the "Qt 6" related changes in this one)

zjeffer:
I squashed commits & addressed the code review comments, now only needs to be tested on macOS and Windows.

zjeffer:
I fixed a problem with the windows build (this means we need a windows qt6 build check before we merge). 

I also noticed that on both Windows and Linux, resizing works strangely. Whenever I start to resize, the left pane grows incredibly large, causing the middle and right panes to shrink down a lot. This happens on both Qt 5.15 and 6.4. Doesn't happen on Notes v2.0.0. Will investigate.

nuttyartist:
Watching your updates. Let me know when you want me to test it.

zjeffer:
Not sure why, but the appveyor build fails because it can't find `qmake`. Is it trying to search for Qt5's qmake? I have 0 experience with appveyor so I'll need some help with this.

guihkx:
I have added initial support for Qt 6 builds on GitHub Actions on my local [qt6](https://github.com/guihkx/notes/tree/qt6) branch (it's syncd with this PR's branch). Right now, only the Qt 6 build of Windows is failing:

https://github.com/guihkx/notes/actions/runs/3304734012

Feel free to incorporate my commits on this PR for your tests.

nuttyartist:
Should we drop Appveyor in favor of GitHub Actions? We already do in Actions whatever we are doing in Appveyor right?

zjeffer:
> Should we drop Appveyor in favor of GitHub Actions?

Yes, but I think we should also include 5.9 builds in windows.yml.

@nuttyartist Could you take a look at the latest windows qt6 build that fails? This is the error in framelesswindow.cpp (line 275): `No member named 'getContentsMargins' in 'QMainWindow'`. `getContentsMargins` is defined in QLayout, not in QMainWindow. What were you trying to do here?

guihkx:
>@nuttyartist:
>Should we drop Appveyor in favor of GitHub Actions? We already do in Actions whatever we are doing in Appveyor right?

I suppose we can do that, yes.

>@zjeffer
>Yes, but I think we should also include 5.9 builds in windows.yml.

Hmm, is that really necessary, though? Aren't we only supporting 5.9 because of Ubuntu 18.08? 🤔

zjeffer:
> Hmm, is that really necessary, though? Aren't we only supporting 5.9 because of Ubuntu 18.08?

Ah, of course. Is there no Windows version still using 5.9?

guihkx:
As far as I know, the only way to install Qt on Windows is through the official installer provided by Qt (or via aqtinstall, but I'm not sure if that's a common thing to do).

So, different from Ubuntu (and Linux distros in general), on Windows it's very easy to upgrade to the latest Qt release.

guihkx:
FWIW, you can also remove `packaging/windows/Notes_Inno_Script.iss` (unless @nuttyartist still needs it?).

nuttyartist:
> FWIW, you can also remove `packaging/windows/Notes_Inno_Script.iss` (unless @nuttyartist still needs it?).

Yes, you can remove it.

zjeffer:
@nuttyartist The appveyor build still runs for some reason. I guess you need to remove some kind of appveyor webhook? 

This PR is now ready for testing.

nuttyartist:
> @nuttyartist The appveyor build still runs for some reason. I guess you need to remove some kind of appveyor webhook?
> 
> This PR is now ready for testing.

Yes, I removed the webhook.

Great, I'll test it soon.

nuttyartist:
Works well on macOS, compiles on QT6.4 and Qt5.15.2. 👍 
Will now test it on Windows.

nuttyartist:
Merging... @zjeffer Thank you so much! Great work.

PR #388, opened 2022-10-01
Reset formatting with Ctrl-Space

AdrianTM:
Pressing Ctrl-Space removes format char `,~,*,**,# from selected text, or from the current block under cursor This removes all these special chars regardless if they are used for formatting or not, so it could be subsequently improved by searching for matching pairs for the special chars and by checking if #s are functioning as heading markers


nuttyartist:
Hi! 

1. Could you please explain again what's the use for this? I didn't get it from your description.
2. Ctrl+Space is used in macOS already for summoning Spotlight. I also think it's too common. You may suggest another shortcut.

AdrianTM:
OK, not fully aware of Mac shortcuts, this is a pretty common shortcut in other programs: Gmail (I think Ctrl-\ is an alternative), I think Outlook has it too. The idea is that you reset the formatting of the selected text, you press a key combination and formatting of the text is cleared.

What shortcut would you prefer?

nuttyartist:
Ctrl-\ Sounds good. 👍

AdrianTM:
Thanks, changed.

nuttyartist:
Great, merging... Thank you!

Issue #369, opened 2022-09-22
Toggling the folders tree makes notes list visible


bjorn:
I noticed that toggling the folders tree through the menu (either on or off) also makes the notes list visible. This action should not affect the visibility of the notes list in my opinion.

nuttyartist:
In what scenario would you want to see only the folder tree and the editor?

nuttyartist:
Oh, you're right! Thanks for reporting.

PR #395, opened 2022-10-10
toogle folder tree should not make note list visible

diepdtnse03145:
Fixed the bug described in issue #369


nuttyartist:
Works, thanks! Merging...

Issue #311, opened 2022-09-10
App crashing after moving tag item to the last position


nuttyartist:
The app crashes when moving a tag item to the last position in the tags list.

PR #396, opened 2022-10-10
fix app crashing after moving tag item to the last position #311

diepdtnse03145:
Fix crash described in issue #311


nuttyartist:
Seems to be fixed. Merging...

Issue #315, opened 2022-09-10
On clicking on notesList when focusing on a new empty note, we should delete the note


nuttyartist:
Same way when clicking on a different note while focusing on a new empty note, remove the empty note. We should the same when the user clicks on the notesList.

PR #397, opened 2022-10-10
when clicking on note list while focusing on a new empty note, remove the empty note

diepdtnse03145:
fix issue  #315


nuttyartist:
It works only when the current notesList isn't empty. When the notesList is empty, and then only with a New Note it doesn't work.
short

nuttyartist:
Works now, thanks!

Issue #184, opened 2017-10-21
App is crashing on Windows 7


nuttyartist:
A friend of mine tried to install the app, but get an error message that says that "MSVCP120.dll is missing".

nuttyartist:
@theshadowx Maybe it's because our config on AppVeyour doesn't include that? Should we include that?

theshadowx:
I think the installer you made doesn't include it, as windeployqt doesn't add it.
In the appveyor, I have manually added it the the folder from which the bunder will be generated : 
https://github.com/nuttyartist/notes/blob/dev/appveyor.yml#L63-L67

```
  - cd C:\ci\notes\bin
  - windeployqt .
  - if %compiler%==msvc2015 copy "C:\\Program Files (x86)\\Microsoft Visual Studio 14.0\\VC\\redist\\x86\\Microsoft.VC140.CRT\\msvcp140.dll" .
  - if %compiler%==msvc2015 copy "C:\\Program Files (x86)\\Microsoft Visual Studio 14.0\\VC\\redist\\x86\\Microsoft.VC140.CRT\\vcruntime140.dll" .
  - if %compiler%==msvc2015_64 copy "C:\\Program Files (x86)\\Microsoft Visual Studio 14.0\\VC\\redist\\x64\\Microsoft.VC140.CRT\\msvcp140.dll" .
  - if %compiler%==msvc2015_64 copy "C:\\Program Files (x86)\\Microsoft Visual Studio 14.0\\VC\\redist\\x64\\Microsoft.VC140.CRT\\vcruntime140.dll" .
  - iscc C:\ci\notes\packaging\windows\Notes_Inno_Script.iss
  - cd C:\ci\notes\installer

```

nuttyartist:
But it states "MSVCP120.dll" on my friend's machine, not msvcp140.dll. And indeed, my fault, I probably forgot to add these necessary libraries.

theshadowx:
yes because you used msvc 120 (2013) to compile the source code and the version used in appveyor is 140 (2015).

the dll follows the version of the compiler which you used

msvc 120 (2013) needs msvcp120.dll / vcruntime120.dll
msvc 140 (2015) needs msvcp140.dll / vcruntime140.dll

theshadowx:
just to check : go to your appveyor Notes page and select

![selection_334](https://user-images.githubusercontent.com/976569/31854578-7336030c-b693-11e7-96ff-8afd9831be97.jpg)
 
[...]

PR #399, opened 2022-10-10
General CI fixes

### LINKED ISSUE #310: Automatic updater doesn't seem to work on Windows
The automatic updater doesn't seem to work on Windows.

guihkx:
This PR mainly includes the required OpenSSL runtime libraries to make the auto-update feature work properly on Windows. These are the dlls:

- `libcrypto-1_1.dll` (32-bit) / `libcrypto-1_1-x64.dll` (64-bit)
- `libssl-1_1.dll` (32-bit) / `libssl-1_1-x64.dll` (64-bit)
- `msvcr100.dll` (required because these OpenSSL libraries were [compiled with MSVC 2010 :(](https://bugreports.qt.io/browse/QTBUG-78242))

I also included a forgotten dll *only* required for the x64 build:

- `vcruntime140_1.dll`

I only tested this on a Windows 7 (64-bit) virtual machine and everything seems to work just fine, but please test it on Windows 10 / Windows 11 as well.

Fixes #310.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -61,7 +61,13 @@ jobs:
         if: matrix.build-type == 'release'
         run: |
           Set-Location bin
-          Copy-Item -Path $env:VCToolsRedistDir\${{ matrix.arch }}\Microsoft.VC142.CRT\* -Include msvcp140.dll,msvcp140_1.dll,vcruntime140.dll,vcruntime140_1.dll
+          Copy-Item $env:VCToolsRedistDir\${{ matrix.arch }}\Microsoft.VC142.CRT\msvcp140.dll
+          Copy-Item $env:VCToolsRedistDir\${{ matrix.arch }}\Microsoft.VC142.CRT\msvcp140_1.dll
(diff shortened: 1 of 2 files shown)


nuttyartist:
Awesome.

Are you sure `msvcp140_1.dll` and `vcruntime140_1.dll` are unnecessary? I remember we had to include them see the discussion here: https://github.com/nuttyartist/notes/issues/184.

guihkx:
Thanks for making me double-check - you're absolutely right:

<p align="center">
  
</p>

It was only working before, because I had both x86 and x64 MSVC runtimes installed on my virtual machine (thank you, VMWare Tools 🙄):

<p align="center">
  
</p>

After uninstalling both runtimes and trying to run the portable builds of Notes, I was able to verify that:

- `msvcp140_1.dll` is required on *both* x64 and x86 builds.
- `vcruntime140_1.dll` is required *only* on x64 builds.

I don't know how this weird DLL naming scheme works, but as far as I could tell, `vcruntime140_1.dll` doesn't even have a x86 version:

<p align="center">
  
</p>

But it does exist for x64:

<p align="center">
  
</p>

nuttyartist:
Great, I will make sure to test on Windows before merging this.

PR #403, opened 2022-10-17
Fixed possible crash on exit

bjorn:
This change makes sure that any `NoteListDelegateEditor` instances are deleted before their `NoteListView` is destroyed. Previously, such instances could get destroyed after the `NoteListDelegateEditor` destructor, at which point the call to `unsetEditorWidget` would crash.

At the same time, it was also necessary to make sure the `NoteListModel` outlives the `NoteListView`, since the model is being used by `NoteListView::closeAllEditor`.

Changed `m_view` in `NoteListDelegateEditor` from `QListView*` to `NoteListView*`, which allowed getting rid of many instances of `dynamic_cast`.


mizhka:
At least it works for me. I see no more crash on exit. 
Thank you!

nuttyartist:
Looks good, merging... Thanks @bjorn!

PR #405, opened 2022-10-19
More CI fixes

### LINKED ISSUE #404: [GitHub Actions] Build jobs using `ubuntu-18.04` might fail arbitrarily :(
I just noticed that the `ubuntu-18.04` build job from commit https://github.com/nuttyartist/notes/commit/f53d059225ea7a5843a68b5b90461a8ad7b20731 has failed.

That happened because, specifically today, we're in GitHub's ["brownout period"](https://github.com/actions/runner-images/issues/6002), which *purposely* prevents any build job using `ubuntu-18.04` from running (due to this version being deprecated).

I honestly didn't know this was a thing, and so I expected the `ubuntu-18.04` runner to be working without any kind of restrictions until its final removal date, in April 1st of 2023...

Anyway, I think it's unfortunate and unacceptable to have build jobs failing randomly, so I either suggest upgrading the Ubuntu runner to `ubuntu-20.04` (which is the most recent LTS release), or maybe figure out if it's possible to build Notes within a Docker image of Ubuntu 18.04, within GitHub Actions...

Let me know your thoughts.

guihkx:
More details in the commit's description.

Highlights:

- Fix [deprecated](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/) `::set-output` command:
    image

- Introduce [problem matching](https://github.com/actions/toolkit/blob/main/docs/problem-matchers.md) for GCC, MSVC and Clang, which is especially helpful on PRs:
    image

And last, but not least:

Fixes #404 by building Notes within a Docker container image of Ubuntu 18.04, instead of using GitHub's official one.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -7,21 +7,41 @@ jobs:
   # So it doesn't make much sense to have different build types other than 'debug' here.
   # The release AppImage is built using aqtinstall instead (the job below this one).
   build-apt:
-    name: Build (${{ matrix.build-type }}, apt, ${{ matrix.os }})
+    name: Build (${{ matrix.build-type }}, apt, ${{ matrix.container != '' && matrix.container || matrix.os }})
     runs-on: ${{ matrix.os }}
+    container: ${{ matrix.container }}
     strategy:
       fail-fast: false
       matrix:
-        build-type:
-          - debug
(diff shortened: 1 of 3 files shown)


nuttyartist:
Woohoo! Thanks! Merging...

PR #420, opened 2022-10-29
Code fixes

AdrianTM:
pass non-trivial types by reference, pass trivially copyable types by value.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -2767,7 +2767,7 @@ void MainWindow::migrateFromV0_9_0()
  * \brief MainWindow::migrateNote
  * \param notePath
  */
-void MainWindow::migrateNoteFromV0_9_0(QString notePath)
+void MainWindow::migrateNoteFromV0_9_0(const QString &notePath)
 {
     QSettings notesIni(notePath, QSettings::IniFormat);
     QStringList dbKeys = notesIni.allKeys();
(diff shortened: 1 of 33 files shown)


src/nodetreeview.cpp
```diff
@@ -230,7 +230,7 @@ void NodeTreeView::onUpdateAbsPath(const QString &oldPath, const QString &newPat
     });
 }
 
-void NodeTreeView::updateEditingIndex(const QPoint& pos)
+void NodeTreeView::updateEditingIndex(QPoint pos)
```
zjeffer, line 233: What is your reasoning behind passing by value for these QPoint arguments? I know it's better to pass ints by value, but is this also the case for QPoint?

AdrianTM: QPoint is a "trivially copyable type" it's just a pair of ints, usually you don't pass ints by reference.

zjeffer (approved):

src/mainwindow.cpp
```diff
@@ -2932,7 +2932,7 @@ void MainWindow::dropShadow(QPainter& painter, ShadowType type, MainWindow::Shad
  * \param rect
  * \param gradient
  */
-void MainWindow::fillRectWithGradient(QPainter& painter, const QRect& rect, QGradient& gradient)
+void MainWindow::fillRectWithGradient(QPainter& painter, QRect rect, QGradient& gradient)
```
bjorn, line 2935: I feel like this better stay `const QRect &` since [`QPainter::fillRect`](https://doc.qt.io/qt-5/qpainter.html#fillRect-13) also takes a const reference (and I think Qt API does this in general). But I don't really care so if it's easier to pass by value it's fine with me.

bjorn (approved):
Just one small comment, but looks fine either way.

src/mainwindow.cpp
AdrianTM: I'm fine either way too, it's not a crucial thing.

PR #421, opened 2022-10-30
Qmake to cmake migration

### LINKED ISSUE #408: Migrate from qmake to CMake
With the conversion to QT6 (#386), it is also a good idea to migrate from `qmake` to CMake, as that is the recommended way to build QT6 projects. It should also be compatible with QT5, all the way down to 5.9

- Migrate to CMake
- Update build workflows to reflect this change
- Update build instructions in README
- ...

More info: https://doc.qt.io/qt-6/qt6-buildsystem.html

zjeffer:
Fixes actions/runner-images#408.

TODO:

- [x] Migrate from qmake to cmake 
- [x] Successful build and run
- [x] Fix Github Action workflows
  - [x] Linux
  - [x] Windows
  - [x] macOS


zjeffer:
Hi @guihkx, you seem to be familiar with macOS building, do you have time to fix the macOS builds? I already fixed the Windows and Linux builds, but I don't have a mac so that makes it very hard to debug.

guihkx:
It looks like it's failing because the macOS image provided by GitHub Actions comes with Python installed by default now? Nevertheless, if I read correctly, I think we can prevent homebrew from updating the dependencies by setting `HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK=1`.

>And don't mind the above commits, it took me a while to fix them lol

No worries, lol. Just squash them afterward.

zjeffer:
@guihkx Thanks. There's also a problem with the non-homebrew builds. Something about lambdas not supporting auto, maybe an old/wrong compiler version?

guihkx:
Upstream bug: https://github.com/actions/runner-images/issues/6817

People have provided some workarounds there. You might wanna try some of them in case `HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK=1` doesn't work.

If you do add a workaround though, please also add a TODO comment pointing to the issue, so we can later remove the workaround once it's fixed by GitHub.

guihkx:
> @guihkx Thanks. There's also a problem with the non-homebrew builds. Something about lambdas not supporting auto, maybe an old/wrong compiler version?

I'll take a look at that. But we might have to update our image to `macos-12`.

zjeffer:
Ok, thanks for your help!

.github/workflows/macos.yml
```diff
@@ -46,11 +46,13 @@ jobs:
         uses: ammaraskar/gcc-problem-matcher@0.2.0
 
       - name: Build (${{ matrix.build-type }})
+        env:
+            CMAKE_PREFIX_PATH: ${{ matrix.qt-version-major == '6'}} && ${{ env.Qt6_Dir }} || ${{ env.Qt5_Dir }}
```
guihkx, line 50: Where does `$Qt6_Dir` / `$Qt5_Dir` come from? Nevertheless this can be removed.

.github/workflows/macos.yml
```diff
@@ -46,11 +46,13 @@ jobs:
...
         run: |
           mkdir build
           cd build
-          qmake -config ${{ matrix.build-type }} ../src/Notes.pro
-          make -j$(sysctl -n hw.logicalcpu)
+          cmake .. -DCMAKE_BUILD_TYPE=${{ matrix.build-type }} -DCMAKE_PREFIX_PATH="${CMAKE_PREFIX_PATH}"
```
guihkx, line 54:
```suggestion
          cmake . -B build-dir -DCMAKE_BUILD_TYPE=${{ matrix.build-type }} -DCMAKE_PREFIX_PATH="$(brew --prefix qt@${{ matrix.qt-version-major }})"
```

Then the `mkdir` and `cd` lines can be removed. Just make sure to also update the path of the build directory in subsequent build steps.

Also, It was not necessary to use `-DCMAKE_PREFIX_PATH` on my macOS 11 VM, but for whatever reason we have to do this on GitHub Actions.

.github/workflows/macos.yml
```diff
@@ -46,11 +46,13 @@ jobs:
...
+          make -j$(nproc)
```
guihkx, line 55:
```suggestion
          cmake --build build-dir
```

P.S.: `nproc` doesn't exist on macOS.

If you want parallelism while building, you should export `CMAKE_BUILD_PARALLEL_LEVEL=$(sysctl -n hw.logicalcpu)`

.github/workflows/macos.yml
```diff
@@ -100,8 +102,8 @@ jobs:
         run: |
```
guihkx, line 102: Previous suggestions also apply here, with the exception of `-DCMAKE_PREFIX_PATH`, which doesn't need to be set since we're installing Qt with `aqtinstall`, and thankfully that just works.

guihkx:
Regarding the compilation errors, I believe we must require a compiler that supports at least the C++14 standard.

Add this to `CMakeLists.txt`:

```cmake
set(CMAKE_CXX_STANDARD 14)
set(CMAKE_CXX_STANDARD_REQUIRED ON)
```

Though I'd wait for other people to confirm this later, it's not exactly my expertise.

guihkx:
Heads up: You may want to sync your branch with `dev` to avoid merge conflicts.

I have some questions/suggestions about the changes I see in the Linux workflow, but I'll wait until you're happy with your changes and the PR is unmarked as draft.

.github/workflows/macos.yml
zjeffer: I had to add -DCMAKE_PREFIX_PATH because Qt gets installed in the `  cwd/Qt`   folder or something, so it doesn't find it by default.

.github/workflows/macos.yml
zjeffer: On Linux, I had to do this in order to point to the correct Qt installation. I had to change it to `Qt6_DIR`  , though (instead of  `..Dir`  ) Not sure why it doesn't fail on macOS.

zjeffer:
Still failing on the `brew update/upgrade`, while this shouldn't be an issue anymore: https://github.com/actions/runner-images/issues/2322 ?

guihkx:
> Still failing on the `brew update/upgrade`, while this shouldn't be an issue anymore: [actions/runner-images#2322](https://github.com/actions/runner-images/issues/2322) ?

That's from 2020, though... Check the new bug report I linked a few posts ago. 

`HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK=1` *should* work around it.

.github/workflows/macos.yml
guihkx: We're in the homebrew step, so they install it to `/opt/homebrew`, IIRC. It's `aqtinstall` that installs Qt on the current directory...

In any case, setting it to `-DCMAKE_PREFIX_PATH="$(brew --prefix qt@${{ matrix.qt-version-major }})"` should work...

zjeffer:
> That's from 2020, though... Check the new bug report I linked a few posts ago.

Oops, I completely forgot about that post, sorry. I just tested positive for Covid, I'll blame it on that :)

guihkx (changes requested):

.github/workflows/macos.yml
```diff
@@ -28,6 +28,7 @@ jobs:
 
       - name: Install Qt ${{ matrix.qt-version-major }} (homebrew)
         run: |
+          export HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK=1
```
guihkx, line 31: Nitpick: Move this env var out of the build commands and add a `TODO` pointing to the upstream issue:

```yaml
- name: Install Qt ${{ matrix.qt-version-major }} (homebrew)
  env:
    # TODO: Remove once https://github.com/actions/setup-python/issues/577 gets fixed.
    HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK: 1
  run: |
    brew update
```

.github/workflows/linux.yml
```diff
@@ -135,13 +162,13 @@ jobs:
...
-          sed -i.bak "/^VERSION *=/ s/$/-g${GITHUB_SHA::7}/" src/Notes.pro
+          sed -i.bak "s/PROJECT_VERSION *[^)]\+/&-g${GITHUB_SHA::7}/" CMakeLists.txt
 
       - name: Setup variables
         id: vars
         run: |
-          version=$(grep -m 1 VERSION src/Notes.pro | sed 's/^VERSION *= *//')
-          artifact_name="Notes_${version}-x86_64-Qt${{ matrix.qt-version }}"
+          version=$(grep -m 1 'PROJECT_VERSION' CMakeLists.txt | sed 's/.*PROJECT_VERSION *//' | sed 's/).*//')
+          artifact_name="Notes_${version}-x86_64"
```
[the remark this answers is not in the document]
zjeffer: Not sure how that happened, I don't remember removing that, strange.

.github/workflows/macos.yml
```diff
@@ -98,16 +98,15 @@ jobs:
...
+          cmake . -B build -DCMAKE_BUILD_TYPE=${{ matrix.build-type }}
+          cmake --build build
 
       - name: Build dmg (${{ matrix.build-type }})
         run: |
-          cd bin
+          cd build
           plutil -insert NSRequiresAquaSystemAppearance -bool true Notes.app/Contents/Info.plist
-          plutil -insert CFBundleName -string Notes Notes.app/Contents/Info.plist
+          plutil -replace CFBundleName -string Notes Notes.app/Contents/Info.plist
```
zjeffer, line 109: I had to change this to `replace` instead of `insert` because it would complain that the "Notes" already existed. Do you have an idea why this suddenly happens?

guihkx: My guess is that when we build with `qmake`, that property didn't get created at all, but now with CMake it does.

CMake probably sets it based on the project name (i.e. `notes`, lowercase), so this change is fine because `Notes` is preferable, I think...

guihkx:
From the Qt 6 [docs](https://doc.qt.io/qt-6/cmake-build-on-cmdline.html):

>For a consistent developer experience on all platforms, use the `Ninja` or `Ninja Multi-Config` generator.
>
>You can select the CMake generator either by setting the `CMAKE_GENERATOR` environment variable or using the `-G` argument:
>
>```
>cmake -G Ninja ...
>```


Didn't know that, but if the docs recommend it we should probably do it. Hopefully it works on Qt 5 builds too...

.github/workflows/linux.yml
```diff
@@ -127,6 +141,13 @@ jobs:
           ln -sf /usr/share/zoneinfo/UTC /etc/localtime
           echo UTC > /etc/timezone
 
+      - name: Setup CMake
```
guihkx, line 144: Unused step?

.github/workflows/linux.yml
```diff
@@ -42,6 +45,12 @@ jobs:
           echo 'deb https://ppa.launchpadcontent.net/git-core/ppa/ubuntu bionic main' > /etc/apt/sources.list.d/git.list
           apt update
 
+      - if: ${{ env.ACT }}
```
guihkx, line 48: Another leftover step?

zjeffer: I used this for debugging with https://github.com/nektos/act, will remove. It didn't really work that well anyway.

.github/workflows/linux.yml
zjeffer: Do I not need this to install CMake?

guihkx: Don't think so... You can install it directly using `apt`, which is preferred, IMO.

You can probably just add `cmake` to the `- name: Install other dependencies in docker container` step.

Please add it at the beginning of the list (just to keep the package names sorted alphabetically :D), and add a small comment to like I did for the other packages. Thank you.

guihkx: Although you'll probably need to move the timezone configuration at the top of this step, since `cmake` will probably pull a bunch of other packages that require the timezone to be configured. Take these lines:

```
# Configure the timezone, otherwise 'install-qt-action' gets stuck at 'configuring tzdata'.
ln -sf /usr/share/zoneinfo/UTC /etc/localtime
echo UTC > /etc/timezone
```

And move them at the very top of the step.

guihkx:
On Ubuntu the package is actually called [`ninja-build`](https://packages.ubuntu.com/search?keywords=ninja-build). 😅 

Quite odd that it doesn't come pre-installed on Linux, but it does on Windows...

Nevertheless, you'll have to install ninja on macOS too. For the `build-homebrew` job on macOS, you can try adding `ninja` to the `Install Qt ${{ matrix.qt-version-major }} (homebrew)` step.

Now for the `dmg-aqtinstall` job, you'll have to add a new step either before or after `Install Qt ${{ matrix.qt-version }} (aqtinstall)`.

.github/workflows/linux.yml
```diff
@@ -59,22 +62,27 @@ jobs:
...
         run: |
-          mkdir build
-          cd build
-          # Workaround for qtchooser not supporting Qt 6: https://bugs.launchpad.net/ubuntu/+source/qtchooser/+bug/1964763
-          echo -e '/usr/lib/qt6/bin\n/usr/lib/x86_64-linux-gnu' | sudo tee /usr/lib/x86_64-linux-gnu/qtchooser/qt6.conf
-          export QT_SELECT=qt${{ matrix.qt-version-major }}
-          qmake -config ${{ matrix.build-type }} ../src/Notes.pro
-          make -j$(nproc)
+          mkdir build && cd build
+          cmake .. -G Ninja -DCMAKE_BUILD_TYPE=${{ matrix.build-type }} -DCMAKE_PREFIX_PATH=${{ env.CMAKE_PREFIX_PATH }}
```
guihkx: `-DCMAKE_PREFIX_PATH=${{ env.CMAKE_PREFIX_PATH }}` can be removed now.

.github/workflows/linux.yml
```diff
@@ -59,22 +62,27 @@ jobs:
           QT_PKGS: ${{ matrix.qt-version-major == '6' && 'qt6-base-private-dev libgl-dev' || 'qtbase5-private-dev' }}
         run: |
           sudo apt update
-          sudo apt install -y build-essential qtchooser ${{ env.QT_PKGS }}
+          sudo apt install -y build-essential qtchooser ${{ env.QT_PKGS }} cmake ninja
```
guihkx: Please remove `qtchooser`, it was only necessary for `qmake`. And please install the packages in alphabetical order? Thanks 🥺❤️

.github/workflows/linux.yml
```diff
@@ -59,22 +62,27 @@ jobs:
...
+          CMAKE_PREFIX_PATH: ${{ matrix.qt-version-major == '6' && env.Qt6_DIR || env.Qt5_DIR }}
         run: |
-          mkdir build
-          cd build
-          # Workaround for qtchooser not supporting Qt 6: https://bugs.launchpad.net/ubuntu/+source/qtchooser/+bug/1964763
-          echo -e '/usr/lib/qt6/bin\n/usr/lib/x86_64-linux-gnu' | sudo tee /usr/lib/x86_64-linux-gnu/qtchooser/qt6.conf
-          export QT_SELECT=qt${{ matrix.qt-version-major }}
-          qmake -config ${{ matrix.build-type }} ../src/Notes.pro
-          make -j$(nproc)
+          mkdir build && cd build
```
guihkx, line 80: Nit:

```suggestion
          mkdir build
          cd build
```

.github/workflows/linux.yml
```diff
@@ -164,10 +175,10 @@ jobs:
 
       - name: Build (${{ matrix.build-type }})
         run: |
-          mkdir build
-          cd build
-          qmake -config ${{ matrix.build-type }} ../src/Notes.pro
-          make -j$(nproc)
+          mkdir build && cd build
```
guihkx, line 176: Nit:

```suggestion
          mkdir build
          cd build
```

.github/workflows/linux.yml
```diff
@@ -59,7 +62,13 @@ jobs:
           QT_PKGS: ${{ matrix.qt-version-major == '6' && 'qt6-base-private-dev libgl-dev' || 'qtbase5-private-dev' }}
         run: |
           sudo apt update
-          sudo apt install -y build-essential qtchooser ${{ env.QT_PKGS }}
+          sudo apt install -y build-essential qtchooser ${{ env.QT_PKGS }} cmake ninja-build
```
guihkx, line 65: Please remove `qtchooser`, it was only necessary for `qmake`. Also, can you please sort the package names alphabetically? 🥺 Thank you ❤️

.github/workflows/macos.yml
```diff
@@ -27,9 +27,12 @@ jobs:
           submodules: recursive
 
       - name: Install Qt ${{ matrix.qt-version-major }} (homebrew)
+        env:
+          # TODO: remove once https://github.com/actions/setup-python/issues/577 gets fixed.
+          HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK: 1
         run: |
           brew update
-          brew install qt@${{ matrix.qt-version-major }}
+          brew install qt@${{ matrix.qt-version-major }} ninja
```
guihkx, line 35: `ninja` before qt, please 🥺

.github/workflows/macos.yml
```diff
@@ -96,23 +98,27 @@ jobs:
           version: ${{ matrix.qt-version }}
           cache: true
 
+      - name: Install Qt ${{ matrix.qt-version-major }} (aqtinstall)
```
guihkx, line 101:
```suggestion
      - name: Install Ninja
```

.github/workflows/windows.yml
```diff
@@ -66,16 +66,16 @@ jobs:
         run: |
           New-Item build -ItemType Directory
           Set-Location build
-          qmake -config ${{ matrix.build-type }} ..\src\Notes.pro
-          nmake
+          cmake .. -G Ninja -DCMAKE_BUILD_TYPE=${{ matrix.build-type }}
```
guihkx, line 69: Does the `-B` switch not work on Windows? 🤔 

Just so we can get rid of the `New-item` / `Set-Location` lines above this one.

guihkx (commented):
Sorry in advance about my silly nitpicks about sorting things alphabetically... 😅

.github/workflows/linux.yml
zjeffer: What's the motivation behind this?

guihkx: The `&&` check is redundant. If `mkdir` fails, the whole step will also fail immediately.

.github/workflows/windows.yml
zjeffer: I believe I tried this and it stopped working once we switched the minimum cmake version to 3.10 in `CMakeLists.txt`. 

I'll test this again to be sure.

zjeffer: Looks like it does work :)

guihkx:
Awesome! Thanks a lot for the fixes! 🏆 

LGTM CI-wise, although I haven't actually tested any of the artifacts generated by it (just yet).

Anyway, I noticed that we had a couple of Linux-specific subcommands in `src/Notes.pro`, like `make deb` or `make appimage`.

I personally never used them, but I wonder if it's easy enough to re-add them here? 🤔

zjeffer:
> Awesome! Thanks a lot for the fixes!

Thank you very much for your help!
 
> Anyway, I noticed that we had a couple of Linux-specific [subcommands](https://github.com/nuttyartist/notes/blob/7f71dae920248a6b517e643d160d8a90656da675/src/Notes.pro#L146-L181) in `src/Notes.pro`, like `make deb` and `make appimage`.

I saw we already have make_deb.sh and make_snap.sh scripts in `packaging`, maybe we can just add a make_appimage.sh script as well, in another PR?

guihkx:
Sure. Sounds good to me.

CMakeLists.txt
```diff
@@ -0,0 +1,186 @@
...
+if(NOT DEFINED INSTALL_EXAMPLESDIR)
+    set(INSTALL_EXAMPLESDIR "examples")
+endif()
+
+set(INSTALL_EXAMPLEDIR "${INSTALL_EXAMPLESDIR}")
+
+find_package(QT NAMES Qt6 Qt5 REQUIRED COMPONENTS Core)
+
+if (NOT DEFINED QT_VERSION_MAJOR)
+    # message(FATAL_ERROR "Qt not found.")
```
guihkx, line 22: Leftover from debugging?

CMakeLists.txt
```diff
@@ -0,0 +1,186 @@
...
+target_include_directories(${PROJECT_NAME} PUBLIC
+    ${PROJECT_SOURCE_DIR}/src/
+    ${PROJECT_SOURCE_DIR}/3rdParty/QSimpleUpdater/include
+    ${PROJECT_SOURCE_DIR}/3rdParty/qautostart/src
+    ${PROJECT_SOURCE_DIR}/3rdParty/qmarkdowntextedit
+    ${PROJECT_SOURCE_DIR}/3rdParty/qxt
+)
+
+target_compile_definitions(${PROJECT_NAME} PUBLIC
+    APP_VERSION="2.0.0"
```
guihkx, line 134:
```suggestion
    APP_VERSION="${CMAKE_PROJECT_VERSION}"
```

Just so we don't have to update the version in two different places.

CMakeLists.txt
```diff
@@ -0,0 +1,186 @@
...
+set (CMAKE_CXX_STANDARD_REQUIRED ON)
+
+set(CMAKE_INCLUDE_CURRENT_DIR ON)
+
+set(CMAKE_AUTOMOC ON)
+set(CMAKE_AUTOUIC ON)
+set(CMAKE_AUTORCC ON)
+
+if(NOT DEFINED INSTALL_EXAMPLESDIR)
+    set(INSTALL_EXAMPLESDIR "examples")
```
guihkx, line 14: I'm not sure what this is. Are we using it, at all?

guihkx (commented):
Some minor things I noticed.

Btw, are you aware of something like clang-format for CMakeLists.txt? Just to help us keep the formatting consistent.

zjeffer:
> Btw, are you aware of something like clang-format for CMakeLists.txt? Just to help us keep the formatting consistent.

I found this: https://github.com/cheshirekow/cmake_format. I'm adding a `.cmake-format` file to the repo, and I'm currently trying to get the vscode extension to work.

CMakeLists.txt
zjeffer: I think I copied this from a QMake to CMake guide, but we're indeed not using this.

nuttyartist:
Thanks for this!

Btw, I just tested this PR using this binary (https://github.com/nuttyartist/notes/actions/runs/3809547762) on Windows, and the app icon doesn't show up (not on the desktop, task manager, install/uninstall). Is it related to this PR? Because it worked well on a previous PR (of the Qt6 Migration).

guihkx:
I also noticed that the version is getting set incorrectly... Will open a PR shortly to see if it fixes both of our problems.

zjeffer:
@guihkx Thanks! Can you explain what you mean by the version getting set incorrecty?

guihkx:
My guess is that `$CMAKE_PROJECT_VERSION` is undefined here:

https://github.com/nuttyartist/notes/blob/11028d0cd44e3db63a835e232269fd57fe0c7e17/CMakeLists.txt#LL126C11-L126C11

For that to work, I believe we should've set the version in `project()`.

Nevertheless this might also not be optimal, because [`$CMAKE_PROJECT_VERSION`](https://cmake.org/cmake/help/latest/variable/CMAKE_PROJECT_VERSION.html) was only introduced in CMake 3.12, but we still have to support 3.10 because of Ubuntu 18.04...

No worries, though. I'm preparing a fix.

Issue #354, opened 2022-09-17
Improve multi-tagging UX


udoprog:
Currently when tagging a note the context menu which is used to tag items is a bit noisy (says `Remove tag <tag>` / `Add tag <tag>`) and gets collapsed as you add or remove individual tags leaving you to have to "chase" the menu again for each tag added or removed.

![image](https://user-images.githubusercontent.com/111092/190875771-f7468034-ad7a-429d-ae00-de6a7d5f1418.png)

An improvement to me would be if:
* The menu indicates which tags apply with a checkbox next to each tag (along with its color) that can be toggled by clicking and;
* The context menu isn't collapsed by a tag being toggled but rather dismissed by e.g. clicking outside of the menu area.

Cheers!

nuttyartist:
> The context menu isn't collapsed by a tag being toggled but rather dismissed by e.g. clicking outside of the menu area.

This is very important, I had this on my mind but forgot to create an issue for that. Thanks!

> The menu indicates which tags apply with a checkbox next to each tag (along with its color) that can be toggled by clicking and;

What would this look like? Something like this?

![Screen Shot 2022-09-18 at 11 42 38 AM](https://user-images.githubusercontent.com/16375940/190891312-c3ecce28-edcc-4321-89ed-f4e05ac16543.png)

udoprog:
Yeah, that looks great, esp. with the grouping. I'd also consider dropping the extra wording (and use only the tag name) in favour of a hint when hovering.

Thanks!

PR #426, opened 2022-12-09
initial try to improve tagging experience, #354

nuttyartist:
I think a better way than what I'm doing here would be to do away with any text and just change the ✓ icon to an X icon on hovering over it. What do you think?

### DIFF
--- a/src/notelistview.cpp
+++ b/src/notelistview.cpp
@@ -732,7 +732,7 @@ void NoteListView::onCustomContextMenu(QPoint point)
         }
         contextMenu->clear();
         if (m_tagPool) {
-            auto tagsMenu = contextMenu->addMenu("Tags ...");
+            tagsMenu = contextMenu->addMenu("Tags ...");
             for (auto action : qAsConst(m_noteTagActions)) {
                 delete action;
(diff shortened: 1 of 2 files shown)


zjeffer:
I rebased so we're starting commit history from the newest dev branch (with the cmake additions, easier for development), and I changed the variable name to m_tagsMenu so it's clearer that it's a member and not a local variable.

As for the UI, I agree we should remove the text altogether and just  show the name of the tag with a checkmark next to it. I'm not sure if the QAction has hover capabilities, but it's a cool idea.

nuttyartist:
> I'm not sure if the QAction has hover capabilities, but it's a cool idea.

Yes. I couldn't find a way to implement that. I didn't merge this because I thought people need to get feedback that they could also remove a tag that already has a check icon when they hover over it (By changing the check icon to X).

nuttyartist:
Okay, unfortunately, I wasn't successful in 1. Keeping the context menu visible after choosing a tag. 2. Making the text/icon change upon hovering. But the changes are still better than before. So I'll merge this without closing #354 .

PR #430, opened 2022-12-19
external links: local and web, initial implementation

kir-gadjello:
This is a basic implementation of a feature I needed. Inspired by qmarkdowntextedit.

If you are interested, we could brainstorm the local link logic. For example: we could agree to use the logical path to note as well as the first header as an URL, for cases where such path is unique. Or we could institute a separate system of note headers.

Eventually, I'd like to make such intra-notes links work, and implement images as well.


zjeffer:
Maybe this can be fixed with the `enterEvent` signal on the document widget: https://stackoverflow.com/questions/8590705/qt-detect-if-mouse-is-over-certain-widget-even-if-that-widget-has-no-focus. I haven't looked into this that well, though.

In theory, this could also be fixed by always making the widget focused as long as the mouse is inside the document, but I understand this would be very annoying for some people.

bjorn:
Since this is handled entirely within the widget, it should be possible to just override `mouseMoveEvent`, `mousePressEvent` and `mouseReleaseEvent`. For the `mouseMoveEvent` to get delivered, it may be necessary to set the [`mouseTracking`](https://doc.qt.io/qt-6/qwidget.html#mouseTracking-prop) property.

A global event filter may still be needed for handling keyboard modifier keys, if we want them to change the cursor immediately and when the widget is not focused.

Another thing to note is that it should not be necessary to parse the entire note to collect all the URLs and then to iterate all these URLs to find out which one currently contains the cursor. Instead, one could for example get the cursor at the mouse position using [`cursorAtPosition`](https://doc.qt.io/qt-6/qtextedit.html#cursorForPosition), then use `cursor.block().text()` and search only this part of the text for URLs.

We might also try to use `QTextEdit::anchorAt`, if we are able to tell Qt where the anchors actually are.

nuttyartist:
Okay, let's merge this, I will open a new issue for:

- Changing the mouse icon on hover without focusing on textEdit
- Optimize getting the URL using the position of the cursor rather than reading the entire note.

PR #437, opened 2023-01-02
Build fixes for CMake and GitHub Actions

guihkx:
Sorry for the rather large pull request, but all the changes are kind of connected... So take your time to review it!

Possible highlights:

- CMake: Add an option to control which Qt version to use (`-DUSE_QT_VERSION=5` or `-DUSE_QT_VERSION=6`)
- CMake/macOS: Add an option to control what macOS version to target (defaults to 10.15)
- CMake: Re-introduce the missing version string
- macOS/Windows: Re-introduce the missing app icon

Happy new year! 🎉

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -218,5 +218,6 @@ jobs:
       - name: Upload AppImage artifact (${{ matrix.build-type }})
         uses: actions/upload-artifact@v3
         with:
+          if-no-files-found: error
           name: ${{ steps.vars.outputs.artifact_name }}-${{ runner.os }}-${{ matrix.build-type }}
           path: build/${{ steps.vars.outputs.file_name }}
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -203,7 +203,7 @@ jobs:
         run: |
           cd build
           install -Dm755 -t Notes/usr/bin notes
-          install -Dm755 -t Notes/usr/share/applications ../packaging/linux/common/notes.desktop
+          install -Dm644 -t Notes/usr/share/applications ../packaging/linux/common/notes.desktop
           mkdir -p Notes/usr/share/icons/hicolor
           cp -r ../packaging/linux/common/icons/* Notes/usr/share/icons/hicolor
           export VERSION='${{ steps.vars.outputs.version }}'
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -178,8 +178,12 @@ jobs:
         run: |
           mkdir build
           cd build
-          cmake .. -DCMAKE_BUILD_TYPE=${{ matrix.build-type }}
+          cmake .. -DCMAKE_BUILD_TYPE=${{ matrix.build-type }} -DCMAKE_INSTALL_PREFIX=/usr
           make -j$(nproc)
+
+      - name: Install (${{ matrix.build-type }})
+        run: |
+          make -C build DESTDIR=Notes install
 
       - name: Setup linuxdeploy
         run: |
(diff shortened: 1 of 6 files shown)


.github/workflows/linux.yml
```diff
@@ -163,13 +171,26 @@ jobs:
           sudo apt update
           sudo apt install -y libxkbcommon-dev libmysqlclient21
 
+      # TODO: Figure out why this error only occurs on the Linux container when building with -DGIT_REVISION=ON
```
zjeffer, line 174: Probably related to this: https://github.com/actions/runner-images/issues/6775

I think the newer git version installed by the `Install git PPA in docker container` step causes this.

.github/workflows/linux.yml
```diff
@@ -163,13 +171,26 @@ jobs:
...
+      # The error: fatal: detected dubious ownership in repository
+      - name: Prevent git's dubious ownership message
+        if: github.ref_type != 'tag'
+        run: |
+          git config --global --add safe.directory "${PWD}"
+
+      # TODO: Migrate to the Ninja generator once we drop support for Ubuntu 18.04.
```
zjeffer, line 181: Does the Ninja generator not work on 18.04?

.github/workflows/linux.yml
```diff
@@ -133,20 +133,28 @@ jobs:
...
-        shell: bash
-        run: |
-          sed -i.bak "s/[^_]PROJECT_VERSION *[^)]\+/&-g${GITHUB_SHA::7}/" CMakeLists.txt
-
       - name: Setup variables
+        shell: bash
         id: vars
         run: |
-          version=$(grep -m 1 '[^_]PROJECT_VERSION' CMakeLists.txt | sed 's/.*[^_]PROJECT_VERSION *//' | sed 's/).*//')
+          set -x
```
zjeffer, line 140: Is this still necessary?

zjeffer (approved):
Absolutely beautiful work! :)

.github/workflows/linux.yml
guihkx: Yeah, that's rather annoying. I forget the reason, but I think I added the PPA because the `checkout` action needs a newer version of git to work.

zjeffer:
Shouldn't we also support ubuntu22.04?

.github/workflows/linux.yml
guihkx: It does, but because Ubuntu 18.04 has an ancient version of CMake that doesn't support either `--install` and `--prefix`, I couldn't figure out a way to set the prefix on the `Install` step, which on macOS (for example), is invoked like this: `cmake --install --prefix foo`.

To be fair, I'm not even sure if that's the correct command, lol...

I really want to use the Ninja generator across all OSes, but Ubuntu 18.04 is kinda making that a PITA, so let me know if you know how to properly do that. Thanks!

zjeffer: No idea, let's just wait until we stop supporting 18.04 lol

.github/workflows/linux.yml
guihkx: Ah, yes. I forgot to mention that. I explicitly set this to make scripting errors stand out more. It's a bit verbose, I guess, but at the same time our scripts are rather small.

guihkx:
> Shouldn't we also support ubuntu22.04?

You mean in the context of the CI pipeline? If so, I believe we already cover that:

- Qt 5: https://github.com/nuttyartist/notes/actions/runs/3822031085/jobs/6501762670 
- Qt 6: https://github.com/nuttyartist/notes/actions/runs/3822031085/jobs/6501762733

Although I just noticed that the Qt 5 job is using `ubuntu-latest` instead of `ubuntu-22.04`, for some reason... lol

Currently it doesn't matter, since they refer to the same runner, but I think we should use `ubuntu-22.04` on both, to avoid ambiguity.

.github/workflows/linux.yml
guihkx: Right. lol. Just a few more months until official support is dropped... 🤞

zjeffer:
> Although I just noticed that the Qt 5 job is using `ubuntu-latest` instead of `ubuntu-22.04`, for some reason... lol
> 
> Currently it doesn't matter, since they refer to the same runner, but I think we should use `ubuntu-22.04` on both, to avoid ambiguity.

Ah, I must have missed it.

nuttyartist:
Works well on my Mac (: 👍

guihkx:
Awesome!

Thank you guys, I'll merge it then.

PR #439, opened 2023-01-10
Add the Qt 6 build to the unified Windows installer

### LINKED ISSUE #438: Truly unified WIndows installer
Right now we have a separate Windows installer for the Qt 6 build (which is x64-only), but we should probably just include the Qt 6 build with the Qt 5 ones (they're x86 and x64), and then maybe just install the Qt 6 build by default (if supported by the OS).

Possible improvements to the installer:

- Detect which version of Windows is running, and make a best-effort decision to pick which Qt build to install by default
- Allow users to choose which Qt build to install (maybe through an 'advanced mode' option?)

Some things to keep in mind:

- Qt 6 is x64-only and only supported on Windows 10+: https://doc.qt.io/qt-6/supported-platforms.html
- Having three distinct builds will obviously enlarge the final size of the installer, but by how much? :)

Feel free to post any suggestion you might have...

guihkx:
As discussed in #438, this PR includes the Qt 6 build (which is x64-only) to the unified Windows installer, alongside a x86 Qt 5 build, meant for older Windows versions and/or 32-bit CPUs.

By default, the installer will prioritize the Qt 6 build, [if supported](https://doc.qt.io/qt-6/supported-platforms.html#windows) by the host. Otherwise, the Qt 5 build is used.

Fixes #438.


nuttyartist:
Ready to merge IMO.

nuttyartist:
Thankss!

Issue #411, opened 2022-10-27
macOS Packaging: auto sign, auto notarize, create Universal app for arm chips


nuttyartist:
We need to add 3 more steps to our app packaging for macOS:

- [x] Automatic code sign
- [x] Automatic notarization
- [x] Creating Universal app for native support of arm chips (M1, etc...)

nuttyartist:
@guihkx How can we sign and notarize our app with GitHub Actions using my Apple Developer certificates? I tried to do it on my machine but only managed to sign and not notarize. Also, I think for transparency's sake, it's better we use a clean environment.

guihkx:
>How can we sign and notarize our app with GitHub Actions using my Apple Developer certificates?

I don't know yet. Can the process be done entirely via command line? If it can, then I think it should be feasible using '[secrets](https://docs.github.com/en/actions/security-guides/encrypted-secrets)' from GitHub Actions.

>Also, I think for transparency's sake, it's better we use a clean environment.

What do you mean by 'clean environment', exactly?

I'm also thinking that the signing and notarizing should only be done for release builds, because I'm not sure if it's possible for a person to send a malicious pull request and read your credentials...

bjorn:
> I don't know yet. Can the process be done entirely via command line?

Yes, this is possible. I'm personally [using some helper actions](https://github.com/mapeditor/tiled/blob/95fe033aa1cb13e39d3a7a5fce55056afc0b4713/.github/workflows/packages.yml#L221-L246) for this, though they are somewhat unmaintained so I would recommend to try to find more up-to-date ones or to perform the needed commands in a script instead. I think the main issue is stapling, since it requires waiting on the notarization to be done. The action I am using is doing this by polling the notarization service for a while.

> What do you mean by 'clean environment', exactly?

That would be the CI environment, as opposed to his local machine. :-)

> I'm also thinking that the signing and notarizing should only be done for release builds, because I'm not sure if it's possible for a person to send a malicious pull request and read your credentials...

It isn't possible for PR builds to access the secrets, but that does mean we should limit this to our own builds, to avoid CI errors.

nuttyartist:
> That would be the CI environment, as opposed to his local machine. :-)

Yes.

> I don't know yet. Can the process be done entirely via command line? If it can, then I think it should be feasible using '[secrets](https://docs.github.com/en/actions/security-guides/encrypted-secrets)' from GitHub Actions.

Ah great. Then I will create the appropriate secrets.

[...]

PR #441, opened 2023-01-14
Build universal x86_x64/arm64 Qt 6 binaries on macOS

guihkx:
Disabled for Qt 5 builds at the moment, as it's only available on commercial builds: https://www.qt.io/blog/qt-on-apple-silicon

This is part of #411.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -98,7 +98,12 @@ jobs:
           then
               version="${version}-g${GITHUB_SHA::7}"
           fi
-          artifact_name="Notes_${version}-Qt${{ matrix.qt-version }}-x86_64"
+          arches='x86_64'
+          if [[ '${{ matrix.qt-version }}' == 6.* ]]
+          then
+              arches+='-arm64'
+          fi
+          artifact_name="Notes_${version}-Qt${{ matrix.qt-version }}-${arches}"
           if [ '${{ matrix.build-type }}' == 'debug' ]
           then
               file_name="${artifact_name}-debug.dmg"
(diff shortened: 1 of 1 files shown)


nuttyartist:
Awesome! I have an M1 Mac lying around, so I can test this. Qt5 is not needed, so Qt6 is fine.

Where can I grab the artifacts from to test this? Only after merging?

guihkx:
Both arm64 and x86_64 binaries are actually bundled into a single binary, so you can test the x86_64 dmg should work on both architectures.

I should probably rename the artifact to include arm64, though...

nuttyartist:
> Both arm64 and x86_64 binaries are actually bundled into a single binary, so the x86_64 dmg should work on both architectures.
> 
> I should probably rename the artifact to include arm64, though...

Ah perfect. Yes, that will be good. Something with both x86_64 and arm64, or universal.

guihkx:
Done. Feel free to merge once you make sure it works.

nuttyartist:
Great I'll try to test it tonight.

nuttyartist:
It works very well on a Mac and runs natively on Arm now (I made sure it's not running in Rosetta 2).

But one thing that needs to change is that I always package Notes on Macs with the name "Notes Better" since all Macs come pre-installed with a "Notes" app so In order to not conflict with the app, I rename the package to "Notes Better," haha. So we should keep doing this in the Releases as well.

guihkx:
I see... Do you think it's enough to change the [CFBundleName property](https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundlename)?

nuttyartist:
So we can apparently drag the app to the "Applications" folder without causing any trouble except, oddly enough the app runs, but the indicator in the task manager is associated with the pre-installed Apple Notes app (our icon doesn't show).
I tested just changing `CFBundleName` to 'Notes Better' (also `CFBundleExecutable`) while keeping the app package name "Notes" but the same happens. Only when changing the package name it works properly.

guihkx:
Hmm, then I'm thinking it might just be a matter of renaming the `.app` folder from `Notes.app` to `Notes Better.app`.

At least that's what was done in v2.0.0, as `CFBundleExecutable` and `CFBundleName` are still set to `Notes`.

Let's try that first...

nuttyartist:
Sure, sounds good! Feel free to merge once you're ready.

guihkx:
I tested on a Big Sur VM and it seems to work, so let's merge this.

PR #445, opened 2023-01-15
Try let the window manager move the window, fixes #429

### LINKED ISSUE #429: Window is not moveable in Gnome 43.2 + Wayland
[Screencast from 2022-12-18 17-53-01.webm](https://user-images.githubusercontent.com/9459891/208296929-ed308bdf-a708-42d7-8dd7-2916a67179bf.webm)

Running on Fedora Workstation 37
Build commit [f71dae](https://github.com/nuttyartist/notes/commit/7f71dae920248a6b517e643d160d8a90656da675)

I am also facing this issue with a Qt 5 app I develop.

nuttyartist:
This should fix #429. Can you please review if the code is satisfactorily written?

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -2624,11 +2624,17 @@ void MainWindow::mouseMoveEvent(QMouseEvent *event)
 #  endif
 
     if (m_canMoveWindow) {
-        int dx = event->globalX() - m_mousePressX;
-        int dy = event->globalY() - m_mousePressY;
-        move(dx, dy);
-
+        #if QT_VERSION >= QT_VERSION_CHECK(5, 15, 0)
+        if (!window()->windowHandle()->startSystemMove()) {
+        #endif
+            int dx = event->globalX() - m_mousePressX;
+            int dy = event->globalY() - m_mousePressY;
+            move(dx, dy);
+        #if QT_VERSION >= QT_VERSION_CHECK(5, 15, 0)
+        }
+        #endif
     }
+
 #  ifndef __APPLE__
     else if (m_canStretchWindow && !isMaximized() && !isFullScreen()) {
         int newX = x();
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -2507,8 +2507,12 @@ void MainWindow::mousePressEvent(QMouseEvent *event)
         if (m_mousePressX < this->width() - m_layoutMargin && m_mousePressX > m_layoutMargin
             && m_mousePressY < this->height() - m_layoutMargin && m_mousePressY > m_layoutMargin) {
 
+#if QT_VERSION >= QT_VERSION_CHECK(5, 15, 0)
+            m_canMoveWindow = !window()->windowHandle()->startSystemMove();
+#else
             m_canMoveWindow = true;
             //            QApplication::setOverrideCursor(QCursor(Qt::ClosedHandCursor));
+#endif
 
 #  ifndef __APPLE__
         } else {
@@ -2624,15 +2628,9 @@ void MainWindow::mouseMoveEvent(QMouseEvent *event)
 #  endif
 
     if (m_canMoveWindow) {
-        #if QT_VERSION >= QT_VERSION_CHECK(5, 15, 0)
-        if (!window()->windowHandle()->startSystemMove()) {
-        #endif
-            int dx = event->globalX() - m_mousePressX;
-            int dy = event->globalY() - m_mousePressY;
-            move(dx, dy);
-        #if QT_VERSION >= QT_VERSION_CHECK(5, 15, 0)
-        }
-        #endif
+        int dx = event->globalX() - m_mousePressX;
+        int dy = event->globalY() - m_mousePressY;
+        move(dx, dy);
     }
 
 #  ifndef __APPLE__
@@ -2769,7 +2767,11 @@ void MainWindow::mousePressEvent(QMouseEvent *event)
         if (event->pos().x() < this->width() - 5 && event->pos().x() > 5
             && event->pos().y() < this->height() - 5 && event->pos().y() > 5) {
 
+#if QT_VERSION >= QT_VERSION_CHECK(5, 15, 0)
+            m_canMoveWindow = !window()->windowHandle()->startSystemMove();
+#else
             m_canMoveWindow = true;
+#endif
             m_mousePressX = event->pos().x();
             m_mousePressY = event->pos().y();
         }
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 build-Notes-*
 build-*
 bin*
+CMakeLists.txt.user
 
 # OS generated files #
 ######################
--- a/src/updaterwindow.cpp
+++ b/src/updaterwindow.cpp
@@ -562,8 +562,12 @@ void UpdaterWindow::mousePressEvent(QMouseEvent *event)
     if (event->button() == Qt::LeftButton) {
         if (event->x() < width() - 5 && event->x() > 5 && event->pos().y() < height() - 5
             && event->pos().y() > 5) {
+#if QT_VERSION >= QT_VERSION_CHECK(5, 15, 0)
+            m_canMoveWindow = !window()->windowHandle()->startSystemMove();
+#else
             m_canMoveWindow = true;
-            m_mousePressX = event->x();
+#endif
+            m_mousePressX = event->pos().x();
             m_mousePressY = event->pos().y();
         }
     }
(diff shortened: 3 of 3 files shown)


nuttyartist:
Whoops, I see I'm uploading here CMakeLists.txt.user. Is it not in gitignore?

CMakeLists.txt.user
```diff
@@ -0,0 +1,645 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
zjeffer, line 1: This should indeed be added in .gitignore.

guihkx:
> Whoops, I see I'm uploading here CMakeLists.txt.user. Is it not in gitignore?

It seems that file is generated by QtCreator (which I personally don't use), but it should be fine to add it to `.gitignore`.

src/mainwindow.cpp
```diff
@@ -2624,11 +2624,13 @@ void MainWindow::mouseMoveEvent(QMouseEvent *event)
 #  endif
 
     if (m_canMoveWindow) {
-        int dx = event->globalX() - m_mousePressX;
-        int dy = event->globalY() - m_mousePressY;
-        move(dx, dy);
-
+        if (!window()->windowHandle()->startSystemMove()) {
```
zjeffer, line 2627: startSystemMove was added in 5.15: https://doc.qt.io/qt-5/qwindow.html#startSystemMove, which is why this fails on the 5.9 build for ubuntu18.04. 

I think you should add a QT_VERSION_CHECK to test if the version is pre 5.15. Wayland users should use 5.15+ anyway.

zjeffer (changes requested):

src/mainwindow.cpp
guihkx, line 2627: This is only available on Qt >= 5.15:

CMakeLists.txt.user
zjeffer: You should also delete the file itself, I believe this will still get added to the repo because you've added the gitignore changes afterwards. @nuttyartist

src/mainwindow.cpp
bjorn, line 2627: I have no idea whether it is fine to continuously call this function on all platforms. Or does `mouseMoveEvent` stop being called when this function succeeded? The documentation doesn't say anything about it, so I guess we could just test.

Alternatively, this function could be called in `MainWindow::mousePressEvent` instead, where we maybe could do `m_canMoveWindow = !window()->windowHandle()->startSystemMove()` instead of just assigning `true` (untested)

bjorn (commented):
Note that you'll want to do the same change to the `UpdaterWindow`.

nuttyartist:
My C++ and Qt skills are quite rusty. If someone wants to take that PR, I'd be glad 😅

Also, why is after adding `CMakeLists.txt.user` to gitignore the file is still added?

> Note that you'll want to do the same change to the `UpdaterWindow`.

Good catch. I tried but with no success. I got `member access into incomplete type QWindow`

bjorn:
> Also, why is after adding `CMakeLists.txt.user` to gitignore the file is still added?

Because you added it to `.gitignore` after it was already added to git. Changing `.gitignore` is not going to rewrite history for you. :-)

If you want to remove that file from git without removing the file from your workspace, you can use `git rm --cached CMakeLists.txt.user` and commit that.

> member access into incomplete type QWindow

It means you need to add `#include <QWindow>` to the affected .cpp file.

nuttyartist:
> You're maintaining this thing, I think I'll let you the opportunity to get a little bit less rusty. ;-)

Haha fair. Thanks for the tip. Seems to work now. Let me know if it's fine.

@zjeffer Thanks 😉

zjeffer (dismissed):

zjeffer:
I fixed the qt5.9 build, rebased and squashed the commits.

@mominul can you test this on wayland by downloading the latest build from here: https://github.com/nuttyartist/notes/actions/runs/3976090004. At the bottom, you can download the AppImages for both Qt5 and Qt6.

guihkx:
The AppImages currently don't include the Wayland platform plugin, so it will just fallback to XWayland.

However, I *think* it's just a matter of installing the `qtwayland5` / `qtwayland6` packages and the `linuxdeploy` tool will pick up those plugins automatically.

If that doesn't work, we'll have to find a way to include the Wayland plugin manually.

I would add those changes myself, but I'm not home at the moment...

@zjeffer Feel free to experiment, if you want. If that doesn't work, I can pick it up from where you left later.

nuttyartist:
Feel free to merge this when it's ready.

guihkx:
Okay, I think we aren't quite there yet regarding Wayland support in AppImages. And even people much smarter than me seem to agree:

https://github.com/linuxdeploy/linuxdeploy-plugin-qt/pull/25

First, because it's a struggle to bundle all the required Wayland libraries. Even though I [succeeded](https://github.com/guihkx/notes/commit/20db9258782004fded6990abf17bfcf2619fa36b) (to the best of my knowledge, anyway).

But I either experienced random, weird behavior, like this one:

https://user-images.githubusercontent.com/626206/213934950-c608afc2-bd7d-47d0-acd2-c8526a429518.mp4

... Or just straight up the crashing of the whole desktop (wtf, GNOME?):

```
Jan 22 16:23:28 arch-des gnome-shell[798]: Buggy client caused popup to be placed outside of parent window
Jan 22 16:23:29 arch-des gnome-shell[798]: **
Jan 22 16:23:29 arch-des gnome-shell[798]: libmutter:ERROR:../mutter/src/wayland/meta-wayland-popup.c:233:meta_wayland_popup_grab_get_top_popup: assertion failed: (!wl_list_empty (&grab->al>
Jan 22 16:23:29 arch-des gnome-shell[798]: Bail out! libmutter:ERROR:../mutter/src/wayland/meta-wayland-popup.c:233:meta_wayland_popup_grab_get_top_popup: assertion failed: (!wl_list_empty >
Jan 22 16:26:46 arch-des systemd-coredump[2336]: [🡕] Process 798 (gnome-shell) of user 1001 dumped core.
```

Even if those are rare and I was just unlucky, I don't think it's something we want to ship to users, considering that the Xbc backend still works and offers a better experience overall.

guihkx:
Regarding the changes itself, I'm experiencing some very weird behavior while testing the Qt 5 build on Linux (**not running on Wayland**)...

After moving the window for a bit (and then releasing it), the mouse cursor will flicker like crazy sometimes:

https://user-images.githubusercontent.com/626206/213937218-35ac631d-a796-4607-b429-e2e0d9532606.mp4

Not only that, clicking on the text editor makes the window manager think we are grabbing the window again, lol.

zjeffer:
I realized I'm also getting some strange bugs:


https://user-images.githubusercontent.com/4633209/213937857-2236bb1c-742e-463a-8e1c-1e8cf6e97042.mp4



Here's me with this PR's branch on bspwm, which uses X11. Resizing with bspwm's keyboard shortcuts works fine: I press super+left click on the window to move, super+right click near a corner to resize. 

1) Moving the window using the window bar works fine. I don't get the bug you're having.
2) Resizing the window by dragging the top border continuously moves the window diagonally to the left until I release the mouse button. Only the top border seems to be interactive.
    * Resizing windows by dragging the border shouldn't even be possible in most tiling window managers. I've been using bspwm for around 5 years, and I've never come across an app that can be resized by dragging one of the window's borders. I always resize using keyboard shortcuts.
3) With native window frame disabled, double clicking the bar will create new borders around the whole window (see 0:40 in the video). This makes all borders have resizing handles, including the corners for diagonal resizing. Dragging these handles results in the same bug as point 2. 
4) Resizing this way widens the notes list pane to be about half the window width.

guihkx:
LOL, that's hilariously broken. We'll need to investigate some more...

nuttyartist:
So is it not preferable to let the window manager handle the move operation if supported?

guihkx:
>So is it not preferable to let the window manager handle the move operation if supported?

It definitely is, IMO. It's just that it's probably broken thanks to our own window management code, so we should fix it first.

nuttyartist:
So what's left here to fix? If this solution is still better than our previous one, I think we can merge this and open a new issue.

nuttyartist:
This is the only issue that currently remains for completing version 2.1.0.

bjorn:
> So what's left here to fix? If this solution is still better than our previous one, I think we can merge this and open a new issue.

I've pushed a change that addresses my comment.

> This is the only issue that currently remains for completing version 2.1.0.

While ideally we'd also improve / fix the resize behavior in 2.1, I think this PR is ready to be merged.

bjorn (approved):

guihkx:
Right now this change only affects non-Windows builds, but I think we should also add this functionality on Windows, which appears to be done here instead:

https://github.com/nuttyartist/notes/blob/1471e87bff5ad1e9b1c78f3e8870db1d9cdf91ef/src/mainwindow.cpp#L2806-L2814

guihkx:
Ah, the Windows code was added elsewhere, my bad.

LGTM!

guihkx (approved):

bjorn:
> Right now this change only affects non-Windows builds, but I think we should also add this functionality on Windows, which appears to be done here instead:

Yeah, in general I think we could do better in unifying certain code paths, but for now i just adjusted both versions of the `mousePressEvent`.

PR #446, opened 2023-01-15
CustomMarkdownHighlighter to override default styles

### LINKED ISSUE #389: QMarkdownTextEdit headers color should be the same color as text
The headers' color should be the same as the text color.

Screen Shot 2022-10-04 at 2 18 00 PM

zjeffer:
Fixes #389 

This class adds a way for us to set our own colors, fonts, etc. to the MarkdownHighlighter from the QMarkdownTextEdit submodule, by simply extending the class and overriding the styles that we want in the constructor.

### DIFF
--- a/src/customMarkdownHighlighter.cpp
+++ b/src/customMarkdownHighlighter.cpp
@@ -1,0 +1,11 @@
+#include "customMarkdownHighlighter.h"
+
+CustomMarkdownHighlighter::CustomMarkdownHighlighter(QTextDocument *parent,
+                                                     HighlightingOptions highlightingOptions)
+    : MarkdownHighlighter(parent, highlightingOptions)
+{
+    // set all header colors
+    for (int i = HighlighterState::H1; i <= HighlighterState::H6; ++i) {
+        _formats[static_cast<HighlighterState>(i)].setForeground(QColor(0, 0, 0));
+    }
+}
--- a/src/customMarkdownHighlighter.cpp
+++ b/src/customMarkdownHighlighter.cpp
@@ -4,8 +4,24 @@ CustomMarkdownHighlighter::CustomMarkdownHighlighter(QTextDocument *parent,
                                                      HighlightingOptions highlightingOptions)
     : MarkdownHighlighter(parent, highlightingOptions)
 {
-    // set all header colors
+}
+
+void CustomMarkdownHighlighter::setHeaderColors(QColor color)
+{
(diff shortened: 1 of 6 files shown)


zjeffer:
I temporarily chose #000000 as the color, which looks like this:

image


This feels a bit bland to me, which color should we use?

zjeffer:
Ah, I just saw in the issue that it should be the same as the text color, meaning it should also change when changing the theme. 

Changing to draft PR, will need some more time.

zjeffer:
Light mode:

image

---

Dark mode:

image


---

Sepia: 

image

src/customMarkdownHighlighter.cpp
```diff
@@ -0,0 +1,27 @@
...
+    for (int i = HighlighterState::H1; i <= HighlighterState::H6; ++i) {
+        _formats[static_cast<HighlighterState>(i)].setForeground(color);
+    }
+}
+
+void CustomMarkdownHighlighter::setTheme(Theme theme, QColor textColor)
+{
+    setHeaderColors(textColor);
+
+    switch (theme) {
```
zjeffer, line 21: In the future, we could add different highlighter settings for every theme here.

nuttyartist:
Oh, finally! That blue header was giving me headaches and was so out of place. I think the colors (The black on the white theme is contrasting too much?) need some more tinkering (also of the themes themselves), but it works perfectly well. I am merging this now and might change the color choices in a different PR. 

Thanks!

zjeffer:
> I think the colors (The black on the white theme is contrasting too much?) need some more tinkering

I agree, I think dark grey text would work better than completely black. Imo the other themes look great, though.

nuttyartist:
> > I think the colors (The black on the white theme is contrasting too much?) need some more tinkering
> 
> I agree, I think dark grey text would work better than completely black. Imo the other themes look great, though.

Yes? Someone told me the Dark theme might be too dark and suggested to go with grayish-blackish colors rather than completely dark, what do you think?

zjeffer:
Personally I'm a big fan of the [Nord theme](https://www.nordtheme.com/) (you can turn their dark mode in the top right). A very dark blue would be perfect for me.

Something else about the dark theme: the dividers are completely white, which looks a bit jarring to me. I think it would be better if they were light grey.

nuttyartist:
I'll try to find better colors and send a PR.

Issue #267, opened 2020-06-05
Bug related to reported Wifi spikes - Make sure not to ship the bearer plugin on Windows


nuttyartist:
Someone emailed us today about a bug related to Wifi Spikes. This might be an upstream problem. Here's the email:

```
Hi guys!

First of, thanks for the Notes. Really cool app which I've been using for years. However a very interesting bug-case recently happened.

I bought a new HP laptop, installed Windows 10 64-bit (1909), Drivers etc. and then Notes of course. After some time I noticed that my WiFi was lagging. It became annoying, so I started to dig into issue, spent a couple of days and finally found this -- a consistent (regular) WiFi spikes from low to huge. Interesting that if Notes are completely off, then no spikes. If Notes are open then I see this:

Ответ от 192.168.0.1: число байт=32 время=3мс TTL=63
Ответ от 192.168.0.1: число байт=32 время=1807мс TTL=63   <--------------
Ответ от 192.168.0.1: число байт=32 время=5мс TTL=63
Ответ от 192.168.0.1: число байт=32 время=5мс TTL=63
Ответ от 192.168.0.1: число байт=32 время=4мс TTL=63
Ответ от 192.168.0.1: число байт=32 время=3мс TTL=63
Ответ от 192.168.0.1: число байт=32 время=5мс TTL=63
Ответ от 192.168.0.1: число байт=32 время=4мс TTL=63
Ответ от 192.168.0.1: число байт=32 время=6мс TTL=63
Ответ от 192.168.0.1: число байт=32 время=18мс TTL=63
Ответ от 192.168.0.1: число байт=32 время=1934мс TTL=63   <--------------
Ответ от 192.168.0.1: число байт=32 время=4мс TTL=63
Ответ от 192.168.0.1: число байт=32 время=5мс TTL=63
Ответ от 192.168.0.1: число байт=32 время=4мс TTL=63
Ответ от 192.168.0.1: число байт=32 время=5мс TTL=63
Ответ от 192.168.0.1: число байт=32 время=5мс TTL=63
Ответ от 192.168.0.1: число байт=32 время=4мс TTL=63
Ответ от 192.168.0.1: число байт=32 время=5мс TTL=63
Ответ от 192.168.0.1: число байт=32 время=11мс TTL=63
Ответ от 192.168.0.1: число байт=32 время=1887мс TTL=63  <--------------
.....


[...]

PR #447, opened 2023-01-18
Installer: Automatically uninstall different versions on Windows

guihkx:
If one or more different versions of Notes are currently installed, we now tell users they must uninstall them before proceeding.

This behavior is enabled by default, but it can by completely bypassed by invoking the installer via the command line with the `/SKIPVERSIONCHECK` parameter.

That's mainly done because it's been reported in the past that some dlls included in previous versions of the installer would cause issues.

And because these dlls don't get removed automatically upon installing a new version, they could still cause issues...

We could also approach this problem by maintaining a list of problematic files in [`[InstallDelete]`](https://jrsoftware.org/ishelp/index.php?topic=installdeletesection), but in my opinion, it's much easier to just uninstall any version that's different from the one the user is trying to currently install.

Part of #267.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -216,7 +216,7 @@ jobs:
           sudo apt update
           sudo apt install -y libpq5 libodbc1
 
-      - name: Build AppImage (${{ matrix.build-type }})
+      - name: Deploy (${{ matrix.build-type }})
         env:
           APPIMAGE_EXTRACT_AND_RUN: 1
         run: |
@@ -225,8 +225,27 @@ jobs:
           mkdir -p Notes/usr/share/icons/hicolor
           cp -r ../packaging/linux/common/icons/* Notes/usr/share/icons/hicolor
           export VERSION='${{ steps.vars.outputs.version }}'
-          ./linuxdeploy-x86_64.AppImage --appdir Notes --output appimage --plugin qt
-          mv Notes*.AppImage '${{ steps.vars.outputs.file_name }}'
+          ./linuxdeploy-x86_64.AppImage --appdir Notes --plugin qt
+
+      - name: Remove unnecessary Qt plugins and libraries
+        shell: bash
+        run: |
+          set -x
+          set -e
+          cd build/Notes
+          if [[ '${{ matrix.qt-version }}' == 5.* ]]
+          then
+              # The bearer plugin has caused problems for us in the past. Plus, it was removed altogether in Qt 6.
+              rm -rv usr/plugins/bearer
+          fi
+
+      - name: Build AppImage (${{ matrix.build-type }})
+        env:
+          APPIMAGE_EXTRACT_AND_RUN: 1
+        run: |
+          cd build
+          ./linuxdeploy-x86_64.AppImage --appdir Notes --output appimage
+          mv -v Notes*.AppImage '${{ steps.vars.outputs.file_name }}'
 
       - name: Upload AppImage artifact (${{ matrix.build-type }})
         uses: actions/upload-artifact@v3
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -144,14 +144,29 @@ jobs:
         run: |
           cmake --install build --prefix .
 
-      - name: Build dmg (${{ matrix.build-type }})
+      - name: Deploy (${{ matrix.build-type }})
         run: |
           cd build
           plutil -insert NSRequiresAquaSystemAppearance -bool true Notes.app/Contents/Info.plist
           # Rename the app folder to "Notes Better", so it doesn't conflict with macOS' "Notes" app.
           mv Notes.app 'Notes Better.app'
-          macdeployqt 'Notes Better.app' -dmg
-          mv 'Notes Better.dmg' '${{ steps.vars.outputs.file_name }}'
+          macdeployqt 'Notes Better.app'
+
+      - name: Remove unnecessary Qt plugins and libraries
+        run: |
+          set -x
+          set -e
+          cd 'build/Notes Better.app'
+          if [[ '${{ matrix.qt-version }}' == 5.* ]]
+          then
+              # The bearer plugin has caused problems for us in the past. Plus, it was removed altogether in Qt 6.
+              rm -rv Contents/PlugIns/bearer
+          fi
+
+      - name: Build dmg (${{ matrix.build-type }})
+        run: |
+          cd build
+          hdiutil create -srcfolder 'Notes Better.app' -format UDZO -fs HFS+ -volname 'Notes Better' '${{ steps.vars.outputs.file_name }}'
 
       - name: Upload dmg artifact (${{ matrix.build-type }})
         uses: actions/upload-artifact@v3
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -83,6 +83,16 @@ jobs:
         run: |
           windeployqt build\bin
 
+      - name: Remove unnecessary Qt plugins and libraries
+        run: |
+          Set-Location build\bin
+          # We ship all required runtime DLLs individually.
+          Remove-Item -Verbose vc_redist.*.exe
+          if ('${{ matrix.qt-version }}'.StartsWith('5.')) {
+              # The bearer plugin has caused problems for us in the past. Plus, it was removed altogether in Qt 6.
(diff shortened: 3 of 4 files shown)


guihkx:
I hit enter by accident while I was still editing the the pull request, damn it. :p 

Anyway... If the installer detects that a different version is already installed, the user is presented with this prompt:



Clicking 'No' aborts the installation and the version that's currently installed is kept.

Clicking 'Yes' invokes the uninstaller in silent mode.

If the uninstall process is successful, the user sees the following message, and can continue the installation normally:



If the removal fails for whatever reason, the user sees the following error message, and the installer aborts:

fail

bjorn:
> If the removal fails for whatever reason, the user sees the following error message, and the installer aborts:

Is it in this case always possible to remove the reference that triggers the detection of the previous version by uinstalling from the Windows Control Panel? Or might users get stuck not being able to install, because they have outdated registry entries?

guihkx:
>Or might users get stuck not being able to install, because they have outdated registry entries?

I'm not exactly sure how something like that could happen, to be honest. Unless the user manually messed up the registry...

A possible scenario where uninstalling from the Control Panel could help, is that if the uninstaller file (`unins000.exe`) was deleted or got corrupt, and we couldn't invoke it.

Then uninstalling from the Control Panel would look like this (on Windows 7, anyway):

cpanel

And then Windows completely removes the broken registry entries for us.

guihkx:
On the other hand, I might have misunderstood your question.

If we don't find any of these registry entries simultaneously:

- `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\{44576435-E098-4581-A6D7-FD7FADC5B063}_is1\QuietUninstallString`
- `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\{44576435-E098-4581-A6D7-FD7FADC5B063}_is1\DisplayVersion`

Or, alternatively, these ones:

- `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\{44576435-E098-4581-A6D7-FD7FADC5B063}_is1\QuietUninstallString`
- `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\{44576435-E098-4581-A6D7-FD7FADC5B063}_is1\DisplayVersion`

The setup just assumes Notes is not installed and continues normally.

nuttyartist:
> but in my opinion, it's much easier to just uninstall any version that's different from the one the user is trying to currently install.

I think that's the preferred approach.

I like the process! Very good UX letting the user know what happens in each step. After the uninstall does the installation of the new version pop up?

Anyway, I'll test this very soon.

> Or, alternatively, these ones (because some older versions would run the setup in [64-bit install mode](https://jrsoftware.org/ishelp/index.php?topic=32vs64bitinstalls)):

Good catch.

guihkx:
>After the uninstall does the installation of the new version pop up?

Yup.

nuttyartist:
I tested now and it seems like the bearer folder with its content isn't deleted. (I installed version 1.5 before). Which should be?

Another thing, for the first prompt we should write:

```
Another version of Notes is installed and must be removed first. Your notes data will not be lost. Once the uninstall is complete, you'll be prompted to install the new version X.
```

And for the second prompt:

```
Uninstall of 1.5.0 is complete. Installation of version X will start after pressing 'OK'...`
```

nuttyartist:
Okay, the bearer is installed as part of this commit setup. But again, should it?

guihkx:
> Okay, the bearer is installed as part of this commit setup. But again, should it?

I thought the problematic plugin was only `qnativewifibearer.dll` (which isn't included in our build artifacts), that's why I didn't actually remove the entire `bearer` folder.

But since the plugin is deprecated anyway, I guess it's fine to remove the whole folder.

I'll also tweak those messages, as you suggested.

guihkx:
I have a suggestion regarding the first prompt:

>Your notes data will not be lost.

I think the 'not' can easily be missed there. How about something more clear, like:

>All your notes will be safe.

Still regarding the first prompt, the last phrase is not really accurate:

>you'll be prompted to install the new version X

We're not really checking if the version trying to be installed is newer than the one already installed; We're just checking if they're different.

That's because in a hypothetical scenario where a new version introduces a broken library, for example, the user should be able to downgrade Notes while also being able to uninstall the new (yet broken) version.

guihkx:
New changes:

- Removed the Qt `bearer` plugin from Linux, macOS and Windows builds.
- There's now a way to disable all this version checking altogether: Invoke the installer using the command line and pass  `/SKIPVERSIONCHECK` as a parameter.
- We now respect (and also forward to the uninstaller) the following Inno Setup [command line parameters](https://jrsoftware.org/ishelp/index.php?topic=setupcmdline): `/SILENT`, `/VERYSILENT` and `/SUPPRESSMSGBOXES`. The only exception to this are error message box, which are always shown.

---

New screenshots of the dialogs (feedback/suggestions are welcome):

prompt

uninstall-ok

uninstall-error

.github/workflows/linux.yml
```diff
@@ -225,8 +225,27 @@ jobs:
...
-          mv Notes*.AppImage '${{ steps.vars.outputs.file_name }}'
+          ./linuxdeploy-x86_64.AppImage --appdir Notes --plugin qt
+
+      - name: Remove unnecessary Qt plugins and libraries
+        shell: bash
+        run: |
+          set -x
+          set -e
+          cd build/Notes
+          if [[ '${{ matrix.qt-version }}' == 5.* ]]
```
zjeffer, line 236: Couldn't we move this if statement to the step itself, something like this:


```yaml
name: Remove unnecessary Qt plugins and libraries
if: matrix.qt-version == 5.*
```

Not sure if this exact code works, but something like this would make the step be skipped altogether on Qt6 builds. Same thing with the macOS and Windows builds.

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif

guihkx: Yes, GitHub allows us to use [`startsWith()`](https://docs.github.com/en/actions/learn-github-actions/expressions#startswith). However, [I plan](https://github.com/guihkx/notes/commit/c39a073d7aa358990196e85aa41634624068ce39) on also removing non-SQLite plugins, and those are present on both Qt 5 and Qt 6 builds, so this step will end up having to run on both builds anyway...

nuttyartist:
> Don't worry, all your notes will be safe.

Yes, this is better.

> We're not really checking if the version trying to be installed is newer than the one already installed; We're just checking if they're different.

> That's because in a hypothetical scenario where a new version introduces a broken library, for example, the user should be able to downgrade Notes while also being able to uninstall the new (yet broken) version.

Ah, okay. But I still think it's important to let them know that there requested installation will begin after they click OK.

Regarding the right type of silent installation, what do you guys think? Should we go with /SILENT or /VERYSILENT? I know many companies these days do updates in the background. Maybe it will be less confusing for the end user because it can look like something is installing rather than uninstalling.

guihkx:
>Ah, okay. But I still think it's important to let them know that there requested installation will begin after they click OK.

Sure! That's done already.

>Regarding the right type of silent installation, what do you guys think? Should we go with /SILENT or /VERYSILENT?

Did you mean silent *uninstallation*, perhaps?

>Maybe it will be less confusing for the end user because it can look like something is installing rather than uninstalling

That's a good point. However, I do appreciate the visual feedback `/SILENT` offers (the uninstall progress bar).

But I'm fine with either.

nuttyartist:
> Sure! That's done already.

Perfect

> Did you mean silent uninstallation, perhaps?

Yes, sorry!

Let's go with /VERYSILENT.

DAW-GUN:
Just jumping in to say, great work guys! I just installed Notes on the computer in my other office and it found the old version and uninstalled it first. Kudos!

nuttyartist:
> Just jumping in to say, great work guys! I just installed Notes on the computer in my other office and it found the old version and uninstalled it first. Kudos!

Kudos to @guihkx 👍 👍

guihkx:
Alright, we now run the uninstaller with `/VERYSILENT`.

I think this is good to merge now, so I'll do that. If you encounter problems with the installer, please open a new issue and tag me.

Thank you guys for testing!

nuttyartist:
Thanks! @guihkx

PR #448, opened 2023-01-19
Fix a couple of deprecation warnings

guihkx:
I noticed these three warnings in our CI pipeline and decided to fix them:

image

image

### DIFF
--- a/src/updaterwindow.cpp
+++ b/src/updaterwindow.cpp
@@ -294,9 +294,8 @@ void UpdaterWindow::startDownload(const QUrl &url)
     m_startTime = QDateTime::currentDateTime().toSecsSinceEpoch();
     QNetworkRequest netReq(url);
 
-#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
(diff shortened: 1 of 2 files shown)


nuttyartist:
Ha! I was about to report these.

Tested and words fine. No warnings anymore. Thanks!

src/updaterwindow.cpp
```diff
@@ -294,9 +294,8 @@ void UpdaterWindow::startDownload(const QUrl &url)
     m_startTime = QDateTime::currentDateTime().toSecsSinceEpoch();
     QNetworkRequest netReq(url);
 
-#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
-    netReq.setAttribute(QNetworkRequest::FollowRedirectsAttribute, true);
-#endif
+    netReq.setAttribute(QNetworkRequest::RedirectPolicyAttribute,
+                        QNetworkRequest::NoLessSafeRedirectPolicy);
```
bjorn, line 298: Just a note, but that should be the default value for `QNetworkRequest::RedirectPolicyAttribute` according to the documentation, so this line could then be removed entirely.

Issue #450, opened 2023-01-19
Usablity issues on Linux


Yasand123:
This is an amazing note-taking application, simple yet efficient. Unfortunately I found some bugs/usability issues on Linux that hinder the overall experience. I understand this is a work-in-progress and I'm sure the devs are aware, but I'm mentioning these issues just in case they were overlooked/missed.

- When closing the window with `native window frame` enabled, it actually closes the app instead of minimizing to systray.
- Left clicking on the tray icon doesn't restore the window, you'd have to right click to bring up the context menu and then click on _Show notes_.
- ~~`Win+N` doesn't work.~~
- ~~When restoring the window from the tray icon, it gets behind the current window, and not in front as you'd expect.~~
- ~~Upon hiding and unhiding the sidebar, the in-app _close, minimize and maximize_ buttons show up even when _native window frame_ is enabled.~~

OS: Arch Linux.
DE: KDE Plasma.

guihkx:
>When closing the window with `native window frame` enabled, it actually closes the app instead of minimizing to systray.

There's an ongoing discussion in #278 about that.

>`Win+N` doesn't work.

I think the shortcut is actually <kbd>Win</kbd>+<kbd>Shift</kbd>+<kbd>N</kbd>, but AFAIK it only works correctly on Qt 5 builds (we now support Qt 6 as well). And judging by my tests, Notes has to be closed to the system tray too, I guess.


>Upon hiding and unhiding the sidebar, the in-app close, minimize and maximize buttons show up even when native window frame is enabled.

I think that was already fixed in #344.

nuttyartist:
Hey @Yasand123!

> When closing the window with native window frame enabled, it actually closes the app instead of minimizing to systray.

Right now, only on Mac does the app hides when closing the window rather than quitting the app because this is the custom around macOS apps. I thought on Windows and Linux, the expected behavior is to quit the app. But since many are saying it will be better just to hide the app, I'm willing to reconsider this. I will reflect on this decision in #278.

> Win+N doesn't work.

That's weird; it should work.

> Left clicking on the tray icon doesn't restore the window, you'd have to right click to bring up the context menu and then click on Show notes.

That is true and very annoying! I'll open an issue about that. Thanks for reporting.

> Upon hiding and unhiding the sidebar, the in-app close, minimize and maximize buttons show up even when native window frame is enabled.

[...]

PR #451, opened 2023-01-20
Toggle window visibility upon system tray activation

### LINKED ISSUE #452: Clicking on system tray icon should trigger show/focus main window
As @Yasand123 said,

> Left clicking on the tray icon doesn't restore the window, you'd have to right click to bring up the context menu and then click on Show notes.

Left-clicking on the system tray icon should trigger show/focus main window.
Should clicking again (after the window is showing) hide the window? I don't think so.

guihkx:
This will either show or hide the main window upon activating the system tray icon.

- On Windows and macOS, this can be triggered by a single left mouse click on the icon.
- On Linux, this will depend on the system tray implementation, but on KDE Plasma and XFCE, this is also triggered by a single left mouse click, while on GNOME (which requires a [third-party system tray extension](https://github.com/ubuntu/gnome-shell-extension-appindicator)), it's two left mouse clicks.

Part of #450 
Closes #452

---

There's also a (somewhat) related changed I added, which is calling `raise()` followed by `activateWindow()`. The Qt documentation [states](https://doc.qt.io/qt-5/qwidget.html#activateWindow):

>If you want to ensure that the window is stacked on top as well you should also call [raise](https://doc.qt.io/qt-5/qwidget.html#raise)(). Note that the window must be visible, otherwise activateWindow() has no effect.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -53,9 +53,11 @@ MainWindow::MainWindow(QWidget *parent)
       m_editorDateLabel(Q_NULLPTR),
       m_splitter(Q_NULLPTR),
       m_trayIcon(new QSystemTrayIcon(this)),
+#if !defined(Q_OS_MAC)
       m_restoreAction(new QAction(tr("&Hide Notes"), this)),
       m_quitAction(new QAction(tr("&Quit"), this)),
       m_trayIconMenu(new QMenu(this)),
+#endif
       m_listView(Q_NULLPTR),
       m_listModel(Q_NULLPTR),
       m_listViewLogic(Q_NULLPTR),
@@ -191,9 +193,13 @@ void MainWindow::setMainWindowVisibility(bool state)
         show();
         raise();
         activateWindow();
+#if !defined(Q_OS_MAC)
         m_restoreAction->setText(tr("&Hide Notes"));
+#endif
     } else {
+#if !defined(Q_OS_MAC)
         m_restoreAction->setText(tr("&Show Notes"));
+#endif
         hide();
     }
 }
@@ -432,13 +438,18 @@ void MainWindow::setupFonts()
  */
 void MainWindow::setupTrayIcon()
 {
+#if !defined(Q_OS_MAC)
     m_trayIconMenu->addAction(m_restoreAction);
     m_trayIconMenu->addSeparator();
     m_trayIconMenu->addAction(m_quitAction);
+#endif
 
     QIcon icon(QStringLiteral(":images/notes_system_tray_icon.png"));
     m_trayIcon->setIcon(icon);
+
+#if !defined(Q_OS_MAC)
     m_trayIcon->setContextMenu(m_trayIconMenu);
+#endif
     m_trayIcon->show();
 }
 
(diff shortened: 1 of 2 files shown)


zjeffer (approved):
Tested on Linux with bspwm, using polybar's systray module. Works perfectly!

guihkx:
Nice, thanks for testing!

I'll wait for the resolution of #452.

nuttyartist:
I'll test on my Mac now and report back.

nuttyartist:
Oh just saw what you wrote about Mac. I think this feature should be included. We can set the context menu to show only when right-clicking. What about other platforms? When a user right-clicks that, does it still shows the options? (we mainly need to give the option to quit, I believe if we do not quit via closing the app)

guihkx:
>We can set the context menu to show only when right-clicking.

Cool, I didn't know that was a possibility. I'll try to implement that.

>What about other platforms? When a user right-clicks that, does it still shows the options?

Yes. On both Windows and Linux, the context menu is still shown.

On Linux, however, it depends on the implementation. I only tested this on the two most popular desktop environments at the moment: GNOME Shell and KDE Plasma.

KDE Plasma behaves exactly like Windows does.

GNOME Shell doesn't have a system tray implementation anymore, so that's done by a third-party GNOME Shell extension ([this being the most popular one](https://github.com/ubuntu/gnome-shell-extension-appindicator)). And on this particular implementation, the context menu is shown after either a single left click, or a single right click. This activation I'm implementing here is triggered by a double left click.

nuttyartist:
> > We can set the context menu to show only when right-clicking.
> 
> Cool, I didn't know that was a possibility. I'll try to implement that.
> 
> > What about other platforms? When a user right-clicks that, does it still shows the options?
> 
> Yes. On both Windows and Linux, the context menu is still shown.
> 
> On Linux, however, it depends on the implementation. I only tested this on the two most popular desktop environments at the moment: GNOME Shell and KDE Plasma.
> 
> KDE Plasma behaves exactly like Windows does.
> 
> GNOME Shell doesn't have a system tray implementation anymore, so that's done by a third-party GNOME Shell extension ([this being the most popular one](https://github.com/ubuntu/gnome-shell-extension-appindicator)). And on this particular implementation, the context menu is shown after either a single left click, or a single right click. This activation I'm implementing here is triggered by a double left click.

Perfect. 

> Cool, I didn't know that was a possibility. I'll try to implement that.

Well, I think it's a possibility haha. Let me know how it goes and I'll test.

guihkx:
>We can set the context menu to show only when right-clicking.

I couldn't find any Qt API that allows us to do that... :(

src/mainwindow.h
```diff
@@ -241,6 +241,7 @@ public slots:
 private slots:
     void InitData();
 
+    void onSystemTrayIconActivated(const QSystemTrayIcon::ActivationReason reason);
```
bjorn, line 244: Note that putting `const` on a by-value parameter is meaningless.

src/mainwindow.cpp
```diff
@@ -181,6 +181,21 @@ void MainWindow::InitData()
...
 }
 
+/*!
+ * \brief Toggles visibility of the main window upon system tray activation
+ * \param reason The reason the system tray was activated
+ */
+void MainWindow::onSystemTrayIconActivated(const QSystemTrayIcon::ActivationReason reason)
+{
+    // We disable this functionality altogether on macOS because a single left/right
+    // mouse click on the tray icon displays its context menu, which is a bit awkward.
```
bjorn, line 191: I noticed there is also `QSystemTrayIcon::MiddleClick`, maybe that could be an alternative for macOS?

Alternatively, since the menu only has "show" and "quit" actions anyway, we can remove the menu entirely on macOS and just keep the toggle behavior. Quitting on macOS anyway has its own action in the application menu and by Command-Q.

nuttyartist: Yes, I guess we don't really need the context menu on macOS when keeping the toggle behavior.

nuttyartist: Let me know when I can test this.

guihkx:
I haven't pushed my local changes yet to address your comments, but I've encountered one annoying bug on macOS while implementing this (but I could also reproduce in v2.0.0, so it isn't new):

If we "close" Notes using the red button, the window actually just hides to the dock, I think -- and that's fine.

However, if we now click on the tray icon to restore the window, Notes still thinks the window is visible instead of hidden, therefore needing two mouse clicks instead of just one to restore the app:

https://user-images.githubusercontent.com/626206/214741606-ec0ce560-194e-4250-8188-c4dbe2ff0cdc.mp4

I'm not really sure how to fix this, because even with the window hidden, [`isVisible()`](https://doc.qt.io/qt-5/qwidget.html#visible-prop) still returns `true`.

I can either open a new bug about this and merge this PR as-is, or I can wait for someone more experienced to give me pointers to fix this.

nuttyartist:
Looks good to me! I'll test it now. I think we can merge this and open a new issue for the problem you mentioned. Do we need native code to detect wether the app is hidden or not?

nuttyartist:
Works great on my computer (apart from the mentioned problem).

guihkx:
Thanks for testing!

I'll merge this and then I'll open an issue about the aforementioned bug.

PR #455, opened 2023-01-23
Windows: Seamless switch between window decoration modes

guihkx:
This fixes switching between native and non-native window decorations on Windows, and therefore we also remove the previous confirmation prompt that would tell users they would need to restart their system after the switch.

Before:

https://user-images.githubusercontent.com/626206/213961948-31393f6b-4247-480b-aa95-c04c45c0ce14.mp4

After:

https://user-images.githubusercontent.com/626206/213962037-4ab78056-57ae-4355-b99c-aecc40107c18.mp4

Tested on:

- Windows 7
  - [x] Qt 5
- Windows 10
  - [x] Qt 5
  - [x] Qt 6
- Windows 11
  - [ ] Qt 5
  - [ ] Qt 6

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -1691,7 +1691,7 @@ void MainWindow::onDotsButtonClicked()
     useNativeFrameAction->setCheckable(true);
     useNativeFrameAction->setChecked(m_useNativeWindowFrame);
     connect(useNativeFrameAction, &QAction::triggered, this,
-            &MainWindow::askBeforeSettingNativeWindowFrame);
+            [this]() { setUseNativeWindowFrame(!m_useNativeWindowFrame); });
 #endif
 
     mainMenu.exec(m_dotsButton->mapToGlobal(QPoint(0, m_dotsButton->height())));
@@ -3538,28 +3538,6 @@ void MainWindow::stayOnTop(bool checked)
 }
 
 /*!
- * \brief MainWindow::askBeforeSettingNativeWindowFrame
- */
-void MainWindow::askBeforeSettingNativeWindowFrame()
-{
-#ifdef _WIN32
-    if (!m_useNativeWindowFrame) {
-        QMessageBox msgBox;
-        msgBox.setText(tr("Warning: After performing this action, you will need to restart your "
-                          "system to revert to custom decoration."));
-        msgBox.setInformativeText(tr("Would you like to continue?"));
-        msgBox.setStandardButtons(QMessageBox::Yes | QMessageBox::No);
-        msgBox.setDefaultButton(QMessageBox::No);
-        if (msgBox.exec() != QMessageBox::Yes) {
-            return;
-        }
-    }
-#endif
-
-    setUseNativeWindowFrame(!m_useNativeWindowFrame);
-}
-
-/*!
  * \brief MainWindow::increaseHeading
  * Increase markdown heading level
  */
@@ -3655,22 +3633,11 @@ void MainWindow::setUseNativeWindowFrame(bool useNativeWindowFrame)
     m_redCloseButton->setVisible(!useNativeWindowFrame);
     m_yellowMinimizeButton->setVisible(!useNativeWindowFrame);
 
-    auto flags = windowFlags();
+    // Reset window flags to its initial state.
+    Qt::WindowFlags flags = Qt::Window;
 
-#  if defined(Q_OS_LINUX) || defined(Q_OS_FREEBSD)
-    if (useNativeWindowFrame)
-        flags &= ~Qt::FramelessWindowHint;
-    else
-        flags |= Qt::FramelessWindowHint;
-#  elif _WIN32
-    if (useNativeWindowFrame) {
-        flags &= ~Qt::CustomizeWindowHint;
-        flags &= ~Qt::FramelessWindowHint;
-    } else {
+    if (!useNativeWindowFrame)
         flags |= Qt::CustomizeWindowHint;
-        flags |= Qt::FramelessWindowHint;
-    }
-#  endif
 
     setWindowFlags(flags);
 #endif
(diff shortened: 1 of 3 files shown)


src/mainwindow.cpp
```diff
@@ -3633,14 +3611,10 @@ void MainWindow::setUseNativeWindowFrame(bool useNativeWindowFrame)
...
-#  elif _WIN32
-    if (useNativeWindowFrame) {
-        flags &= ~Qt::CustomizeWindowHint;
-        flags &= ~Qt::FramelessWindowHint;
-    } else {
+#  elif defined(Q_OS_WINDOWS)
+    flags = Qt::Window;
+    if (!useNativeWindowFrame)
         flags |= Qt::CustomizeWindowHint;
-        flags |= Qt::FramelessWindowHint;
```
bjorn, line 3642: Hmm, this basically takes it back to the initial implementation I did back in e013bb14d873e87cbdd9278a4ff18f54300f50f0. For some reason `FramelessWindowHint` flag was added in 15fa87080fc4f450f512b78983f4a21a6579c66c, but I have no idea what "window ui problem" that was meant to address and whether this flag was needed for something (that flag was already removed on Windows a long time ago in 5831d787d385339cf0fe16104211ed894bb260dc).

Patch looks fine, but I would prefer to use a similar method as for the Linux case to set/unset the flag, to keep the possibility open that we or Qt might use other unrelated flags.

Also, you may want to adjust `MainWindow::setupMainWindow` to also include `Qt::Window` flag for consistency (or use `setWindowFlags(windowFlags() | Qt::CustomizeWindowHint)` instead.

Btw, the `QProgressDialog` in `MainWindow::InitData` uses `Qt::FramelessWindowHint`. Should it also be changed to `Qt::CustomizeWindowHint`?

guihkx: >Patch looks fine, but I would prefer to use a similar method as for the Linux case to set/unset the flag, to keep the possibility open that we or Qt might use other unrelated flags.

Will do!

> Also, you may want to adjust `MainWindow::setupMainWindow` to also include `Qt::Window` flag for consistency (or use `setWindowFlags(windowFlags() | Qt::CustomizeWindowHint)` instead.

Right, good catch. 

> Btw, the `QProgressDialog` in `MainWindow::InitData` uses `Qt::FramelessWindowHint`. Should it also be changed to `Qt::CustomizeWindowHint`?

I think so, yes. At least for consistency. And the updater window as well...

guihkx:
I'm wondering why we are only using `Qt::CustomizeWindowHint` on Windows, however. The documentation [states](https://doc.qt.io/qt-5/qt.html#:~:text=The%20CustomizeWindowHint%20flag):

>The `CustomizeWindowHint` flag is used to enable customization of the window controls. This flag must be set to allow the `WindowTitleHint`, `WindowSystemMenuHint`, `WindowMinimizeButtonHint`, `WindowMaximizeButtonHint` and `WindowCloseButtonHint` flags to be changed.

Which are things that we're also doing on Linux while in 'frameless mode', aren't we? I mean, it seems to work just fine without `Qt::CustomizeWindowHint`, but maybe we should just include it?

nuttyartist:
> I'm wondering why we are only using Qt::CustomizeWindowHint on Windows, however. The documentation [states](https://doc.qt.io/qt-5/qt.html#:~:text=The%20CustomizeWindowHint%20flag):

If I remember correctly, it was related to the way the window is showing. I think the `CustomizeWindowHint` made the window look better (either with shadow or without I can't recall) on Windows 8.

guihkx:
Sorry, my phrasing was ambiguous.

Yes, I did notice that using `Qt::FramelessWindowHint` on Windows makes windows have no shadows.

What I actually meant, though, is that we should probably be using `Qt::CustomizeWindowHint` *not only* on Windows, but on other platforms as well (according to my interpretation of the docs).

Anyway, this subject is not too relevant for this PR in particular...

I haven't addressed @bjorn's comments yet, but I'll try to do it tomorrow 'cause it's a bit late for me now...

nuttyartist:
Great, let me know when it's ready for testing.

bjorn:
> What I actually meant, though, is that we should probably be using `Qt::CustomizeWindowHint` _not only_ on Windows, but on other platforms as well (according to my interpretation of the docs).

Yeah, would be nice to have it consistent if the behavior is the same anyway. I tested both flags on Xfce and see no difference at all. They both cause the native frame to no longer be rendered, and in both cases the window still has shadow.

guihkx:
Okay, I went through every place we call `setWindowFlags()`, and I think we now set them consistently:

On Windows, while in 'frameless mode', we set these window flags: `Qt::Window | Qt::CustomizeWindowHint`.

On any other OSes, we use these instead: `Qt::Window | Qt::CustomizeWindowHint | Qt::FramelessWindowHint`.

To recap the reasons:

- We don't use `Qt::FramelessWindowHint` on Windows specifically, because it makes windows shadowless.
- We now use `Qt::CustomizeWindowHint` on every platform, because that's what Qt docs recommend when we intend to modify window controls.

... I think that's it. Let me know how it goes for you guys.

Thanks!

bjorn:
What is the reason to keep `Qt::FramelessWindowHint`?

guihkx:
No particular reason, I haven't actually tested without it...

Does it still work the same?

bjorn:
> Does it still work the same?

I only tested it on Xfce, like I said previously, I could notice no difference. I would say it's worth using only `Qt::CustomizeWindowHint` and just testing it.

guihkx:
Oh, right... I'll take a look into that.

guihkx:
I tested on GNOME 42 and macOS Big Sur; and I see no difference as well, so I'm dropping `Qt::FramelessWindowHint` in favor of just using `Qt::CustomizeWindowHint` globally (with the exception of `setupMainWindow()` on macOS, because that breaks decorations).

bjorn (approved):

nuttyartist:
Tested and works very well, thanks @guihkx! Ready to merge on my end.

PR #460, opened 2023-02-07
Ignore some events when using native window decorations

guihkx:
Currently, these events are only being used to handle window management (moving, resizing, etc.) when we're in 'frameless mode'.

So, they can be ignored when the native window frame is used.

This fixes some mildly annoying bugs present when we're using native window decorations, such as:

- Being able to maximize/unmaximize the window by double clicking next to its title bar
- Being able to move or resize the window by grabbing next to its edges

Videos of these bugs are [down this thread](https://github.com/nuttyartist/notes/pull/460#issuecomment-1420593061).

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -2526,6 +2526,11 @@ void MainWindow::closeEvent(QCloseEvent *event)
  */
 void MainWindow::mousePressEvent(QMouseEvent *event)
 {
+    if (m_useNativeWindowFrame) {
+        MainWindowBase::mousePressEvent(event);
+        return;
+    }
+
     m_mousePressX = event->pos().x();
     m_mousePressY = event->pos().y();
 
@@ -2578,7 +2583,7 @@ void MainWindow::mousePressEvent(QMouseEvent *event)
         event->accept();
 
     } else {
-        QMainWindow::mousePressEvent(event);
+        MainWindowBase::mousePressEvent(event);
     }
 }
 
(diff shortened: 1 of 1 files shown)


bjorn:
Are you fixing any particular bug with these checks? It seems to me like most of them are superfluous. Also, since these are overridden functions, in general we'll want to call the superclass methods rather than just eating the event (the superclass will generally call `ignore()` on the event).

Note that, for convenience, events are accepted by default. I think basically all `event->accept();` calls are superfluous (exception would be, if we wanted to accept an event after calling the superclass, since it would call ignore on the event).

bjorn:
> Yes, but nothing really serious. Just some mild annoyances:

Ah, those are pretty good to get rid of indeed. I'd suggest listing them in the commit message.

> Would that be something like this? (using `MainWindow::mouseDoubleClickEvent` as an example)

Exactly, though you'd want to write `MainWindowBase::mouseDoubleClickEvent(event);` instead, which will likely resolve to the same function, but would not skip overrides for classes in between `MainWindow` and `QWidget`..

guihkx:
Okay, thank you! I think I understand how it works now.

However, I read this from the [docs](https://doc.qt.io/qt-5/qevent.html#details) (emphasis mine):

>The [accept](https://doc.qt.io/qt-5/qevent.html#accept) flag set with accept(), and cleared with [ignore](https://doc.qt.io/qt-5/qevent.html#ignore)(). It is set by default, **but don't rely on this as subclasses may choose to clear it in their constructor**.

I'm not really sure if that's not happening elsewhere in the codebase (once again, newbie), so should I really get rid of all `event->accept()` calls unconditionally? 😬

bjorn:
> So I'm not really sure if that's not happening elsewhere in the codebase (once again, newbie), so should I really get rid of all `event->accept()` calls unconditionally? I'd rather not touch those... grimacing

Hmm, then just leave them in I guess. I'm not sure since when that doc warning is there nor which event subclasses that would be referring to. I just remember that accepted was chosen as the default for convenience, so that you would not need to call `accept()` all the time. :man_shrugging:

bjorn (approved):

nuttyartist:
Good, it should create a more native experience like people expect to have. Merging, thanks!

PR #462, opened 2023-02-07
Cl: Use create-dmg for better looking dmgs on macOS

guihkx:
#### create-dmg: https://github.com/sindresorhus/create-dmg

It also makes the installation process much easier than just a regular dmg created with `hdiutil`.

I had to add some silly logic to deal with `create-dmg` exiting with code `2` after failing to code sign the app (as we don't do that yet).

So either exit code `0` or `2` means the dmg was created successfully.

---

I believe @nuttyartist mentioned in the past the he was already using `create-dmg` before the AppVeyor -> Github Actions migration, so this shouldn't be news to him. :)

Nevertheless, for the sake of completion, here are some before and after pictures:

Before:

image

After:

with-create-dmg

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -163,10 +163,25 @@ jobs:
               rm -rv Contents/PlugIns/bearer
           fi
 
+      - name: Install create-dmg and dependencies
+        run: |
+          brew install graphicsmagick imagemagick
+          npm install --global create-dmg@^6.0.0
+
       - name: Build dmg (${{ matrix.build-type }})
         run: |
           cd build
-          hdiutil create -srcfolder 'Notes Better.app' -format UDZO -fs HFS+ -volname 'Notes Better' '${{ steps.vars.outputs.file_name }}'
+          # TODO: Remove most of the code below once we support signing/notarization.
+          set +e
+          create-dmg 'Notes Better.app'
+          code=$?
+          set -e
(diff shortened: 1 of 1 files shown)


.github/workflows/macos.yml
```diff
@@ -166,7 +166,18 @@ jobs:
       - name: Build dmg (${{ matrix.build-type }})
         run: |
           cd build
-          hdiutil create -srcfolder 'Notes Better.app' -format UDZO -fs HFS+ -volname 'Notes Better' '${{ steps.vars.outputs.file_name }}'
+          npm install --global create-dmg
```
guihkx, line 169: I should probably set a version here just to avoid future problems, right?

`6.0.0` is the latest one: https://www.npmjs.com/package/create-dmg?activeTab=versions

bjorn: Can we just specify the major version? It may still be good to pick up bugfixes and compatible improvements.

nuttyartist:
Nice. Can you make the app icon show up on the dmg? I forgot I did it but it should say here https://github.com/sindresorhus/create-dmg#dmg-icon

.github/workflows/macos.yml
guihkx: Good thinking. Done.

guihkx:
dmg icon added (before/after):

dmg-icon

Thank you both!

nuttyartist:
Awesome, thanks!

PR #463, opened 2023-02-08
CI: Don't ship non-SQLite Qt drivers

guihkx:
We only use the SQLite driver, so we remove other drivers to reduce the final size of the app.

On macOS, this is accomplished automatically, thanks to the `-appstore-compliant` flag from `macdeployqt`:

```shell
$ macdeployqt 'Notes Better.app' -appstore-compliant
WARNING: Plugin "libqsqlodbc.dylib" uses private API and is not Mac App store compliant.
WARNING: Skip plugin "libqsqlodbc.dylib"
WARNING: Plugin "libqsqlpsql.dylib" uses private API and is not Mac App store compliant.
WARNING: Skip plugin "libqsqlpsql.dylib"
```

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -238,6 +238,15 @@ jobs:
               # The bearer plugin has caused problems for us in the past. Plus, it was removed altogether in Qt 6.
               rm -rv usr/plugins/bearer
           fi
+          # We only use the SQLite Qt driver, so it's fine to delete others.
+          rm -v usr/plugins/sqldrivers/libqsqlodbc.so
+          rm -v usr/plugins/sqldrivers/libqsqlpsql.so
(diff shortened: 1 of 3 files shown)


bjorn (approved):

nuttyartist:
Oh, so our app can technically be uploaded to the App Store? I always used to see those warnings and think it's not possible.

BTW, so these libraries have a `sql` in them but are not used by our app?

bjorn:
> BTW, so these libraries have a `sql` in them but are not used by our app?

Yes, SQL is a query language used by many databases (usually with their own dialects). Notes only uses the SQLite database and as such only needs the plugin for that particular database.

nuttyartist:
Okay, merging...

PR #469, opened 2023-02-08
Fixed window flags of popup windows

zjeffer:
This has been annoying me for a while: on tiling window managers, the popup windows (Style editor, Updater, AboutWindow) have the wrong window flags, causing them to be seen as separate windows instead of popups.

This is before:

https://user-images.githubusercontent.com/4633209/217641899-e408d5e3-2350-4162-b2c5-1af28d6863d8.mp4

This is after:

https://user-images.githubusercontent.com/4633209/217642072-02f91f59-66ca-476d-8480-97dfae7f3c17.mp4

Should also be tested with non-tiling window managers, and on macOS and Windows.

### DIFF
--- a/src/aboutwindow.cpp
+++ b/src/aboutwindow.cpp
@@ -6,10 +6,9 @@
 /**
  * Initializes the window components and configures the AboutWindow
  */
-AboutWindow::AboutWindow(QWidget *parent) : QWidget(parent), m_ui(new Ui::AboutWindow)
+AboutWindow::AboutWindow(QWidget *parent) : QDialog(parent), m_ui(new Ui::AboutWindow)
 {
     m_ui->setupUi(this);
-    this->setWindowFlags(Qt::Window);
     setWindowTitle(tr("About") + " " + qApp->applicationName());
 
     m_ui->aboutText->setText(
--- a/src/aboutwindow.h
+++ b/src/aboutwindow.h
@@ -1,13 +1,13 @@
 #ifndef ABOUTWINDOW_H
 #define ABOUTWINDOW_H
 
-#include <QWidget>
+#include <QDialog>
 
 namespace Ui {
 class AboutWindow;
 }
 
-class AboutWindow : public QWidget
+class AboutWindow : public QDialog
 {
     Q_OBJECT
 
--- a/src/styleeditorwindow.cpp
+++ b/src/styleeditorwindow.cpp
@@ -9,7 +9,7 @@
  * Initializes the window components and configures the StyleEditorWindow
  */
 StyleEditorWindow::StyleEditorWindow(QWidget *parent)
-    : QWidget(parent),
+    : QDialog(parent),
       m_ui(new Ui::StyleEditorWindow),
       m_currentlyClickedButton(Q_NULLPTR),
       m_currentSelectedFontButton(Q_NULLPTR),
@@ -17,8 +17,7 @@ StyleEditorWindow::StyleEditorWindow(QWidget *parent)
       m_isFullWidthClicked(false)
 {
     m_ui->setupUi(this);
-    this->setWindowTitle(tr("Editor Settings"));
-    this->setWindowFlags(Qt::Window | Qt::WindowStaysOnTopHint);
+    setWindowTitle(tr("Editor Settings"));
 
     m_ui->serifButton->setToolTip(
             "Change the text editor font to a serif font.\nClick again to try different fonts");
--- a/src/styleeditorwindow.cpp
+++ b/src/styleeditorwindow.cpp
@@ -9,7 +9,7 @@
  * Initializes the window components and configures the StyleEditorWindow
  */
 StyleEditorWindow::StyleEditorWindow(QWidget *parent)
-    : QDialog(parent),
+    : QDialog(parent, Qt::Tool),
       m_ui(new Ui::StyleEditorWindow),
       m_currentlyClickedButton(Q_NULLPTR),
       m_currentSelectedFontButton(Q_NULLPTR),
--- a/src/updaterwindow.cpp
+++ b/src/updaterwindow.cpp
@@ -48,7 +48,7 @@ static const QDir DOWNLOAD_DIR(QDir::homePath() + "/Downloads/");
  * Initializes the window components and configures the QSimpleUpdater
  */
 UpdaterWindow::UpdaterWindow(QWidget *parent)
-    : QWidget(parent),
+    : QDialog(parent),
       m_fileName(""),
       m_ui(new Ui::UpdaterWindow),
       m_canMoveWindow(false),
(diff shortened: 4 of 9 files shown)


src/updaterwindow.cpp
```diff
@@ -100,7 +100,7 @@ UpdaterWindow::UpdaterWindow(QWidget *parent)
     updateTitleLabel();
 
     /* Remove window border */
-    setWindowFlags(Qt::Window | Qt::CustomizeWindowHint);
```
bjorn, line 103: Would it make sense to subclass `UpdaterWindow` from `QDialog` in the first place?

src/aboutwindow.cpp
```diff
@@ -9,7 +9,7 @@
 AboutWindow::AboutWindow(QWidget *parent) : QWidget(parent), m_ui(new Ui::AboutWindow)
 {
     m_ui->setupUi(this);
-    this->setWindowFlags(Qt::Window);
+    this->setWindowFlags(Qt::SplashScreen | Qt::WindowStaysOnTopHint);
```
bjorn, line 12: It seems to me that the `AboutWindow` should really derive from `QDialog` instead. I don't see the point of `Qt::WindowStaysOnTopHint` and `Qt::SplashScreen` (which is just the same as `ToolTip | Dialog`) seems a little out of place (on Xfce, this creates a window without frame that can't be closed).

Btw, while you're changing this anyway, please remove the superfluous `this->`. :-)

src/styleeditorwindow.cpp
```diff
@@ -18,7 +18,7 @@ StyleEditorWindow::StyleEditorWindow(QWidget *parent)
 {
     m_ui->setupUi(this);
     this->setWindowTitle(tr("Editor Settings"));
-    this->setWindowFlags(Qt::Window | Qt::WindowStaysOnTopHint);
+    this->setWindowFlags(Qt::Tool | Qt::WindowStaysOnTopHint);
```
bjorn, line 21: The existing `Qt::WindowStaysOnTopHint` seems wrong. On Xfce, this results in a window that will stay on top by itself, even if another application is brought to the foreground. We really only want the `StyleEditorWindow` to stay on top of the main Notes window (which is achieved by `Qt::Tool` btw, so that change seems appropriate).

zjeffer: On bspwm, only having `Qt:Tool` will make the style editor window appear behind the main window, for some reason :thinking:

zjeffer: `Qt::Dialog` makes it behave as expected.

zjeffer:
What if we just make all three windows subclasses of QDialog instead of QWidget, and make the UpdaterWindow also have CustomizedWindowHint? @bjorn

src/updaterwindow.cpp
```diff
@@ -100,7 +100,7 @@ UpdaterWindow::UpdaterWindow(QWidget *parent)
...
+    setWindowFlags(Qt::CustomizeWindowHint);
```
bjorn, line 103: I think this may unset some flags, so instead it should probably be:
```suggestion
    setWindowFlags(windowFlags() | Qt::CustomizeWindowHint);
```

src/styleeditorwindow.cpp
```diff
@@ -9,16 +9,15 @@
...
 StyleEditorWindow::StyleEditorWindow(QWidget *parent)
-    : QWidget(parent),
+    : QDialog(parent),
       m_ui(new Ui::StyleEditorWindow),
       m_currentlyClickedButton(Q_NULLPTR),
       m_currentSelectedFontButton(Q_NULLPTR),
       m_currentSelectedThemeButton(Q_NULLPTR),
       m_isFullWidthClicked(false)
 {
     m_ui->setupUi(this);
```
bjorn, line 19: We may also want to change `class="QWidget"` to `class="QDialog"` in `styleeditorwindow.ui`. I think that change can only be made using a text editor, though. Same for the other two classes.

src/aboutwindow.h
```diff
@@ -2,12 +2,13 @@
 #define ABOUTWINDOW_H
 
 #include <QWidget>
+#include <QDialog>
```
bjorn, line 5: The QDialog include can replace the QWidget include.

bjorn:
Sorry for all the separate reviews, next time I'll try to finish reviewing first. :-)

In any case, I've tested the latest version on GNOME 43 and all windows seem to work fine there.

nuttyartist:
I want to test this but `git clone --recursive --single-branch --branch windowflags https://github.com/zjeffer/notes.git` doesn't work. What should I use?

bjorn:
> I want to test this but `git clone --recursive --single-branch --branch windowflags https://github.com/zjeffer/notes.git` doesn't work. What should I use?

That doesn't work because the branch isn't `windowflags`, but `fix/zjeffer/windowflags`. Of course, you could also test using the binaries compiled by GitHub Actions. :-)

bjorn:
> Also, I created that branch on this repo, not my own fork.

Ah, hadn't noticed that. I personally don't think it matters where we create the branches.

Btw, I can really recommend installing [`gh`](https://cli.github.com/) and just using `gh pr checkout 469` to checkout this pull request. Avoids needing to get all those options right, and automatically creates a local branch tracking the remote one, so you can also easily contribute changes to somebody else's PR.

nuttyartist:
@zjeffer I don't have a preference on this. 

> Of course, you could also test using the binaries compiled by GitHub Actions

You're right.

I tested this, but I think it's not ideal that the editor settings keep fading to the background. For a good UX it should float above the main window so users can easily keep seeing their changes as they affect the settings. Ideal situation is where the editor settings keep floating on top of the main window but not of other apps.

bjorn:
> I tested this, but I think it's not ideal that the editor settings keep fading to the background. For a good UX it should float above the main window so users can easily keep seeing their changes as they affect the settings. Ideal situation is where the editor settings keep floating on top of the main window but not of other apps.

The desired behavior is exactly what I am seeing in GNOME. Can you provide a screenshot, showing what you mean with the editor settings fading to the background?

nuttyartist:
Here in the old version i can continue using the main window without the editor settings fading into the background:

editor-settings

With this PR I can't:

editor-settings2

bjorn:
@nuttyartist Hmm, this seems like pretty much a bug on macOS, cause I think the dialog stays on top of the main Notes window on Linux and Windows (and I'm seeing the same annoying behavior with the Preferences dialog in Tiled, which is also not modal).

If this really only affects macOS, I would suggest we keep the `Qt::WindowStaysOnTopHint` only on that system. It's not perfect, because I guess that will cause this window to be even on top of other applications? Alternatively, we can test the behavior of `Qt::Tool` on macOS.

bjorn:
I've pushed two small changes to the style editor window and the about window to improve the behavior on macOS. I hope this doesn't make things worse on other platforms.

nuttyartist:
I would like to test it but I flew for a short trip and forgot to bring the charger for my laptop (which is out of power now) 🥲

I will either try to get my hands on another computer or wait till Tuesday.

zjeffer:
I tested the latest build, works perfectly on bspwm.

guihkx (approved):

zjeffer:
Ok, thanks for testing. Merging...

nuttyartist:
Yes, works for me as well now.

Issue #290, opened 2021-06-30
Feature e


goldencut:

nuttyartist:
Fair enough! Adding to our to-do list!

Thanks!

PR #470, opened 2023-02-09
Fix closing Notes when hidden, added hiding to tray option. Fixes #278 & #290

### LINKED ISSUE #278: Clicking on native close button quits the app on Windows rather than hiding it
Clicking on the red closing button on Windows quits the app, where it should just hide it (should keep on running using the system tray). It seems like the problem is upstream due to a newer version of Qt since I didn't find a change in the source code to lead to this and version 1.0.0 on the website works as expected. But we might solve this using a different code?

### LINKED ISSUE #461: Can't quit the app when it's minimized to the system tray
This only affects the Qt 6 build for some reason, but I could reproduce the problem on both Linux and Windows:

1. Close Notes to system tray
2. Right click the system tray icon
3. Choose 'Quit'

**Expected result:**

The app should quit, like the Qt 5 build does.

**Current result:**

The app does not quit; you have to restore it from the system tray and then quit.

guihkx:
After porting Notes to Qt 6, the "Quit" system tray menu action wouldn't work if Notes was minimized to the system tray.

It might be a Qt regression, but since this approach works on both Qt 5 and 6, let's use it instead.

Rationale: https://github.com/nuttyartist/notes/issues/461#issuecomment-1423622933

Fixes #461

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -85,6 +85,7 @@ MainWindow::MainWindow(QWidget *parent)
 #endif
       m_alwaysStayOnTop(false),
       m_useNativeWindowFrame(false),
+      m_hideToTray(false),
       m_listOfSerifFonts(
               { QStringLiteral("Trykker"), QStringLiteral("PT Serif"), QStringLiteral("Mate") }),
       m_listOfSansSerifFonts({ QStringLiteral("Source Sans Pro"), QStringLiteral("Roboto") }),
@@ -95,10 +96,8 @@ MainWindow::MainWindow(QWidget *parent)
       m_chosenMonoFontIndex(0),
       m_isNoteListCollapsed(false),
       m_isTreeCollapsed(false),
-      m_currentCharsLimitPerFont({ 64 // Mono    TODO: is this the proper way to initialize?
-                                   ,
-                                   80 // Serif
-                                   ,
+      m_currentCharsLimitPerFont({ 64, // Mono    TODO: is this the proper way to initialize?
+                                   80, // Serif
                                    80 }), // SansSerif
       m_currentFontTypeface(FontTypeface::SansSerif),
 #ifdef __APPLE__
@@ -122,7 +121,6 @@ MainWindow::MainWindow(QWidget *parent)
     ui->setupUi(this);
     setupMainWindow();
     setupFonts();
-    setupTrayIcon();
     setupKeyboardShortcuts();
     setupNewNoteButtonAndTrahButton();
     setupSplitter();
@@ -1237,6 +1235,11 @@ void MainWindow::restoreStates()
     setUseNativeWindowFrame(
             m_settingsDatabase->value(QStringLiteral("useNativeWindowFrame"), false).toBool());
 
+    setHideToTray(m_settingsDatabase->value(QStringLiteral("hideToTray"), true).toBool());
+    if (m_hideToTray) {
+        setupTrayIcon();
+    }
+
     if (m_settingsDatabase->value(QStringLiteral("windowGeometry"), "NULL") != "NULL")
         this->restoreGeometry(
                 m_settingsDatabase->value(QStringLiteral("windowGeometry")).toByteArray());
@@ -1629,6 +1632,22 @@ void MainWindow::onDotsButtonClicked()
             [=]() { m_autostart.setAutostart(autostartAction->isChecked()); });
     autostartAction->setCheckable(true);
     autostartAction->setChecked(m_autostart.isAutostart());
+
+    // hide to tray
+    QAction *hideToTrayAction = mainMenu.addAction(tr("&Hide to tray"));
+    connect(hideToTrayAction, &QAction::triggered, this, [=]() {
+        m_settingsDatabase->setValue(QStringLiteral("hideToTray"), hideToTrayAction->isChecked());
+    });
+    hideToTrayAction->setCheckable(true);
+    hideToTrayAction->setChecked(m_hideToTray);
+    connect(hideToTrayAction, &QAction::triggered, this, [this]() {
+        setHideToTray(!m_hideToTray);
+        if (m_hideToTray) {
+            setupTrayIcon();
+        } else {
+            m_trayIcon->hide();
+        }
+    });
 
     QAction *changeDBPathAction = mainMenu.addAction(tr("&Change database path"));
     connect(changeDBPathAction, &QAction::triggered, this, [=]() {
@@ -2506,12 +2525,9 @@ void MainWindow::onRedCloseButtonClicked()
     m_redCloseButton->setIcon(QIcon(QStringLiteral(":images/red.png")));
 #endif
 
-    if (m_hideToTray)
-    {
+    if (m_hideToTray) {
         setMainWindowVisibility(false);
-    }
-    else
-    {
+    } else {
         close();
     }
 }
@@ -2528,7 +2544,7 @@ void MainWindow::closeEvent(QCloseEvent *event)
     if (!event->spontaneous() || !isVisible()) {
         return;
     }
-    if (m_trayIcon->isVisible()) {
+    if (m_hideToTray && m_trayIcon->isVisible()) {
         // don't close the application, just hide to tray
         setMainWindowVisibility(false);
         event->ignore();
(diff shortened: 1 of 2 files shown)


src/mainwindow.cpp
```diff
@@ -2189,7 +2189,7 @@ void MainWindow::minimizeWindow()
 void MainWindow::QuitApplication()
 {
     m_noteEditorLogic->closeEditor();
-    MainWindow::close();
+    qApp->quit();
```
bjorn, line 2192: Since this is a static method, I'd suggest:
```suggestion
    QCoreApplication::quit();
```

bjorn:
Please check whether `MainWindow::closeEvent` is still called after this change, since it's doing some saving of settings.

guihkx:
Nice catch, it actually isn't being called... :S

Do you have any suggestion for what approach I should take?

Do you think keeping `MainWindow::close()` and adding `QCoreApplication::quit()` right after is too much of a stupid idea?

bjorn:
> Do you think keeping `MainWindow::close()` and adding `QCoreApplication::quit()` right after is too much of a stupid idea?

Would it break on Qt 5 due to quit potentially getting called twice? Ideally we'd find out why the app isn't quitting... normally that happens once the last window closes.

guihkx:
>Would it break on Qt 5 due to quit potentially getting called twice?

On my test it didn't, but to be honest I'm not a big fan of that approach...

I tried to debug this for a while with gdb, but I didn't get too far. The Qt code base is too complex for me...

The "least worse" workaround I found, is to make the main window visible before closing it. :/

I'm open to further suggestions, or pointers. :)

src/mainwindow.cpp
```diff
@@ -2189,6 +2189,12 @@ void MainWindow::minimizeWindow()
 void MainWindow::QuitApplication()
 {
     m_noteEditorLogic->closeEditor();
+#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
+    // Works around an issue where trying to exit Notes using the
+    // system tray menu action while our window is hidden, fails.
+    if (MainWindow::isHidden())
+        MainWindow::show();
```
bjorn, line 2196: Really interesting that this hack works. It makes it somewhat likely to be a Qt issue, and we should probably report it, but in that case we need to create a minimal example app that reproduces the problem.

bjorn (approved):

src/mainwindow.cpp
zjeffer: I'll look into creating a minimal example.

zjeffer:
While trying to make a minimal working example, I stumbled upon this Qt article: https://doc.qt.io/qt-6/qtwidgets-desktop-systray-example.html. We're doing some things differently:

At the bottom of that article is an example of a `closeEvent` method, which works well with tray icons. I copied their approach. Before, on bspwm, closing the app didn't hide it to tray, but quit the app entirely. Minimizing isn't possible on most tiling managers.

Also, I noticed we're doing our settings saving in closeEvent instead of `MainWindow::QuitApplication`, which didn't make much sense to me, because `closeEvent` also get's called on my system when closing the app using my system shortcuts. 

In the future, we should probably give the user the option to close to tray or not.

zjeffer:
I just tested the Actions' AppImages for Qt5 and Qt6, and they both work flawlessly from my point of view. No hacks needed.

Let me know if this works for you guys too on your systems.

guihkx:
@zjeffer I haven't tested your approach yet, but based on a cursory reading I like it.

It would also fix #290 and #278.

zjeffer:
Also tested the appimages on Ubuntu 22.10, both Qt5 and Qt6 work perfectly.

Needs to be tested on Windows & macOS.

bjorn (approved):
That's a great find!

guihkx:
On Linux, there's the possibility that app will stay running in the background indefinitely after the user closes the window.

The required conditions for that:

1. 'Hide to tray' is unchecked
2. We're running without native decorations (aka 'frameless mode')

Demo:

https://user-images.githubusercontent.com/626206/218202891-3fc7061a-4dab-4000-85f0-78e72dbf735e.mp4

I can test on Windows and macOS later, once the above gets addressed.

P.S.: You may want to get rid of my first commit, which is fairly useless now.

zjeffer:
Thanks for testing, I'll look into this.

> P.S.: You may want to get rid of my first commit, which is fairly useless now.

I already squashed that commit :)

guihkx:
I think the 'Hide to tray' option will have to be entirely removed on macOS (at least for now)...

If we disable that option and close the window, the app won't actually exit, it will only hide to the dock (but they tray icon will be gone):

https://user-images.githubusercontent.com/626206/218228698-06eb7651-5ddb-46a7-bbe9-15b9ccf6ebe4.mp4

That's deliberately done here:

https://github.com/nuttyartist/notes/blob/1142bce844839c6757b2c94f3c7ffa6a92494844/src/framelesswindow.mm#L33-L37

Apparently, CFramelessWindow does provide an API to make the app actually quit when we click the close button, but just by reading its description I immediately think it's a really bad idea... lol

https://github.com/nuttyartist/notes/blob/1142bce844839c6757b2c94f3c7ffa6a92494844/src/framelesswindow.h#L102-L107

src/mainwindow.cpp
```diff
@@ -2446,7 +2524,14 @@ void MainWindow::onRedCloseButtonClicked()
     m_redCloseButton->setIcon(QIcon(QStringLiteral(":images/red.png")));
 #endif
 
-    setMainWindowVisibility(false);
+    if (m_hideToTray)
+    {
```
guihkx, line 2528: nit: styling (we open curly braces on the same line)

guihkx:
Tested on Windows 7 and Windows 11, works great!

zjeffer:
> If we disable that option and close the window, the app won't actually exit, it will only hide to the dock (but they tray icon will be gone)

I'll look into setting up a macOS VM so I can test that behaviour and maybe find a solution before we merge this PR.

nuttyartist:
I'm back.

> Also, I noticed we're doing our settings saving in closeEvent instead of MainWindow::QuitApplication, which didn't make much sense to me, because closeEvent also get's called on my system when closing the app using my system shortcuts.

Well, don't we want to save the settings even in that occasion where users close the app via shortcuts? Or you mean closeEvent is being called when the app is being hidden (to the tray)?

> If we disable that option and close the window, the app won't actually exit, it will only hide to the dock (but they tray icon will be gone):

That's okay. That's normal UX behavior on macOS. Users quit apps usually by right-clicking it on the dock and then Quit.

I like the Hide to tray option but I think it should be under the View category. What do you think?

zjeffer:
> I like the Hide to tray option but I think it should be under the View category. What do you think?

I feel like `View` menu is more for options that change how the app looks, not how it behaves. For instance, the autostart checkbox is also not under view, because it handles how/when the app should open. Similarly, the hide to tray option handles how the app should close. That's my reasoning behind putting them together. 

Although, in the future, we should really be looking into creating a preferences window, in my opinion.

nuttyartist:
> I feel like a `view` menu is more for options that change how the app looks, not how it behaves. For instance, the autostart checkbox is also not under view, because it handles how/when the app should open. Similarly, the hide to tray option handles how the app should close. That's my reasoning behind putting them together.

@zjeffer Very well, then. I agree with you

> Although, in the future, we should really be looking into creating a preferences window, in my opinion.

 Definitely.

nuttyartist (approved):

zjeffer (approved):
I rebased the git branch to fix the merge conflicts. 

I haven't found a way to fix the behaviour on macOS, but if the behaviour is expected, all the better ;)

guihkx:
Thank you, @zjeffer!

Though I still think the hide to tray option should not be added to macOS at all, as it's kind of useless there IMO.

Nevertheless, everything else looks good to me.

zjeffer:
@nuttyartist Ready to release 2.1.0? :smiley:

guihkx:
@zjeffer @nuttyartist Waaaait, don't do that just yet... 😄

I'm working on fixing a couple of very easy bugs 🥲

nuttyartist:
@zjeffer Haha, as soon as @guihkx finishes, I'll start my testing routine.

PR #473, opened 2023-02-13
Build: Enable extra compiler warnings by default

guihkx:
This enables some extra warnings on supported compilers, by default. The options are:

- GCC/Clang: [`-Wall -Wextra -Wpedantic -Wshadow`](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html)
- MSVC: [`/W4`](https://learn.microsoft.com/en-us/cpp/build/reference/compiler-option-warning-level?view=msvc-140#remarks)

For obvious reasons, these flags had to be disabled on third-party sources, that's why I moved them into their own list in `CMakeLists.txt`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -188,8 +188,10 @@ set(SOURCES
     # qrc files
     ${PROJECT_SOURCE_DIR}/src/fonts.qrc
     ${PROJECT_SOURCE_DIR}/src/images.qrc
-    ${PROJECT_SOURCE_DIR}/src/styles.qrc
-    # 3rdParty source files
+    ${PROJECT_SOURCE_DIR}/src/styles.qrc)
+
+# Third-party source files.
+set(SOURCES_3RD_PARTY
     ${PROJECT_SOURCE_DIR}/3rdParty/qautostart/src/qautostart.cpp
     ${PROJECT_SOURCE_DIR}/3rdParty/qautostart/src/qautostart.h
     ${PROJECT_SOURCE_DIR}/3rdParty/qmarkdowntextedit/linenumberarea.h
@@ -213,22 +215,25 @@ set(SOURCES
     ${PROJECT_SOURCE_DIR}/3rdParty/qxt/qxtglobalshortcut.h
     ${PROJECT_SOURCE_DIR}/3rdParty/qxt/qxtglobalshortcut_p.h)
 
-set(MACX_PROJECT_SOURCES
-    ${PROJECT_SOURCE_DIR}/src/framelesswindow.mm
-    ${PROJECT_SOURCE_DIR}/src/framelesswindow.h
-    ${PROJECT_SOURCE_DIR}/src/images/notes_icon.icns
-    ${PROJECT_SOURCE_DIR}/3rdParty/qxt/qxtglobalshortcut_mac.cpp)
+if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
+  list(APPEND SOURCES ${PROJECT_SOURCE_DIR}/src/framelesswindow.mm
+       ${PROJECT_SOURCE_DIR}/src/framelesswindow.h
+       ${PROJECT_SOURCE_DIR}/src/images/notes_icon.icns)
+  list(APPEND SOURCES_3RD_PARTY
+       ${PROJECT_SOURCE_DIR}/3rdParty/qxt/qxtglobalshortcut_mac.cpp)
+elseif(UNIX)
+  list(APPEND SOURCES_3RD_PARTY
+       ${PROJECT_SOURCE_DIR}/3rdParty/qxt/qxtglobalshortcut_x11.cpp)
+elseif(WIN32)
+  list(APPEND SOURCES ${PROJECT_SOURCE_DIR}/src/framelesswindow.cpp
+       ${PROJECT_SOURCE_DIR}/src/framelesswindow.h
+       ${PROJECT_SOURCE_DIR}/src/images/notes.rc)
+  list(APPEND SOURCES_3RD_PARTY
+       ${PROJECT_SOURCE_DIR}/3rdParty/qxt/qxtglobalshortcut_win.cpp)
+endif()
 
-set(WIN32_PROJECT_SOURCES
-    ${PROJECT_SOURCE_DIR}/src/framelesswindow.cpp
-    ${PROJECT_SOURCE_DIR}/src/framelesswindow.h
-    ${PROJECT_SOURCE_DIR}/src/images/notes.rc
-    ${PROJECT_SOURCE_DIR}/3rdParty/qxt/qxtglobalshortcut_win.cpp)
-
-set(LINUX_PROJECT_SOURCES
-    ${PROJECT_SOURCE_DIR}/3rdParty/qxt/qxtglobalshortcut_x11.cpp)
-
-add_executable(${PROJECT_NAME} WIN32 MACOSX_BUNDLE ${SOURCES})
+add_executable(${PROJECT_NAME} WIN32 MACOSX_BUNDLE ${SOURCES}
+                                                   ${SOURCES_3RD_PARTY})
 
 target_include_directories(
   ${PROJECT_NAME}
@@ -274,8 +279,6 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
   set_source_files_properties(${PROJECT_SOURCE_DIR}/src/images/notes_icon.icns
                               PROPERTIES MACOSX_PACKAGE_LOCATION "Resources")
 
-  target_sources(${PROJECT_NAME} PUBLIC ${MACX_PROJECT_SOURCES})
-
   target_link_libraries(${PROJECT_NAME} PUBLIC "-framework Carbon"
                                                "-framework Cocoa")
 elseif(UNIX)
@@ -283,12 +286,8 @@ elseif(UNIX)
   string(TOLOWER ${PROJECT_NAME} BINARY_NAME)
   set_target_properties(${PROJECT_NAME} PROPERTIES OUTPUT_NAME ${BINARY_NAME})
 
-  target_sources(${PROJECT_NAME} PUBLIC ${LINUX_PROJECT_SOURCES})
-
   target_link_libraries(${PROJECT_NAME} PUBLIC X11)
 elseif(WIN32)
-  target_sources(${PROJECT_NAME} PUBLIC ${WIN32_PROJECT_SOURCES})
-
   target_compile_definitions(${PROJECT_NAME} PUBLIC QXT_STATIC)
 
   target_link_libraries(${PROJECT_NAME} PUBLIC user32)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -235,10 +235,12 @@ endif()
 add_executable(${PROJECT_NAME} WIN32 MACOSX_BUNDLE ${SOURCES}
                                                    ${SOURCES_3RD_PARTY})
 
+target_include_directories(${PROJECT_NAME} PUBLIC ${PROJECT_SOURCE_DIR}/src)
+
+# Include third-party headers as 'system' files, in order to silence compiler warnings on them.
 target_include_directories(
-  ${PROJECT_NAME}
-  PUBLIC ${PROJECT_SOURCE_DIR}/src
-         ${PROJECT_SOURCE_DIR}/3rdParty/QSimpleUpdater/include
+  ${PROJECT_NAME} SYSTEM
+  PUBLIC ${PROJECT_SOURCE_DIR}/3rdParty/QSimpleUpdater/include
          ${PROJECT_SOURCE_DIR}/3rdParty/qautostart/src
          ${PROJECT_SOURCE_DIR}/3rdParty/qmarkdowntextedit
          ${PROJECT_SOURCE_DIR}/3rdParty/qxt)
@@ -247,6 +249,17 @@ target_compile_definitions(
   ${PROJECT_NAME}
   PUBLIC APP_VERSION="${PROJECT_VERSION}${GIT_REV}"
          QT_DISABLE_DEPRECATED_BEFORE=0x050900 QSU_INCLUDE_MOC=1)
+
+# Enable useful compiler warnings (except on third-party code).
+if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" OR CMAKE_CXX_COMPILER_ID MATCHES
+                                           "Clang")
+  target_compile_options(${PROJECT_NAME} PUBLIC -Wall -Wextra -Wpedantic
+                                                -Wshadow)
+  set_source_files_properties(${SOURCES_3RD_PARTY} PROPERTIES COMPILE_FLAGS -w)
+elseif(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
+  target_compile_options(${PROJECT_NAME} PUBLIC /W4)
+  set_source_files_properties(${SOURCES_3RD_PARTY} PROPERTIES COMPILE_FLAGS /w)
+endif()
 
 target_link_libraries(
   ${PROJECT_NAME}
--- a/src/customapplicationstyle.cpp
+++ b/src/customapplicationstyle.cpp
@@ -8,25 +8,18 @@ void CustomApplicationStyle::drawPrimitive(PrimitiveElement element, const QStyl
     if (element == QStyle::PE_IndicatorItemViewItemDrop) {
         painter->setRenderHint(QPainter::Antialiasing, true);
 
-        QColor c;
-        //        if (m_theme == Theme::Dark) {
-        //            c = QColor(15, 45, 90);
-        //        } else {
-        c = QColor(207, 207, 207);
-        //        }
-        QPen pen(c);
+        QColor color(207, 207, 207);
+        QPen pen(color);
         pen.setWidth(2);
-        c.setAlpha(50);
-        QBrush brush(c);
-
         painter->setPen(pen);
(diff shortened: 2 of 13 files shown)


CMakeLists.txt
```diff
@@ -248,6 +250,17 @@ target_compile_definitions(
   PUBLIC APP_VERSION="${PROJECT_VERSION}${GIT_REV}"
          QT_DISABLE_DEPRECATED_BEFORE=0x050900 QSU_INCLUDE_MOC=1)
 
+# Enable useful compiler warnings (except on third-party code).
+if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" OR CMAKE_CXX_COMPILER_ID MATCHES
+                                           "Clang")
+  target_compile_options(${PROJECT_NAME} PUBLIC -Wall -Wextra -Wpedantic
```
guihkx, line 256: Assuming this gets approved at all, what do guys think about treating all warnings as errors from now on?

CMakeLists.txt
```diff
@@ -213,27 +215,32 @@ set(SOURCES
...
+if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
+  list(APPEND SOURCES ${PROJECT_SOURCE_DIR}/src/framelesswindow.mm
+       ${PROJECT_SOURCE_DIR}/src/framelesswindow.h
+       ${PROJECT_SOURCE_DIR}/src/images/notes_icon.icns)
+  list(APPEND SOURCES_3RD_PARTY
+       ${PROJECT_SOURCE_DIR}/3rdParty/qxt/qxtglobalshortcut_mac.cpp)
+elseif(UNIX)
+  list(APPEND SOURCES_3RD_PARTY
+       ${PROJECT_SOURCE_DIR}/3rdParty/qxt/qxtglobalshortcut_x11.cpp)
+else()
```
guihkx, line 227: Derp. Note to self: Use `elseif(WIN32)` here.

src/notelistmodel.cpp
```diff
@@ -384,13 +384,11 @@ bool NoteListModel::dropMimeData(const QMimeData *mime, Qt::DropAction action, i
...
-        if (row < m_pinnedList.size()) {
-            return m_pinnedList[row];
-        } else {
-            row = row - m_pinnedList.size();
-            return m_noteList[row];
+    auto getRef = [this](int rowIndex) -> NodeData & {
+        if (rowIndex < m_pinnedList.size()) {
+            return m_pinnedList[rowIndex];
         }
+        return m_noteList[rowIndex - m_pinnedList.size()];
```
bjorn, line 391: You're changing one of 6 occurrences of this `getRef` function. Consider changing it to a private member and replacing all occurrences.

src/notelistmodel.cpp
```diff
@@ -453,7 +451,7 @@ bool NoteListModel::dropMimeData(const QMimeData *mime, Qt::DropAction action, i
                 auto lastMod = index.data(NoteDeletionDateTime).toDateTime();
                 for (destinationChild = 0; destinationChild < m_noteList.size();
                      ++destinationChild) {
-                    const auto &note = m_noteList[destinationChild];
+                    note = m_noteList[destinationChild];
```
bjorn, line 454: Erm, this massively changes the behavior. Previously, it was storing a read-only `NodeData` reference to `m_noteList[destinationChild]`, which was used to inspect the `deletionDateTime`. Whereas the new line copies `m_noteList[destinationChild]` into the previously referenced node data instead.

(also applies below)

src/nodetreeview.cpp
```diff
@@ -602,7 +602,7 @@ void NodeTreeView::mousePressEvent(QMouseEvent *event)
             case NodeItem::Type::TagItem: {
                 auto oldIndexes = selectionModel()->selectedIndexes();
                 for (const auto &ix : qAsConst(oldIndexes)) {
-                    auto itemType =
+                    itemType =
```
bjorn, line 605: I think this arbitrary variable re-use is confusing. This should be a new variable since it has little to do with the previous one. Maybe in the future, we end up changing the code such that the original `itemType` is still used beyond this scope, and then it may be changed to something random only in this case, which can be easy to miss.

If you really want to avoid shadowing warnings, the fix is to choose different names in my opinion.

src/customapplicationstyle.cpp
```diff
@@ -25,7 +25,7 @@ void CustomApplicationStyle::drawPrimitive(PrimitiveElement element, const QStyl
             painter->drawLine(option->rect.topLeft(), option->rect.topRight());
         } else {
             c.setAlpha(200);
-            QBrush brush(c);
+            brush.setColor(c);
```
bjorn, line 28: Here I would get rid of both `brush` variables instead. You can pass the `c` directly to `setBrush` or `fillRect` since colors implicitly convert to a solid brush.

src/dbmanager.cpp
```diff
@@ -1019,7 +1019,7 @@ NodeData DBManager::getNode(int nodeId)
                            R"("title" )"
                            R"(FROM node_table WHERE id=:id LIMIT 1;)");
             query2.bindValue(":id", node.parentId());
-            bool status = query2.exec();
+            status = query2.exec();
```
bjorn, line 1022: Do we need `status` at all? Why would we not just write `if (query2.exec()) {`?

src/mainwindow.cpp
```diff
@@ -396,17 +396,6 @@ void MainWindow::setupMainWindow()
...
-                                  : QStringLiteral("Roboto"));
-#elif _WIN32
-    QString m_displayFont(QFont(QStringLiteral("Segoe UI")).exactMatch()
-                                  ? QStringLiteral("Segoe UI")
-                                  : QStringLiteral("Roboto"));
-#else
-    QString m_displayFont(QStringLiteral("Roboto"));
-#endif
 #ifdef __APPLE__
     QFont m_titleFont(m_displayFont, 13, QFont::DemiBold);
```
bjorn, line 400: Can we rename this to `titleFont` while we're at it? Since it's not a member variable.

src/mainwindow.cpp
```diff
@@ -1181,7 +1170,6 @@ void MainWindow::setupDatabases()
         QFile noteDBFile(noteDBFilePath);
         noteDBFile.rename(dir.path() + QDir::separator() + QStringLiteral("oldNotes.db"));
         {
-            QFile noteDBFile(noteDBFilePath);
```
bjorn, line 1184: You can't just remove this line as far as I understand. That file was moved, and we want to open the new location for writing and not overwrite `oldNotes.db`. Likely the explicit new scope existed exactly for the purpose of introducing this new variable.

We could likely still reuse the variable, if we call `setFileName(noteDBFilePath)` on it after the call to `rename`. The extra scope can be removed as well.

bjorn (changes requested):
I think it's a good idea to enable more compiler warnings, but please be careful when "fixing" shadowed variable warnings. :-)

CMakeLists.txt
bjorn: As long as it is only enabled for debug builds, since we don't want packages failing to build because of that.

guihkx:
Oh, dear god. lol

And to think the only part I was questioning myself about was the removal of `QFile noteDBFile(noteDBFilePath);`, which I confidently ended up thinking was okay after reading the documentation... 🤡 

It only goes to show how much I have to learn. :~

@bjorn Thank you so much for taking your time to review this!

It's a bit late for me now and I have to rest, but I'll try to address every point as soon as I can...

CMakeLists.txt
zjeffer: I agree. Enabling these warnings and enforcing them will improve code quality.

src/notelistmodel.cpp
guihkx: I tried to implement this, but I don't think my way is the "correct" way of doing it, so it'd be great if you could take a look. Thanks!

src/notelistmodel.cpp
guihkx: You're absolutely right, thank you for spotting this!

src/nodetreeview.cpp
guihkx: Nice catch, and I completely agree. I renamed it to `selectedItemType`, which seems more accurate.

src/customapplicationstyle.cpp
guihkx: Perfect!

src/mainwindow.cpp
guihkx: Done!

src/mainwindow.cpp
guihkx: You're absolutely right. It was a complete documentation misread on my part. Sorry!

guihkx:
Round 2...

Besides addressing every point (or rather, trying to), I added an extra fix for the code that handles the v1.5.0 database migration to the new format. I basically just got rid of this warning:

```
QSqlQuery::value: not positioned on a valid record
```

Thanks!

src/dbmanager.cpp
guihkx: Good point! But I don't think I should update *all* instances of `status = query2.exec()` in this PR (there are around 80 of them!). It seems a little out of scope to me...

CMakeLists.txt
guihkx: @bjorn While that's an easy check to add to CMake, our Windows pipeline doesn't run debug builds, so these extra warnings will be essentially useless...

Wouldn't it be desirable to have this enabled for release builds as well, though? Especially considering that all these warnings are now fixed?

In any case, if you think it's going to cause more harm than good, I'm also fine with leaving them as just warnings.

src/notelistmodel.cpp
```diff
@@ -328,6 +312,22 @@ bool NoteListModel::isInAllNote() const
             && (m_listViewInfo.parentFolderId == SpecialNodeID::RootFolder);
 }
 
+NodeData &NoteListModel::getRef(int row)
```
guihkx, line 315: Is there a better way to do what I'm doing here? 😬 

These two methods are doing the exact same thing...

CMakeLists.txt
guihkx: Okay, for now I think we can leave this as-is.

Our CI pipeline will generate annotations whenever the compiler gives us warnings, so I think that's good enough and not disruptive...

Thanks!

src/notelistview.cpp
```diff
@@ -752,8 +751,8 @@ void NoteListView::onCustomContextMenu(QPoint point)
             const auto tagIds = m_tagPool->tagIds();
             for (const auto &id : tagIds) {
                 bool all = true;
-                for (const auto &index : qAsConst(indexList)) {
-                    auto tags = index.data(NoteListModel::NoteTagsList).value<QSet<int>>();
+                for (const auto &idx : qAsConst(indexList)) {
```
bjorn, line 754: Rather than creating variations of generic names, we should prefer specific names. I think in this case `selectedIndex` would be suitable (and in some other places, as far as I could see).

Renaming `indexList` to `selectedIndexes` would be good, but not necessarily in this patch.

bjorn (approved):
Looking good now! I added a few more comments, but nothing critical. :-)

src/notelistview.cpp
```diff
@@ -298,7 +298,6 @@ void NoteListView::mousePressEvent(QMouseEvent *e)
         if (!oldIndexes.contains(index)) {
             if (e->modifiers() == Qt::ControlModifier) {
                 setSelectionMode(QAbstractItemView::MultiSelection);
-                auto oldIndexes = selectionModel()->selectedIndexes();
                 if (oldIndexes.contains(index) && oldIndexes.size() > 1) {
```
bjorn, line 301: This `oldIndexes.contains(index)` check makes no sense, btw. If we get here, we already know that `!oldIndexes.contains(index)` was `true` after all. So either this code can be simplified or there is a bug.

src/notelistmodel.cpp
guihkx: TIL.

Then I guess I'll keep both versions and apply the style you mentioned. Thanks!

src/notelistmodel.cpp
```diff
@@ -453,17 +445,17 @@ bool NoteListModel::dropMimeData(const QMimeData *mime, Qt::DropAction action, i
                 auto lastMod = index.data(NoteDeletionDateTime).toDateTime();
                 for (destinationChild = 0; destinationChild < m_noteList.size();
                      ++destinationChild) {
-                    const auto &note = m_noteList[destinationChild];
-                    if (note.deletionDateTime() <= lastMod) {
+                    const auto &refNote = m_noteList[destinationChild];
```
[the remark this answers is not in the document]
guihkx: Agreed. I'll apply your first suggestion, thanks.

In part because even after I Googled 'inline variables', I'm still a bit confused about how to use them (and they seem to require a C++17 compiler, but I think we still support C++14 ones?)...

But most importantly, it's not cool to make you my personal Wikipedia every time... 😆

src/notelistview.cpp
guihkx: > I think in this case `selectedIndex` would be suitable (and in some other places, as far as I could see).

Agreed. 

> Renaming `indexList` to `selectedIndexes` would be good, but not necessarily in this patch.

Yeah, let's leave that one for our future selves... ^_^

src/notelistview.cpp
guihkx: Seems like a bogus check to me as well. It was probably a copy-and-paste mistake from here (including the shadowing I removed):

https://github.com/nuttyartist/notes/blob/7027131e309b2d374b4047c6b0c712ff5ab4f5ec/src/notelistview.cpp#L337-L345

I'll remove this check on a separate commit. Thanks for noticing it!

src/notelistmodel.cpp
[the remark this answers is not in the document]
bjorn: Erm, I don't mean asking the compiler to inline the variable, but to just do it by changing the code to:
```c++
if (m_noteList[destinationChild].deletionDateTime() <= lastMod) {
```

src/notelistview.cpp
bjorn: Yeah, somebody really liked to copy/paste code... there's so much duplication. :/

src/notelistmodel.cpp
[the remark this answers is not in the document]
guihkx: LMAO... Okay, I gotcha now.

src/notelistview.cpp
guihkx: Not only that, but it seems like that bogus `oldIndexes.contains(index)` check is what kept things working as they should, lol...

Check for yourself: Remove that check, and then use Ctrl+Click to select notes in the list view. You can no longer select more than two notes...

I removed most of those checks, and de-selecting notes while holding Ctrl still works here...

It'd be great if you could also confirm it before I commit:

```diff
diff --git a/src/notelistview.cpp b/src/notelistview.cpp
index 4ad8a96..9681ffd 100644
--- a/src/notelistview.cpp
+++ b/src/notelistview.cpp
@@ -298,12 +298,8 @@ void NoteListView::mousePressEvent(QMouseEvent *e)
         if (!oldIndexes.contains(index)) {
             if (e->modifiers() == Qt::ControlModifier) {
                 setSelectionMode(QAbstractItemView::MultiSelection);
-                if (oldIndexes.contains(index) && oldIndexes.size() > 1) {
-                    selectionModel()->select(index, QItemSelectionModel::Deselect);
-                } else {
-                    setCurrentIndex(index);
-                    selectionModel()->setCurrentIndex(index, QItemSelectionModel::SelectCurrent);
-                }
+                setCurrentIndex(index);
+                selectionModel()->setCurrentIndex(index, QItemSelectionModel::SelectCurrent);
                 auto selectedIndexes = selectionModel()->selectedIndexes();
                 emit notePressed(selectedIndexes);
             } else {
```

bjorn: > Check for yourself: Remove that check, and then use Ctrl+Click to select multiple notes in the list view. You can no longer select more than two notes...

Well, you can't just remove a check that is guaranteed to be `false`, because that changes the behavior of the code as if it were `true`. You indeed need to remove the conditional code as well, like you ended up doing. :-)

guihkx:
Just to be extra sure before I merge this, @bjorn could you please take a final look?

Thank you for reviewing and for your patience, by the way! 🏆

zjeffer (approved):

src/customapplicationstyle.cpp
```diff
@@ -16,17 +16,15 @@ void CustomApplicationStyle::drawPrimitive(PrimitiveElement element, const QStyl
         //        }
```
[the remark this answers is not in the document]
guihkx: Agreed! I'll do that.

src/notelistview.cpp
```diff
@@ -864,9 +859,9 @@ void NoteListView::onCustomContextMenu(QPoint point)
                 auto action = new QAction(folders[id], this);
                 connect(action, &QAction::triggered, this, [this, id] {
                     auto indexes = selectedIndexes();
-                    for (const auto &index : qAsConst(indexes)) {
-                        if (index.isValid()) {
-                            emit moveNoteRequested(index.data(NoteListModel::NoteID).toInt(), id);
+                    for (const auto &idx : qAsConst(indexes)) {
```
bjorn, line 862: This one could be renamed too.

bjorn (approved):
Looks fine, just one small comment. :-)

src/notelistview.cpp
guihkx: Riight. Done! 

P.S.: The formatting on this one looks a bit weird now (to my eyes anyway), because of `clang-format`.

guihkx:
Thank you all! Merging...

PR #474, opened 2023-02-15
Add CMake option to build without auto-updating features

guihkx:
This option allows us to build Notes without any sort of update checking code (including the auto-updater).

This is mostly useful for Linux package maintainers, where any update checking feature is essentially useless, since we don't (and shouldn't) have control over a Linux distribution's package upgrading process.

Nevertheless, this option is still enabled by default, meaning that both the update checker and the auto-updater _will_ still work.

To disable them, invoke CMake with `-DUPDATE_CHECKER=OFF`.


src/mainwindow.cpp
```diff
@@ -8,7 +8,11 @@
 #include "ui_mainwindow.h"
 #include "notelistdelegate.h"
 #include "qxtglobalshortcut.h"
-#include "updaterwindow.h"
+
+#if defined(UPDATE_CHECKER)
+#  include "updaterwindow.h"
+#endif
```
bjorn, line 14: Let's just remove this include since `mainwindow.h` already includes this file.

bjorn (approved):
Looks good.

nuttyartist:
Good feature to have for Linux users (or anyone using some package manager to install Notes), and good that you kept it on by default.

But it's critical to test that it works right and that updates are really being fetched. Can you try verify this by changing your local executable version to less than 2.0.0 and see if the updater automatically fetches version 2.0.0?

nuttyartist:
Okay, I've tested it, and it works just fine still. 👍  Thanks!

nuttyartist (approved):

guihkx:
Thanks, guys! Merging...

PR #476, opened 2023-02-16
Update shortcut to summon Notes on Windows

### LINKED ISSUE #475: The global shortcut to summon Notes doesn't work on Windows 11
Pressing <kbd>Win</kbd>+<kbd>N</kbd> pops up their [Notification Center](https://support.microsoft.com/en-us/windows/how-to-open-notification-center-and-quick-settings-f8dc196e-82db-5d67-f55e-ba5586fbb038) instead:

image

How about we change it to <kbd>Win</kbd>+<kbd>Shift</kbd>+<kbd>N</kbd> instead? Like we do for Linux/FreeBSD?

https://github.com/nuttyartist/notes/blob/06e8f849b914db679ce64a78bd73d12681eed271/src/mainwindow.cpp#L508-L509

By the way, if we're going to change it, we should probably use something like:

```cpp
#if (defined(Q_OS_UNIX) && !defined(Q_OS_MAC)) || defined(Q_OS_WIN)
```
Right?

guihkx:
The <kbd>Win</kbd>+<kbd>N</kbd> shortcut is now used by Windows 11 to bring up their [notification center](https://support.microsoft.com/en-us/windows/how-to-open-notification-center-and-quick-settings-f8dc196e-82db-5d67-f55e-ba5586fbb038), so replace it by <kbd>Win</kbd>+<kbd>Shift</kbd>+<kbd>N</kbd>.

Fixes #475

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ Our vision for the future on our [Wiki](https://github.com/nuttyartist/notes/wik
 - Markdown Support. Format text without lifting your hands from the keyboard.
 - Different themes. Switch between Light, Dark, and Sepia.
 - Feed View. Select multiple notes to see them all one after another in the editor.
-- Always runs in the background. Use the hotkey "Windows" + 'N' to summon Notes. "control" + "N" for macOS.
+- Always runs in the background. Use the hotkey <kbd>Win</kbd>+<kbd>Shift</kbd>+<kbd>N</kbd> to summon Notes. <kbd>Ctrl</kbd>+<kbd>N</kbd> for macOS.
 - Keyboard shortcuts. Meant to have the option to be used solely with a keyboard (but more work needs to be done there).
 - What feature will you contribute?
 
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -505,10 +505,10 @@ void MainWindow::setupKeyboardShortcuts()
     new QShortcut(QKeySequence(Qt::CTRL | Qt::Key_Backslash), this, SLOT(resetBlockFormat()));
 
     QxtGlobalShortcut *shortcut = new QxtGlobalShortcut(this);
-#if defined(Q_OS_LINUX) || defined(Q_OS_FREEBSD)
-    shortcut->setShortcut(QKeySequence(Qt::META | Qt::SHIFT | Qt::Key_N));
-#else
+#if defined(Q_OS_MACOS)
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -27,7 +27,7 @@ Our vision for the future on our [Wiki](https://github.com/nuttyartist/notes/wik
 - Markdown Support. Format text without lifting your hands from the keyboard.
 - Different themes. Switch between Light, Dark, and Sepia.
 - Feed View. Select multiple notes to see them all one after another in the editor.
-- Always runs in the background. Use the hotkey "Windows" + 'N' to summon Notes. "control" + "N" for macOS.
+- Always runs in the background. Use the hotkey <kbd>Win</kbd>+<kbd>Shift</kbd>+<kbd>N</kbd> to summon Notes. <kbd>Ctrl</kbd>+<kbd>N</kbd> for macOS.
```
bjorn, line 30: I would suggest to wrap the entire shortcut in `<kbd>` rather than each individual key. Or does that look bad?

src/mainwindow.cpp
```diff
@@ -505,7 +505,7 @@ void MainWindow::setupKeyboardShortcuts()
     new QShortcut(QKeySequence(Qt::CTRL | Qt::Key_Backslash), this, SLOT(resetBlockFormat()));
 
     QxtGlobalShortcut *shortcut = new QxtGlobalShortcut(this);
-#if defined(Q_OS_LINUX) || defined(Q_OS_FREEBSD)
+#if (defined(Q_OS_UNIX) && !defined(Q_OS_MAC)) || defined(Q_OS_WIN)
```
bjorn, line 508: At this point is seems easier to reverse this check, and just make `Qt::META | Qt::Key_N` a special case for macOS.

bjorn (approved):

README.md
guihkx: It would look *kinda* weird like that, in my opinion:

image

Right now it looks like this:

image

I'm not a designer though, so I could be wrong... lol

src/mainwindow.cpp
guihkx: Good point.

README.md
bjorn: Yeah, it looks fine like you did it.

nuttyartist (approved):

PR #478, opened 2023-02-16
macOS: Always 'close' ourselves to dock

### LINKED ISSUE #458: Hiding Notes to the macOS dock doesn't seem to change its visibility state
After "closing" Notes to the macOS dock, somehow Qt still thinks the window is visible (verifiable by calling [`isVisible()`](https://doc.qt.io/qt-5/qwidget.html#visible-prop)).

The side effect of this is having to double-click the system tray icon to restore Notes, instead of just clicking it once, which is what the intended behavior should be.

---

**Steps to reproduce:**

1. Close Notes to dock (by clicking on the red close button)
2. Click on Notes' tray icon only once

**What should happen:**

Notes' window should be restored.

**What happens instead:**

Nothing, because Qt still thinks the window is visible at this point, so this single click ends up hiding the window again, therefore requiring yet another click to actually restore the window.

**Bug in action:**

https://user-images.githubusercontent.com/626206/214741606-ec0ce560-194e-4250-8188-c4dbe2ff0cdc.mp4

guihkx:
**Bug 1:**

Closing the window on macOS merely hides the app to the dock. However, because Qt is never informed of this, it will still think its window is visible when we try to restore it, requiring the user to click on the tray icon twice, for example.

We fix this by monitoring AppKit's [`hide`](https://developer.apple.com/documentation/appkit/nsworkspacedidhideapplicationnotification) / [`unhide`](https://developer.apple.com/documentation/appkit/nsworkspacedidunhideapplicationnotification) notifications (events), and then we simply call Qt's `hide()` / `show()` methods, accordingly.

Fixes #458

---

**Bug 2:**

This is still related to bug 1: Clicking on the tray icon will call Qt's `hide()` / `show()` methods, but AppKit is also never informed of this.

So, we listen to Qt's `showEvent` / `hideEvent` and then we simply call AppKit's `hide:` / `unhide:` methods, accordingly.

Unfortunately, this does not seem to work every time, for some reason.

But I'd argue that's still better than not doing anything...🤷

### DIFF
--- a/src/framelesswindow.h
+++ b/src/framelesswindow.h
@@ -131,6 +131,8 @@ signals:
     void toggleFullScreen(bool isFullScreen);
 
 protected:
+    void hideEvent(QHideEvent *event);
+    void showEvent(QShowEvent *event);
     void mousePressEvent(QMouseEvent *event);
     void mouseReleaseEvent(QMouseEvent *event);
     void mouseMoveEvent(QMouseEvent *event);
--- a/src/framelesswindow.mm
+++ b/src/framelesswindow.mm
@@ -205,6 +205,20 @@ void CFramelessWindow::setDraggableAreaHeight(int height)
     m_draggableHeight = height;
 }
 
+void CFramelessWindow::hideEvent(QHideEvent *event)
+{
+    // FIXME: Figure out why this fails to trigger our AppObserver listener sometimes. :/
+    [NSApp hide:nil];
+    QMainWindow::hideEvent(event);
+}
+
+void CFramelessWindow::showEvent(QShowEvent *event)
+{
+    // FIXME: Figure out why this fails to trigger our AppObserver listener sometimes. :/
+    [NSApp unhide:nil];
+    QMainWindow::showEvent(event);
+}
+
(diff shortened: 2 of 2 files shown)


guihkx:
Sorry for the extra `push`, I had to make the comment I added a bit more accurate there.

nuttyartist:
Alrighty. The system tray isn't very consistent yet, but it's ok for now. 

I see that system tray is not default on macOS. What about other operating systems?

guihkx:
> I see that system tray is not default on macOS.

Well, technically there is a system tray icon on macOS, and it's enabled by default...

What we don't there have is a system tray context menu.

>What about other operating systems?

On non-macOS, we have both a system tray icon and context menu.

nuttyartist:
Oh okay, I probably had the settings off before testing already hah.

Thanks for the fixes, @guihkx!

nuttyartist (approved):

guihkx:
Thanks!

guihkx:
I don't know how I messed up this badly, but these changes were merged.

PR #487, opened 2023-02-19
Fix settings not saving when closing the app using native buttons or system shortcut

zjeffer:
I made a mistake in #470: I didn't call QuitApplication(), so the states were never saved to the settings file. This should be fixed now.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -2535,7 +2535,7 @@ void MainWindow::onRedCloseButtonClicked()
     if (m_hideToTray) {
         setMainWindowVisibility(false);
     } else {
-        close();
+        QuitApplication();
     }
 }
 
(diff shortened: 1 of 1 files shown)


zjeffer:
Thanks for rebasing :+1:

guihkx:
No problem!

Change looks sane to me.

guihkx (approved):

nuttyartist:
That's very important 😅. I'll test all scenarios of quitting the app once you tell me it's ready for testing

zjeffer:
Tested on bpswm and Ubuntu 22.10. Here's how I tested it in Gnome:

https://user-images.githubusercontent.com/4633209/219964862-a8b3ea85-a872-40ff-86a8-2f28e418e86e.mp4



* First with an older build: quitting the application won't make changes to the settings file
* Then with this build: quitting the application does make changes to the file.

zjeffer:
Found another bug: closing with the red close button (non-native frame) would also not call the QuitApplication() method, now fixed.

nuttyartist:
Great, feel free to merge this. I'll do the testing tomorrow.

PR #488, opened 2023-02-20
Added version to About window

### LINKED ISSUE #485: The app version should also be added to the About window
Right now, the only way for users to check what version of Notes they're running, is by opening the update checker.

However, after #474 was added, we can now build Notes without any update checking feature, leaving users with no other place to check what version of Notes they're running.

My suggestion would be adding that information to the About window.

Perhaps another easy and nice addition would be adding what Qt version was Notes built against, and also what Qt version we're currently using to run Notes.

zjeffer:
closes #485

I added the app version and qt version at the bottom of the About window, here's how it looks on Ubuntu22.10:

image

It uses the app version from the CMakeLists.txt file, so here it shows I'm using the AppImage from a GitHub action.


guihkx:
Nice. Thanks for opening this PR.

If I may suggest something, I think this information would make more sense at the very top instead (because of its relevancy).

Also, couldn't we add what Qt version was used to build Notes too? Maybe something like:

**Qt version**: 5.15.8 (built with 5.15.2)

zjeffer:
New location + "(built with x)" added: 

image

zjeffer:
I'm not sure what's the best way to test whether the compiled version & runtime version are correct.  I guess the AppImages always have the same version?

guihkx:
It might look a bit out of place there, but I still think it should be at the very top somehow, so LGTM as an initial design. :p

>I guess the AppImages always have the same version?

They have. The runtime version will only really change in case Notes is using a distro's Qt package, as they might update independently from us.

guihkx (approved):

nuttyartist:
Great. But please move the QT version to the bottom of the page (keeping only the app version at the top). Our app is used by many who are not programmers. Most of them don't need to see that info at the top. But it's good the app version stays at the top.

zjeffer:
Ok, it looks like this now: 

image

Either way, I think we should do a complete redesign of the About window in the future.

src/aboutwindow.cpp
```diff
@@ -12,24 +12,30 @@ AboutWindow::AboutWindow(QWidget *parent) : QDialog(parent), m_ui(new Ui::AboutW
...
-            "Github</a>.<br/><br/><strong>Acknowledgments</strong><br/>This project couldn't have "
-            "come this far without the help of these amazing "
-            "people:<br/><br/><strong>Programmers:</strong><br/>Alex Spataru<br/>Ali Diouri"
-            "<br/>David Planella<br/>Diep Ngoc<br/>Guilherme "
-            "Silva<br/>Thorbjørn Lindeijer<br/>Tuur Vanhoutte<br/>Waqar "
-            "Ahmed<br/><br/><strong>Designers:</strong><br/>Kevin Doyle<br/><br/>And to the "
-            "many of our beloved users who keep sending us feedback, you are an essential force in "
-            "helping us improve, thank you!<br/><br/><strong>Notes makes use of the following "
-            "third-party libraries:</strong><br/><br/>QMarkdownTextEdit<br/>"
+            "<strong>Version:</strong> " + qApp->applicationVersion()
```
bjorn, line 15: I would suggest to write `QCoreApplication::applicationVersion()` (also for the name above), since these functions are static so I think it's a bit confusing to call them like member functions.

Issue #489, opened 2023-02-20
Issues on Linux?


nuttyartist:
Testing version 2.1.0, I'm not sure if it's just on the distro I've tested (Elementary OS 6), everything seems to work well apart from these two issues:

1. I can't run the QT 6 appimage (I get the following error from in terminal), although Qt5 works well:

```
/tmp/.mount_Notes_zZcQgr/AppRun.wrapped: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by /tmp/.mount_Notes_zZcQgr/AppRun.wrapped)
/tmp/.mount_Notes_zZcQgr/AppRun.wrapped: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by /tmp/.mount_Notes_zZcQgr/usr/bin/../lib/libQt6Core.so.6)
/tmp/.mount_Notes_zZcQgr/AppRun.wrapped: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by /tmp/.mount_Notes_zZcQgr/usr/bin/../lib/libglib-2.0.so.0)
/tmp/.mount_Notes_zZcQgr/AppRun.wrapped: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by /tmp/.mount_Notes_zZcQgr/usr/bin/../lib/libsystemd.so.0)
/tmp/.mount_Notes_zZcQgr/AppRun.wrapped: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.30' not found (required by /tmp/.mount_Notes_zZcQgr/usr/bin/../lib/libsystemd.so.0)
```

2.  The app icon looks blurry (at least on Elementary).

![Screenshot from 2023-02-20 21 44 07](https://user-images.githubusercontent.com/16375940/220203634-a2612fbb-c7cf-4dbf-ad19-f3c66d62602e.png)

BTW, seems to work well on macOS.
I'll test on Windows tomorrow as well.

@guihkx

guihkx:
I'm not familiar with elementaryOS, but based on my Google search, version 6 is based on Ubuntu 20.04 Focal Fossa.

And Ubuntu 20.04 already ships [glibc 2.31](https://packages.ubuntu.com/source/focal/glibc), so none of these errors shouldn't even be showing up.

So either I'm completely wrong, or perhaps you're using an older version of elementary. Can you please double check with:

[...]

Issue #472, opened 2023-02-13
Qt 5 "deactivates" the window once you stop moving it around


guihkx:
Sorry for the terrible title, but I don't know how to better describe it. 😄 

Basically, after the [introduction of `startSystemMove()`](https://github.com/nuttyartist/notes/pull/445), after you finish moving the window, Qt 5.15 seems to "deactivate" the window (tested only on Linux at the moment), and you can't "interact" with any of the widgets using the mouse, unless you activate the window again.

Still lost? Hopefully the videos below will explain this better. 😅 

**Steps to reproduce**

1. Drag the window around a bit
2. Release it
3. Hover over any widget inside the window (e.g. trash icon, buttons, notes list, etc.)

**Current behavior**

You can not interact with widgets right away using your mouse, you must first click on the window to "re-activate" it:

https://user-images.githubusercontent.com/626206/218461733-0f347fd9-dced-41d0-974c-9aa16a4addd0.mp4

**Expected behavior**

You should still be able to interact with widgets using your mouse, like this (Qt 6.3):

https://user-images.githubusercontent.com/626206/218461771-20d89d65-d77c-4845-b266-9b47140064f3.mp4

I haven't found a workaround for this yet...

nuttyartist:
Hmm very weird. I can't reproduce this on my macOS. Have you tested on Windows?

guihkx:
Curiously, I have, and it doesn't seem to be a problem there.

I have yet to check if `startSystemMove()` is actually being called there, though.

Because if it were, I *think* we'd be able to snap windows by now (#414), and we currently can't... 🤔

nuttyartist:
> Because if it were, I think we'd be able to aero snap our window by now (https://github.com/nuttyartist/notes/issues/414), and we currently can't... 🤔

You sure? Last time I tested it worked. I will test again.

guihkx:
I tried it a couple of days ago on a Windows 11 VM and it wasn't working... 🤷 

But please do check, so we can close #414!

guihkx:
I think I found the upstream bug, which appears to only affect the Xcb plugin, namely X11 sessions:

- http://bugreports.qt.io/browse/QTBUG-91077

And the fix, which unfortunately was only included in Qt 6.3:

- https://codereview.qt-project.org/c/qt/qtbase/+/343520

I personally don't see any easy workarounds that we could add to circumvent this in our Qt 5 build, but if someone has an idea, please do share. :)

[...]

PR #492, opened 2023-02-21
Linux: Restore code for frameless window resizing

### LINKED ISSUE #483: [Regression] [Linux] Window resizing is broken in frameless mode
Window resizing in v1.5.0 (frameless mode):

https://user-images.githubusercontent.com/626206/219870703-41d77515-b652-43bd-a2e4-1997e93eed0f.mp4

<br>

Window resizing in v2.0.0 or `dev` (frameless mode):

https://user-images.githubusercontent.com/626206/219870792-bab25c1f-5d6c-4e2d-8a7c-4f8ca2db5eda.mp4

<br>

As you can see, we can't resize the window in v2.0.0 like we could in v1.5.0.

Because frameless mode is the default experience, we should probably fix this before releasing 2.1.0.

PS: This is unrelated to #454.

guihkx:
Linux users can not easily resize Notes' window if they have 'frameless mode' enabled (which is the default behavior).

This change restores this ability, which was present in v1.5.0, but was later removed in v2.0.0 by commit ddd79edf077158e0e62cc2b432f669e60a7d9bfb.

Fixes #483

---

**Before:**

https://user-images.githubusercontent.com/626206/220225135-ad2ba1c8-092b-4ab9-8646-dbc659a2f3bc.mp4

<br>

**After:**

https://user-images.githubusercontent.com/626206/220225150-ac97a46c-1e80-4b7e-8930-ab6c258ed64f.mp4

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -248,7 +248,7 @@ void MainWindow::saveLastSelectedNote(const QSet<int> &notesId)
  */
 void MainWindow::paintEvent(QPaintEvent *event)
 {
-#if !defined(__APPLE__) && !defined(Q_OS_LINUX) && !defined(Q_OS_FREEBSD)
+#if defined(Q_OS_UNIX) && !defined(Q_OS_MACOS)
     if (!m_useNativeWindowFrame) {
         QPainter painter(this);
         painter.save();
@@ -309,6 +309,9 @@ void MainWindow::setupMainWindow()
 #if !defined(Q_OS_MAC)
     this->setWindowFlags(Qt::Window | Qt::CustomizeWindowHint);
 #endif
+#if defined(Q_OS_UNIX) && !defined(Q_OS_MACOS)
+    this->setAttribute(Qt::WA_TranslucentBackground);
+#endif
 
     m_greenMaximizeButton = new QPushButton(this);
     m_redCloseButton = new QPushButton(this);
@@ -343,19 +346,15 @@ void MainWindow::setupMainWindow()
     m_editorDateLabel = ui->editorDateLabel;
     m_splitter = ui->splitter;
 
-#if defined(Q_OS_LINUX) || defined(Q_OS_FREEBSD)
-    //    QMargins margins(m_layoutMargin, m_layoutMargin, m_layoutMargin, m_layoutMargin);
-    //    setMargins(margins);
-    setMargins(QMargins(0, 0, 0, 0));
+#if defined(Q_OS_UNIX) && !defined(Q_OS_MACOS)
+    QMargins margins(m_layoutMargin, m_layoutMargin, m_layoutMargin, m_layoutMargin);
+    setMargins(margins);
 #elif _WIN32
     ui->verticalSpacer->changeSize(0, 7, QSizePolicy::Fixed, QSizePolicy::Fixed);
     ui->verticalSpacer_upEditorDateLabel->changeSize(0, 27, QSizePolicy::Fixed, QSizePolicy::Fixed);
 #endif
     ui->frame->installEventFilter(this);
-    ui->frameRight->setMouseTracking(true);
     ui->centralWidget->setMouseTracking(true);
-    ui->frame->setMouseTracking(true);
-    ui->frameRightTop->setMouseTracking(true);
     this->setMouseTracking(true);
     QPalette pal(palette());
     pal.setColor(QPalette::Window, QColor(248, 248, 248));
@@ -958,7 +957,7 @@ void MainWindow::setupTextEditStyleSheet(int paddingLeft, int paddingRight)
 {
     m_textEdit->setDocumentPadding(paddingLeft, 0, paddingRight, 2);
 
-#if defined(Q_OS_LINUX) || defined(Q_OS_FREEBSD)
+#if defined(Q_OS_UNIX) && !defined(Q_OS_MACOS)
     QString ss =
             QString("QTextEdit {background-color: %1;} "
                     "QTextEdit{selection-background-color: rgb(63, 99, 139);}"
@@ -1252,11 +1251,11 @@ void MainWindow::restoreStates()
                 m_settingsDatabase->value(QStringLiteral("editorSettingsWindowGeometry"))
                         .toByteArray());
 
-#if defined(Q_OS_LINUX) || defined(Q_OS_FREEBSD)
-        /// Set margin to zero if the window is maximized
-        //    if (isMaximized()) {
-        //        setMargins(QMargins(0,0,0,0));
-        //    }
+#if defined(Q_OS_UNIX) && !defined(Q_OS_MACOS)
+    // Set margin to zero if the window is maximized
+    if (isMaximized()) {
+        setMargins(QMargins());
+    }
 #endif
 
 #if defined(UPDATE_CHECKER)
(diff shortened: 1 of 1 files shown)


src/mainwindow.cpp
```diff
@@ -2189,8 +2190,8 @@ void MainWindow::maximizeWindow()
 void MainWindow::minimizeWindow()
 {
 #if defined(Q_OS_LINUX) || defined(Q_OS_FREEBSD)
```
bjorn, line 2192: In other places we're now using `#if defined(Q_OS_UNIX) && !defined(Q_OS_MACOS)`, and I think it would be good to change it here too.

src/mainwindow.cpp
```diff
@@ -2166,7 +2167,7 @@ void MainWindow::maximizeWindow()
 
     } else {
         setWindowState(windowState() | Qt::WindowMaximized);
-        //        setMargins(QMargins(0,0,0,0));
+        setMargins(QMargins(0, 0, 0, 0));
```
bjorn, line 2170: I'll just note that `QMargins()` already "Constructs a margins object with all margins set to 0.", so I would suggest to leave out the `0, 0, 0, 0` everywhere.

bjorn (commented):
Haven't tested this, just adding two small comments.

Regarding the before / after captures, what happened to the shadow in the after version?

src/mainwindow.cpp
guihkx: Absolutely! I just deliberately chose not to do that in this PR to avoid making it too overwhelming for reviewers. :p

As soon as this and other PRs I have are accepted (or rejected :p), I'll send a new one updating those as well. :D

src/mainwindow.cpp
guihkx: Good catch.

guihkx:
> Regarding the before / after captures, what happened to the shadow in the after version?

Good question... I'll try to investigate that once I get home.

guihkx:
Ah, I *think* I got it.

The shadows are still there, they're just less intensive. On a complete white background, they stand out a bit better:

image

And I think that happens because we're making the window background translucent (with [`Qt::WA_TranslucentBackground`](https://doc.qt.io/qt-5/qt.html#:~:text=Qt%3A%3AWA_TranslucentBackground)).

If I don't set the translucent attribute, the shadows are much more intensive (though the window itself looks terrible because of the margins added to resize the window):

image

I couldn't find a way to set the background completely transparent, but in my opinion this does not look too bad. It's certainly better than not being able to resize the window...

nuttyartist:
I tested the latest build and works very well. Solves the problem I mentioned in #489.

nuttyartist:
Oh well. It seems to work only on the top part of the window, not the rest.

guihkx:
> Oh well. It seems to work only on the top part of the window, not the rest.

It doesn't match what I show on my second video, then?

>I'll try to create a video.

Thanks, that will be helpful, I think.

nuttyartist:
> > Oh well. It seems to work only on the top part of the window, not the rest.
> 
> It doesn't match what I show on my second video, then?
> 
> > I'll try to create a video.
> 
> Thanks, that will be helpful, I think.

No... it seems to work only well on the top editor area.

https://user-images.githubusercontent.com/16375940/220425795-d463cfb5-cf04-4718-bd26-5371c6949590.mp4

guihkx:
Uh-oh. That looks unusual to me. 

I'm not sure what if the problem is elementaryOS-specific (works well on Ubuntu 18.04 with Unity though, which is what elementaryOS 5 is based on).

I'll setup a VM to test it out.

guihkx:
I just noticed something on your video that should be fixed (in a subsequent PR).

At around 0:16 seconds, you were able to resize the window there. But that's wrong, because you should never be able to resize the window while the cursor is hovering *inside* the window, only *outside* of it.

Anyway, I could not reproduce your issue on a live CD image of elementaryOS 5:

https://user-images.githubusercontent.com/626206/220441018-23f4ce2f-6a08-4235-b09c-d3fe0582bde1.mp4

<br>

But yeah, the blurry icon issue you mentioned in #489 is real.

guihkx:
I'm thinking you could've had a different version of Notes already running when you tried to open the AppImage from this pull request, because the shadows around the window on your video look different from mine.

bjorn:
> I'm thinking you could've had a different version of Notes already running when you tried to open the AppImage from this pull request, because the shadows around the window on your video look different from mine.

This happens to me all the time as well when I try to test stuff, heh.

> The shadows are still there, they're just less intensive. On a complete white background, they stand out a bit better:

In that case, could it be that we're rendering the shadows in gray rather than black? Because they look like they are lighter than the background in your second screen capture.

nuttyartist:
> I'm thinking you could've had a different version of Notes already running when you tried to open the AppImage from this pull request, because the shadows around the window on your video look different from mine.

Oh, maybe. I'll test again and report back.

> In that case, could it be that we're rendering the shadows in gray rather than black? Because they look like they are lighter than the background in your second screen capture.

They definitely don't look good in that video. I will test now on my end...

guihkx:
>In that case, could it be that we're rendering the shadows in gray rather than black?

It could be, but I'm not really a graphics person and the code in  `gaussianDist()` / `fillRectWithGradient()` scares me a little bit. 😅

Just keep in mind that the shadows have been like that v1.0.0, all I'm doing here is restoring the code so we can resize the window again.

>it's still very inconsistent on my machine

Just keep in mind that the Qt 5 build has some mouse-grabbing issues with `startSystemMove()` (I reported it in #472), so once you release the mouse button, the cursor gets "stuck" until you left click anywhere inside the window again.

I also just fixed the issue where you were able to resize the window while your mouse was inside of it.

Now you should only be able to resize it by its outer edges. Please test again once you can.

>And apparently, it's because Hide to tray icon was on but elementary doesn't show up the icon in the tray area (I think it doesn't support it?) so I kept thinking the app was closed while it was still open until I disabled it.

Ohhh, right. Both GNOME and elementary don't support tray icons natively, so when we close the window, Notes "hides" itself and there's no way of restoring it because we don't have the system tray...

It should be an easy fix, I think: We probably just have to check if there's a working system tray with [`QSystemTrayIcon::isSystemTrayAvailable()`](https://doc.qt.io/qt-5/qsystemtrayicon.html#isSystemTrayAvailable) when the user requests to close the app.

nuttyartist:
> It should be an easy fix, I think: We probably just have to check if there's a working system tray with [QSystemTrayIcon::isSystemTrayAvailable()](https://doc.qt.io/qt-5/qsystemtrayicon.html#isSystemTrayAvailable) when the user requests to close the app.

Nice one.

> Now you should only be able to resize it by its outer edges. Please test again once you can.

Sure I'll test now.

nuttyartist:
> The Qt 5 build has some mouse-grabbing issues with startSystemMove() (I reported it in https://github.com/nuttyartist/notes/issues/472), so once you release the mouse button, the cursor gets "stuck" until you left click anywhere inside the window again.

That's annoying. With Qt6 build it doesn't happen?

Anyway, tt seems to work like 2.0.0 (but better because the width of the panes doesn't change now on each resize).

So I approve this PR.

guihkx:
>That's annoying. With Qt6 build it doesn't happen?

That's correct, it's a Qt5-only issue on Linux (I think it's also specific to X11).

Anyway, I just rebased against `dev` so I can merge this branch manually (merging with GitHub's own tool removes my commit signature).

Thanks for testing, y'all!

PR #493, opened 2023-02-22
Make sure a system tray exists before hiding ourselves

guihkx:
Some Linux desktop environments (e.g. GNOME) don't support system tray icons natively, so let's make sure a system tray exists before we try to hide our window, otherwise users won't be able to quit the app.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -176,7 +176,8 @@ void MainWindow::InitData()
         watcher->setFuture(migration);
     }
     /// Check if it is running with an argument (ex. hide)
-    if (qApp->arguments().contains(QStringLiteral("--autostart"))) {
+    if (qApp->arguments().contains(QStringLiteral("--autostart"))
+        && QSystemTrayIcon::isSystemTrayAvailable()) {
         setMainWindowVisibility(false);
     }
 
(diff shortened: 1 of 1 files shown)


guihkx:
This was first noticed by @nuttyartist in https://github.com/nuttyartist/notes/pull/492#issuecomment-1439978645 while he was testing in elementaryOS 5.

nuttyartist:
Well, if system tray doesn't exist maybe we shouldn't support/show the Menu Action at all? I think that way is better.

src/mainwindow.cpp
```diff
@@ -2551,7 +2552,7 @@ void MainWindow::closeEvent(QCloseEvent *event)
     if (!event->spontaneous() || !isVisible()) {
         return;
     }
-    if (m_hideToTray && m_trayIcon->isVisible()) {
+    if (m_hideToTray && m_trayIcon->isVisible() && QSystemTrayIcon::isSystemTrayAvailable()) {
```
bjorn, line 2555: I suppose we could drop all the `m_trayIcon->isVisible()` checks, since it is always set to visible.

nuttyartist:
Although as a note, the PR does work for me on Elementary OS 5 Qt5.

bjorn:
> Well, if system tray doesn't exist maybe we shouldn't support/show the Menu Action at all? I think that way is better.

I thought the same thing, though at least on Linux the system tray can become available or disappear while the app is running. The documentation notes:

> If the system tray is currently unavailable but becomes available later, [QSystemTrayIcon](https://doc.qt.io/qt-6/qsystemtrayicon.html) will automatically add an entry in the system tray if it is [visible](https://doc.qt.io/qt-6/qsystemtrayicon.html#visible-prop).

Unfortunately there appears to be no signal provided so we could know when the system tray becomes available. I'm not sure if it would be alright to just check the availability on startup.

bjorn (approved):

nuttyartist:
Oh ok, then. Let's keep it as it is.

nuttyartist (approved):

guihkx:
Thanks for testing :)

Merging...

PR #495, opened 2023-02-23
Fixed markdown header colors after re-enabling markdown

bjorn:
They were getting reset to the defaults when the highlighter was re-created. Fixed by not re-creating the `CustomMarkdownHighlighter`.


src/noteeditorlogic.cpp
```diff
@@ -20,6 +20,7 @@ NoteEditorLogic::NoteEditorLogic(CustomDocument *textEdit, QLabel *editorDateLab
                                  DBManager *dbManager, QObject *parent)
     : QObject(parent),
       m_textEdit{ textEdit },
+      m_highlighter{ new CustomMarkdownHighlighter { m_textEdit->document() } },
```
guihkx, line 23:
```suggestion
      m_highlighter{ new CustomMarkdownHighlighter{ m_textEdit->document() } },
```

Just to make `clang-format` happy.

guihkx (approved):

zjeffer (approved):

src/noteeditorlogic.cpp
bjorn: It's not the way I like it, but oh well. :-)

nuttyartist:
Good catch!

PR #499, opened 2023-02-23
Use `Qt::FramelessWindowHint` on Linux and friends

guihkx:
This hint was removed in commits 8d7cb3e and 20a4d42, but without it, Linux window managers will still use native window decorations on Wayland sessions.

**Before (on KDE Plasma 5.27 + Wayland)**:

kde-before

**After (on KDE Plasma 5.27 + Wayland)**:

kde-after

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -156,7 +156,11 @@ void MainWindow::InitData()
         QProgressDialog *pd =
                 new QProgressDialog(tr("Migrating database, please wait."), QString(), 0, 0, this);
         pd->setCancelButton(Q_NULLPTR);
+#if defined(Q_OS_UNIX) && !defined(Q_OS_MACOS)
+        pd->setWindowFlags(Qt::Window | Qt::FramelessWindowHint | Qt::FramelessWindowHint);
+#else
         pd->setWindowFlags(Qt::Window | Qt::CustomizeWindowHint);
+#endif
         pd->setMinimumDuration(0);
         pd->show();
 
@@ -308,8 +312,13 @@ MainWindow::~MainWindow()
 void MainWindow::setupMainWindow()
 {
 #if !defined(Q_OS_MAC)
-    this->setWindowFlags(Qt::Window | Qt::CustomizeWindowHint);
+    auto flags = Qt::Window | Qt::CustomizeWindowHint;
+#  if defined(Q_OS_UNIX)
+    flags |= Qt::FramelessWindowHint;
+#  endif
+    setWindowFlags(flags);
 #endif
+
 #if defined(Q_OS_UNIX) && !defined(Q_OS_MACOS)
     this->setAttribute(Qt::WA_TranslucentBackground);
 #endif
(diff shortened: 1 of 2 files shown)


src/mainwindow.cpp
```diff
@@ -157,6 +157,9 @@ void MainWindow::InitData()
                 new QProgressDialog(tr("Migrating database, please wait."), QString(), 0, 0, this);
         pd->setCancelButton(Q_NULLPTR);
         pd->setWindowFlags(Qt::Window | Qt::CustomizeWindowHint);
+#if defined(Q_OS_UNIX) && !defined(Q_OS_MACOS)
+        pd->setWindowFlags(pd->windowFlags() | Qt::FramelessWindowHint);
+#endif
```
bjorn, line 162: I'd prefer if where possible you'd avoid calling `setWindowFlags` twice where easily possible.

guihkx: Right, that's easy avoidable. On `MainWindow::setupMainWindow()` it's a bit awkward because of how it looks right now, but I could probably do something like this:

```cpp
#if !defined(Q_OS_MAC)
    auto flags = Qt::Window | Qt::CustomizeWindowHint;
#  if defined(Q_OS_UNIX)
    flags |= Qt::FramelessWindowHint;
#  endif
    setWindowFlags(flags);
#endif
```

But my question is: Does that look good and readable?

guihkx: Okay, I've applied that. ☝️ 

Let me know if there's a better-looking away and I'll do it. Thanks!

nuttyartist:
👏👏

src/mainwindow.cpp
bjorn: Since we're essentially always using `Qt::CustomizeWindowHint` on some platforms and `Qt::CustomizeWindowHint | Qt::FramelessWindowHint` on other platforms, we could define a global variable that we can then use without platform defines elsewhere:
```c++
#if defined(Q_OS_UNIX)
constexpr auto framelessWindowFlags = Qt::CustomizeWindowHint | Qt::FramelessWindowHint;
#else
constexpr auto framelessWindowFlags = Qt::CustomizeWindowHint;
#end
```
Not sure if we already have some appropriate header file for this.

And actually, was there a reason to use `Qt::CustomizeWindowHint` in addition to `Qt::FramelessWindowHint` on Unix? Otherwise we could simplify that to:
```c++
#if defined(Q_OS_UNIX)
constexpr auto framelessWindowFlag = Qt::FramelessWindowHint;
#else
constexpr auto framelessWindowFlag = Qt::CustomizeWindowHint;
#end
```

guihkx: >we could define a global variable that we can then use without platform defines elsewhere

I like this approach.

>Not sure if we already have some appropriate header file for this.

I don't think we do... Could I perhaps that to `framelesswindow.h`? Though it'd be a bit awkward because that header is actually from a third-party project. I can create another header, but I don't know, it seems a bit overkill just to include that small code, no? Or is that not uncommon in C++ projects?

>And actually, was there a reason to use `Qt::CustomizeWindowHint` in addition to `Qt::FramelessWindowHint` on Unix? 

That was done in #455, according to my interpretation of the Qt [documentation](https://doc.qt.io/qt-5/qt.html#:~:text=The%20CustomizeWindowHint%20flag):

>The `CustomizeWindowHint` flag is used to enable customization of the window controls. This flag must be set to allow the `WindowTitleHint`, `WindowSystemMenuHint`, `WindowMinimizeButtonHint`, `WindowMaximizeButtonHint` and `WindowCloseButtonHint` flags to be changed.

I figured: "We might as well add `CustomizeWindowHint` to non-Windows platforms too, since we're doing all of those things described in the docs". But sure, it seems to work just fine without it too...

bjorn (approved):

src/mainwindow.cpp
bjorn: I think the latest version of your patch is fine, let's not overthink things. :-)

PR #500, opened 2023-02-23
Linux: Update window margins on maximize/unmaximize

guihkx:
Now that we delegate window moving to the window manager via `startSystemMove()`, some window managers on Linux allow us to snap the window to the top edge of the screen, in order to maximize it.

Since we didn't know when this would happen, we were not removing the margins added around the window (when in 'frameless mode'), causing them to stay visible even when the window was maximized.

This is a follow-up to #492.

**Before:**

https://user-images.githubusercontent.com/626206/221020903-9cf53ade-5570-48c1-aedb-7375ebeceb17.mp4

<br>

**After:**

https://user-images.githubusercontent.com/626206/221020953-a802a396-2ffd-495f-a170-e9036eb46c08.mp4


src/mainwindow.cpp
```diff
@@ -3215,6 +3215,22 @@ void MainWindow::leaveEvent(QEvent *)
     this->unsetCursor();
 }
 
+/*!
+ * \brief MainWindow::changeEvent
+ */
+void MainWindow::changeEvent(QEvent *event)
+{
+#if defined(Q_OS_LINUX) && !defined(Q_OS_MACOS)
```
guihkx, line 3223: Dang it, I meant `Q_OS_UNIX`.

src/mainwindow.cpp
```diff
@@ -3215,6 +3215,22 @@ void MainWindow::leaveEvent(QEvent *)
     this->unsetCursor();
 }
 
+/*!
+ * \brief MainWindow::changeEvent
+ */
+void MainWindow::changeEvent(QEvent *event)
+{
```
bjorn, line 3222: When overriding virtual functions, we should generally not assume that the superclass doesn't need their implementation to be called, so I think we should add `MainWindowBase::changeEvent(event);`.

src/mainwindow.cpp
```diff
@@ -3215,6 +3215,22 @@ void MainWindow::leaveEvent(QEvent *)
...
+#if defined(Q_OS_UNIX) && !defined(Q_OS_MACOS)
+    if (!event->spontaneous() || m_useNativeWindowFrame)
```
bjorn, line 3224: What is the `spontaneous()` check meant to achieve? Why do we care whether this event "originated outside the application"?

Since this is only about adjusting to the maximized state, I think it would be appropriate to check `event->type() == QEvent::WindowStateChange` here.

guihkx:
image

☝️  Any way I can elegantly get rid of these warnings on macOS and Windows? I'm thinking of:

```cpp
#if defined(Q_OS_UNIX) && !defined(Q_OS_MACOS)
    // do the thing
#else
    Q_UNUSED(event);
#endif
```

But that doesn't look very elegant, I think. :p 

I can probably declare `changeEvent()` only on Linux too, but at this point I'm not sure would be the "best-looking" solution.

src/mainwindow.h
```diff
@@ -115,6 +115,7 @@ public slots:
     void mouseReleaseEvent(QMouseEvent *event) Q_DECL_OVERRIDE;
     void mouseDoubleClickEvent(QMouseEvent *event) Q_DECL_OVERRIDE;
     void leaveEvent(QEvent *) Q_DECL_OVERRIDE;
+    void changeEvent(QEvent *event) Q_DECL_OVERRIDE;
```
bjorn, line 118: Note for future patch, we should replace all `Q_DECL_OVERRIDE` with just `override`, since we're relying on C++14 anyway.

src/mainwindow.cpp
guihkx: Right, I always forget about that. 😓

src/mainwindow.cpp
guihkx: At least on Linux, this event fires *a lot* if I don't check for spontaneity. Besides, we only really care if the window was maximized/unmaximized by the window manager.

>`event->type() == QEvent::WindowStateChange`

I tried this first, but the event never fired on GNOME upon maximizing/unmaximizing... 🤔

bjorn: I guess with firing a lot, you mean without checking for `WindowStateChange`? Because `changeEvent` covers a lot of stuff, much of which is indeed not "spontaneous", but checking the type of the event should be the preferred approach.

Which event were you seeing instead on GNOME?

guihkx: Ohh, I was doing things wrong then. I was checking for `QEvent::WindowStateChange` in `MainWindow::eventFilter()`, but it never fired there.

Alright, I'll check for that in `changeEvent()`. Thanks!

src/mainwindow.h
guihkx: Meh, that seems like an trivial change I could easily include here...

nuttyartist:
Very nice!

I noticed is as well, but didn't expect a fix that quickly. Thanks!

bjorn (approved):

Issue #496, opened 2023-02-23
Issues with Windows build for 2.1.0


nuttyartist:
I'm testing the Windows build now. While there aren't major issues, here are some points:

~1. Clicking on the Notes icon in the task manager doesn't seem to minimize it back to the task manager (although system tray works well). I forgot if it's a known problem or not.~ EDIT: doesn't seem like a problem anymore.
2. Some windows have a question mark button on their title bar:
![Screen Shot 2023-02-23 at 11 00 05 AM](https://user-images.githubusercontent.com/16375940/220871397-a7721c5c-4367-4bf2-ac84-76eef78d76f6.png)
![Screen Shot 2023-02-23 at 11 04 31 AM](https://user-images.githubusercontent.com/16375940/220871408-13c23b37-cf6a-44d2-98e4-75f0ce8e5a4a.png)
3. Many styling issues. Many buttons look extremely old. See the scroller in the Updater window above and the About window. And the tick box in the updater window. Also here in these buttons and context menu:
![Screen Shot 2023-02-23 at 11 05 26 AM](https://user-images.githubusercontent.com/16375940/220871822-dd71f67a-3b81-4ca0-a828-4d240db97856.png)
![Screen Shot 2023-02-23 at 11 09 30 AM](https://user-images.githubusercontent.com/16375940/220871831-5d97b7e2-e789-4f86-a178-7a63361398c1.png)

Is it only solved by custom styling?

BTW, I'm getting an error trying to compile Notes. I want to test a couple of things (making sure updater works, some styling things, etc...). Any idea how can I solve this?

![Screen Shot 2023-02-23 at 11 34 41 AM](https://user-images.githubusercontent.com/16375940/220871996-a9f89ecb-b046-4d64-8873-467868eb09be.png)

bjorn:
> 2. Some windows have a question mark button on their title bar:

I guess you were testing with Qt 5 builds? Since this question mark should be disabled by default in Qt 6. For Qt 5 builds, we can get rid of it by setting an application attribute:
```c++
#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
    QCoreApplication::setAttribute(Qt::AA_DisableWindowContextHelpButton);
#endif
```
(Not sure what is our minimum Qt version, but this attribute is available since Qt 5.10 - before that, the flag had to be disabled for each window separately)

[...]

PR #502, opened 2023-02-26
Windows Installer: Don't install both Qt builds

guihkx:
We were incorrectly installing both Qt 5 and Qt 6 builds on Windows 8 and newer.

The Qt 5 build should only be installed on Windows 7, or on 32-bit versions of Windows.

I also added the [`ExtraDiskSpaceRequired` option](https://jrsoftware.org/ishelp/index.php?topic=setup_extradiskspacerequired) to work around a possible bug in Inno Setup ([[1]](https://groups.google.com/g/innosetup/c/DaadHhvmCp0/m/RgPdV9CKBAAJ), [[2]](https://stackoverflow.com/a/22044538)), where it's unable to calculate the minimum required disk space to install the app, so it ends up reporting only 3.1 MB:

image

---

This was noticed by @nuttyartist in #496. Thanks!

### DIFF
--- a/packaging/windows/Notes_Inno_Script_Github.iss
+++ b/packaging/windows/Notes_Inno_Script_Github.iss
@@ -18,6 +18,9 @@ UninstallDisplayName=Notes
 OutputBaseFilename=NotesSetup_{#Version}
 Compression=lzma
 SolidCompression=yes
+; Inno Setup can't seem to calculate the required disk space correctly, and asks for only 3.1 MB.
+; Let's use an approximate value to represent the size of the app while installed on disk (70 MB).
+ExtraDiskSpaceRequired=73400320
 
 [Code]
 function GetDefaultInstallDir(Param: string): string;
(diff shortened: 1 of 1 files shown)


nuttyartist:
Great, I'll test this.

nuttyartist:
It's probably still using Qt5?

Screen Shot 2023-02-27 at 1 11 48 PM
Screen Shot 2023-02-27 at 1 12 08 PM
Screen Shot 2023-02-27 at 1 12 22 PM

BTW, not related, but I think the installer default should be to check the option for creating a desktop icon.

guihkx:
Oh, there were a bunch of copy/paste mistakes in the Inno Setup file while creating directories, that's what was causing the theming issue when the app was installed (the "portable" version was fine). 🤦

I fixed all of them now, please retest. :)

>I think the installer default should be to check the option for creating a desktop icon.

Okay, I did that too. Keep in mind that the installer remembers what options you chose in a previous installation, so if you want to test this, make sure to uninstall any previous version of Notes first.

nuttyartist:
Yay, works well now, thanks!

nuttyartist (approved):

PR #503, opened 2023-02-26
Fixes panel widths resetting when resizing (fixes #482)

### LINKED ISSUE #482: Resizing the window causes the width of the 'All Notes' panel to be reset
I'm not sure if this is supposed to be intended behavior, but after we adjust the width of the 'All Notes' panel and then resize the window, the width gets reset:

https://user-images.githubusercontent.com/626206/219869627-b73252be-9733-4877-912a-bb25b704dce7.mp4

<br>

Should be reproducible on all platforms.

Feel free to close if this is intended behavior.

zjeffer:
This PR removes the m_noteListWidth and m_nodeTreeWidth variables, as I bypass them entirely. It also adds 2 new boolean settings: isNoteListCollapsed and isTreeCollapsed.

I fixed the resizing issue by using the splitter's setStretchFactor method to 0 for the first two panes, and 1 for the editor pane. Collapsing the 2 panes happens by simply hiding the widgets, using the setHidden() method.

Tested on bspwm with a clean settings file.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -71,8 +71,6 @@ MainWindow::MainWindow(QWidget *parent)
       m_trashCounter(0),
       m_layoutMargin(10),
       m_shadowWidth(10),
-      m_noteListWidth(185),
-      m_nodeTreeWidth(185),
       m_smallEditorWidth(420),
       m_largeEditorWidth(1250),
       m_canMoveWindow(false),
@@ -94,8 +92,6 @@ MainWindow::MainWindow(QWidget *parent)
       m_chosenSerifFontIndex(0),
       m_chosenSansSerifFontIndex(0),
       m_chosenMonoFontIndex(0),
-      m_isNoteListCollapsed(false),
-      m_isTreeCollapsed(false),
       m_currentCharsLimitPerFont({ 64, // Mono    TODO: is this the proper way to initialize?
                                    80, // Serif
                                    80 }), // SansSerif
@@ -355,6 +351,12 @@ void MainWindow::setupMainWindow()
     m_textEdit = ui->textEdit;
     m_editorDateLabel = ui->editorDateLabel;
     m_splitter = ui->splitter;
+    m_foldersWidget = ui->frameLeft;
+    m_noteListWidget = ui->frameMiddle;
+    // don't resize first two panes when resizing
+    m_splitter->setStretchFactor(0, 0);
+    m_splitter->setStretchFactor(1, 0);
+    m_splitter->setStretchFactor(2, 1);
 
 #if defined(Q_OS_UNIX) && !defined(Q_OS_MACOS)
     QMargins margins(m_layoutMargin, m_layoutMargin, m_layoutMargin, m_layoutMargin);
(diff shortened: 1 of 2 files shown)


src/mainwindow.cpp
```diff
@@ -1104,6 +1106,16 @@ void MainWindow::initializeSettingsDatabase()
...
+
+    // isNoteListCollapsed
+    if (m_settingsDatabase->value(QStringLiteral("isNoteListCollapsed"), "NULL") == "NULL") {
+        m_settingsDatabase->setValue(QStringLiteral("isNoteListCollapsed"), false);
+    }
+
+    // isTreeCollapsed
+    if (m_settingsDatabase->value(QStringLiteral("isTreeCollapsed"), "NULL") == "NULL") {
+        m_settingsDatabase->setValue(QStringLiteral("isTreeCollapsed"), false);
+    }
```
bjorn, line 1118: I think we can just remove this. For bool values, `false` is the default anyway, so there should be no point in inserting false for initialization.

src/mainwindow.cpp
```diff
@@ -1273,15 +1285,17 @@ void MainWindow::restoreStates()
         m_splitter->restoreState(
                 m_settingsDatabase->value(QStringLiteral("splitterSizes")).toByteArray());
 
+    if (m_settingsDatabase->value(QStringLiteral("isTreeCollapsed"), "NULL") != "NULL") {
```
bjorn, line 1288: I think we can just remove these "NULL" checks, since the behavior will be exactly the same without them (if there is no value set, the `toBool` will return `false`.

bjorn (commented):
Nice cleanup! Just two comments for now.

guihkx:
Hiding and restoring panels using the options in the menu is a bit buggy at the moment (tested with the Qt 5 AppImage):

https://user-images.githubusercontent.com/626206/221446655-f53fdcf4-7b43-4e67-8a20-e195c2128a77.mp4

<br>

It works just fine using keyboard shortcuts, though.

bjorn:
If we're just toggling the visibility of those widgets (which I think is a great idea, btw), couldn't we also get rid of the two extra booleans, `m_isNoteListCollapsed` and `m_isTreeCollapsed` and just check their `isVisible()` directly instead?

Also, I think instead of `m_splitter->widget(0)` we can use the more readable pointers `ui->treeView` and `ui->listView` (which could be renamed to `foldersTreeView` and `noteListView` for additional clarity.

nuttyartist:
> Hiding and restoring panels using the options in the menu is a bit buggy at the moment (tested with the Qt 5 AppImage):
> 
> 
> It works just fine using keyboard shortcuts, though.

Hmm, I think we should fix this before merging...

bjorn:
> `ui->treeView` and `ui->listView` both point to widgets inside of the left and middle pane, not the left and middle pane themselves.

Ah, sorry I missed that part.

> What should I call these left and middle panes?

Maybe something like `foldersWidget` and `noteListWidget` then.

src/mainwindow.cpp
```diff
@@ -346,6 +342,12 @@ void MainWindow::setupMainWindow()
     m_textEdit = ui->textEdit;
     m_editorDateLabel = ui->editorDateLabel;
     m_splitter = ui->splitter;
+    m_foldersWidget = m_splitter->widget(0);
+    m_noteListWidget = m_splitter->widget(1);
```
bjorn, line 346: I had another look at the `mainwindow.ui`. Actually, these pointers are already available as `ui->frameLeft` and `ui->frameMiddle`. I would suggest we just use those.

image

zjeffer:
Rebased to master and we're now setting m_foldersWidget and m_noteListWidget to ui->frameLeft and ui->frameMiddle, respectively. Thanks taking the time to review.

nuttyartist:
Awesome, thanks! 👍

PR #505, opened 2023-02-28
A couple of fixes for frameless mode

### LINKED ISSUE #494: We can drag the window by grabbing on places other than the "title bar"
Another episode of "note sure if this is a bug or intended behavior" 😄 

I'm not sure if this is Linux specific, but right now in frameless mode we can grab and drag the window from places other than the "title bar":

https://user-images.githubusercontent.com/626206/220690891-613bed70-9adf-49ed-b9c8-a5b9bb530232.mp4

<br>

Does that look correct? Or should we only be able to move the window by grabbing the title bar? The part I highlighted in red here:

image

### LINKED ISSUE #508: Unable to move the window around when only the text editor widget is visible at startup
If you hide both the notes and folders panels (while in frameless mode), the next time you launch Notes, you won't be able to move the window (unless you unhide one of the panels first).

**Steps to reproduce:**

1. Enable frameless mode
2. Hide the folders panel
3. Hide the notes panel
4. Quit Notes
5. Open Notes
6. Try to move the window around

I bisected this and it points to ef4bb7f08ea84fbb6213e9fdea4fc053c58bbb23, but I haven't found a fix yet.

@zjeffer Would you have any idea?

guihkx:
**1 - Don't allow resizing the window in certain states:**

While in frameless mode, if the window is either maximized or in full screen, we probably shouldn't allow the user to resize it.

**Bug:**

https://user-images.githubusercontent.com/626206/221879465-beab71e3-ca36-4de0-9974-ea5a786d1fcd.mp4

<br>

---

**2 - Only allow moving the window by grabbing its title bar:**

Previously, while in frameless mode, we could move the window by grabbing some unusual places. Now, we copy what most window managers do and only allow moving the window by grabbing its title bar.

The height of the title bar is defined by the height of the `verticalSpacer_upTreeView` widget (currently 25 pixels).

Closes #494

**Bug:**

https://user-images.githubusercontent.com/626206/220690891-613bed70-9adf-49ed-b9c8-a5b9bb530232.mp4

---

**3 - Don't filter out mouse click events**

This filtering was added to ensure we could only drag the window by grabbing anywhere above the search widget (`ui->searchEdit`).

However, due to changes introduced by ef4bb7f08ea84fbb6213e9fdea4fc053c58bbb23, now the search widget can start hidden (by hiding the notes panel and quitting the app).

This scenario is now handled directly in `MainWindow::mousePressEvent` and `MainWindow::mouseDoubleClickEvent`, so we can remove this entirely.

Fixes #508

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -2568,14 +2568,12 @@ void MainWindow::mousePressEvent(QMouseEvent *event)
     m_mousePressY = event->pos().y();
 
     if (event->buttons() == Qt::LeftButton) {
-        if (m_mousePressX < width() - m_layoutMargin && m_mousePressX > m_layoutMargin
-            && m_mousePressY < height() - m_layoutMargin && m_mousePressY > m_layoutMargin) {
+        if (isTitleBar(m_mousePressX, m_mousePressY)) {
 
 #  if QT_VERSION >= QT_VERSION_CHECK(5, 15, 0)
             m_canMoveWindow = !window()->windowHandle()->startSystemMove();
 #  else
             m_canMoveWindow = true;
-            //            QApplication::setOverrideCursor(QCursor(Qt::ClosedHandCursor));
 #  endif
 
 #  ifndef __APPLE__
@@ -2833,8 +2831,7 @@ void MainWindow::mousePressEvent(QMouseEvent *event)
     }
 
     if (event->button() == Qt::LeftButton) {
-        if (event->pos().x() < width() - 5 && event->pos().x() > 5
-            && event->pos().y() < height() - 5 && event->pos().y() > 5) {
+        if (isTitleBar(event->pos().x(), event->pos().y())) {
 
 #  if QT_VERSION >= QT_VERSION_CHECK(5, 15, 0)
             m_canMoveWindow = !window()->windowHandle()->startSystemMove();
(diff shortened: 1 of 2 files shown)


zjeffer (approved):

nuttyartist:
Looks good 👍 

> While in frameless mode, if the window is either maximized or in full screen, we probably shouldn't allow the user to resize it.

👍

guihkx:
Sorry guys, I've made a couple of additional changes, mainly to help fix #508. So it'd be great to have your approval on this new code again.

Thanks!

zjeffer (approved):
LGTM :+1:

src/mainwindow.cpp
```diff
@@ -3751,3 +3739,25 @@ void MainWindow::setMargins(QMargins margins)
     ui->centralWidget->layout()->setContentsMargins(margins);
     m_trafficLightLayout.setGeometry(QRect(4 + margins.left(), 4 + margins.top(), 56, 16));
 }
+
+bool MainWindow::isTitleBar(const int &x, const int &y)
```
bjorn, line 3743: Just two things:

* Please don't pass an `int` (4-byte value) as `const int &` (8-byte reference to that value). It's just more typing and makes things needlessly more expensive. :-)
* This function should be `const`.

guihkx: >Please don't pass an `int` (4-byte value) as `const int &` (8-byte reference to that value)

TIL... 😅 

>This function should be `const`.

Done. Thanks!

bjorn (approved):

PR #509, opened 2023-03-04
UI: Get rid of the "context help" button

guihkx:
We currently don't do anything with it, and the button is disabled by default on Qt 6, so let's also disable it on Qt 5 builds.

This was reported in #496, and this fix is based on https://github.com/nuttyartist/notes/issues/496#issuecomment-1441486941! :)

### DIFF
--- a/src/aboutwindow.cpp
+++ b/src/aboutwindow.cpp
@@ -9,6 +9,9 @@
 AboutWindow::AboutWindow(QWidget *parent) : QDialog(parent), m_ui(new Ui::AboutWindow)
 {
     m_ui->setupUi(this);
+#if QT_VERSION < QT_VERSION_CHECK(5, 10, 0)
+    setWindowFlag(Qt::WindowContextHelpButtonHint, false);
+#endif
     setWindowTitle(tr("About") + " " + QCoreApplication::applicationName());
 
     m_ui->aboutText->setText(
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -16,6 +16,10 @@ int main(int argc, char *argv[])
     app.setApplicationName("Notes");
     app.setApplicationVersion(APP_VERSION);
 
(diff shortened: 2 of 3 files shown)


zjeffer:
I can't test it on Windows right now, but does the styleeditorwindow not have this problem?

zjeffer:
Also, `setWindowFlag(Qt::WindowContextHelpButtonHint, false);` also works on Qt6 (even thought the default is false anyway), maybe it's cleaner to drop the #ifdefs? Not sure what the convention is there.

nuttyartist:
Works well. In my testing, also the look of the UI is not old as before for some reason (it looks like the Qt6 version).

I think we can release 2.1.0 on Monday, what do you say?

guihkx:
>but does the styleeditorwindow not have this problem?

Not really... Maybe because it derives from`QDialog`?

> Also, `setWindowFlag(Qt::WindowContextHelpButtonHint, false);` also works on Qt6 (even thought the default is false anyway), maybe it's cleaner to drop the #ifdefs? Not sure what the convention is there.

Sure, I'd be fine with that. I'm also not sure what's the preferred solution is though, so I'll let @nuttyartist decide.

>Works well. In my testing, also the look of the UI is not old as before for some reason (it looks like the Qt6 version).

That was fixed in #502. The styling plugin was being installed in the wrong place.

>I think we can release 2.1.0 on Monday, what do you say?

Sounds good to me. I have some very easy fixes (mostly related to packaging), so I'll try to push them all today.

bjorn (approved):

bjorn:
> Sure, I'd be fine with that. I'm also not sure what's the preferred solution is though, so I'll let @nuttyartist decide.

I like the #ifdefs because it also clearly marks the code as ready for removal, in the event that we'd drop support for Qt older than 5.10.

guihkx:
Oh, that's also a good point. Let's keep this as-is then.

Thanks!

PR #514, opened 2023-03-06
CI: Add simple code linting workflow

guihkx:
Right now it only does two things:

- Run `clang-format` on first-party sources (and returns an error if it finds any unformatted file)
- Run `cmake-format` on `CMakeLists.txt` (and returns an error if the file is not properly formatted)

In the future, we can use this workflow to add more helpful tools, like `clang-tidy`, to help us keep the code base consistent.

Ah, I obviously had to format all files in `src/`, otherwise the CI would fail. :)

### DIFF
--- a/.github/workflows/format-lint.yml
+++ b/.github/workflows/format-lint.yml
@@ -1,0 +1,31 @@
+name: Code Linting
+
+on:
+  push:
+  pull_request:
+
+jobs:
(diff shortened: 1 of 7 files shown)


nuttyartist:
Nicee 👍

zjeffer (approved):

Issue #521, opened 2023-03-11
HiDPI bug in GUI? Misaligned buttons, very confused


Darthagnon:
![Notes_g90eHH4yJN](https://user-images.githubusercontent.com/20792748/224476709-a7c836b7-2da2-414e-9f24-9675558f993c.png)

... I have a HiDPI 1080p screen (I use [XP Explorer](https://windows10-dpi-fix.xpexplorer.com/) as a global compatibility patch on my PC), and that's what it looks like... I'm very confused by what I'm looking at.

- 2 separate hamburger settings menus: Aa and `. . .` - why not combine both into one settings menu ⚙ ?
- Random floating "Notes 📁" line
- Same thing happens with "Tags 💡"
- Editor settings window:
  - Aa settings x3, overlapping with font names, and multiple microscopic Next buttons (I'm guessing this is font settings, but it's broken for me)
  - Unlabelled +/- buttons (Trial and error: font size)
  - Unlabelled Full Width ➡⬅ buttons (uncertain purpose)
  - Rest all settings button is very prominent and large... there's not that many settings that would ever need resetting, and Reset All shouldn't be that easy to click in my opinion.

zjeffer:
Thanks for reporting.

* The HiDPI bug is a known issue, but not much progress is being made sadly. Here's an existing issue tracking it: https://github.com/nuttyartist/notes/issues/39.  
* We're thinking about implementing a complete settings menu, and replacing the Aa button with something better: https://github.com/nuttyartist/notes/issues/309
* I completely agree native window decorations look much better on Windows and should be the default, but this is something @nuttyartist has to decide.
* We indeed need to look into improving the editor settings window, but we don't really have any designers or concrete ideas/mockups for this as far as I know.
  * The arrow buttons change the margins of the text editor. Full width makes the text fill the full width of the editor window, while the arrow buttons make the margins larger or smaller.

Closing as a duplicate of #39. I'll open a new issue with your comments about the editor settings.

nuttyartist:
@Darthagnon, thanks for your feedback! It's highly appreciated.

> I completely agree native window decorations look much better on Windows and should be the default, but this is something @nuttyartist has to decide.
> Native window looks much better on Windows, too, not sure why it's not the default?
I got this feedback enough, and from my last testing I also thought about it myself. We should make native window the default, until we can draw custom native window ourselves.

> 2 separate hamburger settings menus: Aa and . . . - why not combine both into one settings menu ⚙ ?
[...]

PR #526, opened 2023-03-11
Make windows use native window decorations by default

### LINKED ISSUE #525: Make native window decoration the default on Windows
Since we get so many requests about that, and also because it just makes sense since are current custom decoration doesn't look quite right on Windows, this fix only makes sense.

I'd love one day to have custom native window decoration like we have on macOS, but until then let's use native window decoration by default.

zjeffer:
Users have reported (#521) that it would be better if Windows used native window decorations by default. This PR sets that.

(not tested yet but I assume this small change works fine)

Closes #525

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -1243,8 +1243,14 @@ void MainWindow::setupModelView()
  */
 void MainWindow::restoreStates()
 {
+#if defined(Q_OS_WIN)
+    bool nativeByDefault = true;
+#else
+    bool nativeByDefault = false;
+#endif
     setUseNativeWindowFrame(
-            m_settingsDatabase->value(QStringLiteral("useNativeWindowFrame"), false).toBool());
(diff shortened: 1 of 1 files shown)


guihkx:
In my opinion, we should not be changing options set by the user (as some may prefer custom decorations, for example).

A safer approach, in my opinion, would be using native window decorations by default when the our config file is not present (i.e., it's the first time the program is running).

zjeffer:
> A safer approach, in my opinion, would be using native window decorations by default when our config file is not present 

That's what this PR does, no?

guihkx:
Oh, you're right. The second parameter of `value()` sets the default one.

In any case, I'll test this shortly.

guihkx (approved):
Works on Windows 7!

A minor styling choice nitpick: I'd use `#if defined(Q_OS_WIN)` instead of `#ifdef _WIN32`.

Other than that, LGTM.

zjeffer:
Thanks for testing!

PR #549, opened 2023-03-17
Updated qmarkdowntextedit submodule

bjorn:
Notable changes:

* Improve Home/End key handling under macOS
* Add support for bold and italic together in the note editor

However, if we look at the before and after images, we see that our font family option is no longer being taken into account by the bold and italic sections:

Before:
Screenshot from 2023-03-17 11-42-20

After:
Screenshot from 2023-03-17 11-42-40

So I'm proposing a fix to QMarkdownTextEdit first: https://github.com/pbek/qmarkdowntextedit/pull/179


bjorn:
> So I'm proposing a fix to QMarkdownTextEdit first: [pbek/qmarkdowntextedit#179](https://github.com/pbek/qmarkdowntextedit/pull/179)

Merged, so I've updated submodule. :-)

guihkx (approved):

bjorn:
Maybe somebody could confirm whether the Home/End key handling is really better under macOS?

nuttyartist:
Nice!

> Maybe somebody could confirm whether the Home/End key handling is really better under macOS? Though I'm not sure what it is about exactly, it might depend on user configuration different from macOS default.

I'm not sure what changed here. Pressing `Command-Left Arrow or Right Arrow` and `Fn–Left Arrow or Right Arrow` seems to work the same way.

zjeffer (approved):

nuttyartist (approved):

bjorn:
> I'm not sure what changed here. Pressing `Command-Left Arrow or Right Arrow` and `Fn–Left Arrow or Right Arrow` seems to work the same way.

Yeah, I think this fix was for people that changed the default behavior, as in [#1593](https://github.com/pbek/QOwnNotes/issues/1593).

PR #551, opened 2023-03-18
CI: Create RPM packages for Fedora 36 and openSUSE Leap 15

guihkx:
We now create RPM packages for the following Linux distributions:

- Fedora 36
- OpenSUSE Leap 15

The rationale for choosing those specific versions is simple: They're the oldest, yet still supported releases.

We also tell CMake to build a position independent executable ([PIE](https://cmake.org/cmake/help/latest/prop_tgt/POSITION_INDEPENDENT_CODE.html)) to get rid of this warning given by rpmlint:

```
notes.x86_64: W: position-independent-executable-suggested /usr/bin/notes
```

Likewise with the DEB package, there's also some other warnings related to missing documentation, but those can be addressed in the future.

---

P.S.: There's an unrelated change in this PR: I've bumped our C++ standard to C++17, which Qt 6 already asks for anyway. And even though we still support Qt 5, this change shouldn't be too much of an issue, because it's possible to install a C++17-compatible compiler on older platforms we still support.


guihkx:
Changes since my last `push`:

- CI: Replaced `build-essential` (group of packages) installation by just `gcc`, which is the only package we actually need
- CI: Prevent the installation of an unused package on openSUSE
- Documentation: Updated build steps on Linux (added more distros, removed unused packages, added helpful footnotes)
- Documentation: Added a 'Packaging' section to the Linux documentation, briefly explaining how to use CPack.

Link to the rendered documentation: https://github.com/nuttyartist/notes/blob/95f335bb957405db3e54b85763c65427da5f1252/docs/build_on_linux.md

zjeffer (approved):

nuttyartist:
Awesome stuff.

I think we should explain to people using what commands they can install the required dependencies. The guide goes into good detail on how to build and install notes, but maybe they'll get stuck at the first step.

guihkx:
> I think we should explain to people using what commands they can install the required dependencies

I don't know... In my opinion, building a program from source is already a very advanced topic, so people who decide to actually do it should, at the very least, be familiar with installing packages on their distro.

I think documenting how to install the deb/rpm packages we offer is valid, though. But I think this should be added on a different document, maybe:

- `docs/install_on_windows.md`
- `docs/install_on_macos.md`
- `docs/install_on_linux.md`

Thoughts?

zjeffer:
I like that idea, will be useful especially for Linux: we could list the official package for every package manager/distro we support.

nuttyartist:
Sounds good 👍

guihkx:
Actually, let me merge this now and I'll address the documentation about installation in a separate PR, which will probably be big...

PR #552, opened 2023-03-18
Fixed ctrl+hover over links not changing cursor

### LINKED ISSUE #534: Pressing control changes cursor icon even if the cursor is not hovering on a link
The cursor icon should change only upon hovering a link. This bug causes the cursor icon to change every time control is pressed.

test

zjeffer:
Fixes #534

https://user-images.githubusercontent.com/4633209/226137306-3a0abc71-2d87-44d5-bc1a-8281009cd834.mp4

### DIFF
--- a/src/customdocument.cpp
+++ b/src/customdocument.cpp
@@ -2,11 +2,14 @@
 #include <QDebug>
 #include <QGuiApplication>
 #include <QTextCursor>
+#include <QMessageBox>
 
 CustomDocument::CustomDocument(QWidget *parent) : QTextEdit(parent)
 {
     installEventFilter(this);
     viewport()->installEventFilter(this);
+
+    setAttribute(Qt::WidgetAttribute::WA_Hover, true);
 }
 
 /*!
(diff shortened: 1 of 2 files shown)


src/customdocument.cpp
```diff
@@ -154,13 +165,20 @@ bool CustomDocument::openLinkAtCursorPosition()
         if (isFileUrl) {
             QString trimmed = urlString.mid(7);
             if (!QFile::exists(trimmed)) {
-                qDebug() << __func__ << "file does not exist:" << urlString;
+                qDebug() << __func__ << ": File does not exist:" << urlString;
+                // show a message box
+                QMessageBox::warning(
```
zjeffer, line 170: I'm showing a messagebox here, because before it might not have been clear to the user that the file doesn't exist.

src/customdocument.cpp
```diff
@@ -103,7 +106,7 @@ QString CustomDocument::getMarkdownUrlAtPosition(const QString &text, int positi
             const int foundPositionEnd = foundPositionStart + linkText.size();
 
             // check if position is in found string range
-            if ((position >= foundPositionStart) && (position <= foundPositionEnd)) {
+            if ((position >= foundPositionStart) && (position < foundPositionEnd)) {
```
zjeffer, line 109: This fixes a bug where if a link is the last thing in a line, all whitespace to the right of the link will be clickable. Adding a space to the right of the link would be a temporary fix.

Maybe there's a better way to fix this?

src/customdocument.cpp
```diff
@@ -2,11 +2,14 @@
...
 #include <QGuiApplication>
 #include <QTextCursor>
+#include <QMessageBox>
 
 CustomDocument::CustomDocument(QWidget *parent) : QTextEdit(parent)
 {
     installEventFilter(this);
     viewport()->installEventFilter(this);
+
+    setAttribute(Qt::WidgetAttribute::WA_Hover, true);
```
zjeffer, line 12: This needed to  be enabled for QEvent::HoverMove events to work.

src/customdocument.cpp
zjeffer: Maybe we should provide a custom style for messageboxes (in a separate PR)?

nuttyartist: I think we can stick with native one hah (less headache, and it should look good, at least on macOS it looks good).
But I I think we don't need to provide the name of the file. Just "This file doesn't exist." should be good enough, the link is already shown to the user in front of him.

nuttyartist:
Other than the one change I suggested, I approve.

src/customdocument.cpp
zjeffer: I think it's useful to have the filename in the messagebox, because if something goes wrong with the link grabbing (for instance, maybe a bug could occur where only a part of the link is extracted), then the user would easily see why it went wrong. Or maybe the user might think they clicked the wrong link by accident and now gets confirmation for which link they clicked.

Or we could just show the clicked link in the terminal logs (like we do now) and I can change the messagebox text.
Your call.

nuttyartist: Ah, well that makes sense. I tried to open a very long link that didn't work (on purpose) and it showed a very big message box, this is why I said it. But then, I agree with your approach. 👍

PR #558, opened 2023-03-24
Workaround to fix #531

### LINKED ISSUE #531: folders tree not showing up at all (elementary os)
the app was working great yesterday, but when i started up this morning the folders tree wasn't showing up at all, no matter how much i clicked the icon. i opened the settings and it appears that the app is indeed showing and hiding the folders, it's just that the ui simply doesn't reflect that. i restarted my computer and the problem persisted. please help! i love the app in general and i hope this is easily fixable :)

it's stuck

zjeffer:
This is kind of a dirty workaround to fix #531. 

I have no idea how this bug happened in the first place. Somehow the issue creator had a settings file with the splittersizes line having a 0 in the QByteArray. 

I'm interested to hear your opinions whether we should merge this in the first place. I have no idea if this issue will come up again, and the fix is kind of dirty.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -1282,9 +1282,15 @@ void MainWindow::restoreStates()
     m_splitter->setCollapsible(1, true);
     m_splitter->resize(width() - m_layoutMargin, height() - m_layoutMargin);
 
-    if (m_settingsDatabase->value(QStringLiteral("splitterSizes"), "NULL") != "NULL")
+    if (m_settingsDatabase->contains(QStringLiteral("splitterSizes"))) {
         m_splitter->restoreState(
                 m_settingsDatabase->value(QStringLiteral("splitterSizes")).toByteArray());
+        // in rare cases, the splitter sizes can be zero, which causes bugs (issue #531)
+        auto splitterSizes = m_splitter->sizes();
+        splitterSizes[0] = std::max(splitterSizes[0], m_foldersWidget->minimumWidth());
+        splitterSizes[1] = std::max(splitterSizes[1], m_noteListWidget->minimumWidth());
+        m_splitter->setSizes(splitterSizes);
+    }
 
     m_foldersWidget->setHidden(
             m_settingsDatabase->value(QStringLiteral("isTreeCollapsed")).toBool());


src/mainwindow.cpp
```diff
@@ -3700,7 +3714,7 @@ void MainWindow::setUseNativeWindowFrame(bool useNativeWindowFrame)
 void MainWindow::setHideToTray(bool enabled)
 {
     m_hideToTray = enabled;
-    m_settingsDatabase->setValue(QStringLiteral("hideToTrayEnabled"), enabled);
+    m_settingsDatabase->setValue(QStringLiteral("hideToTray"), enabled);
```
guihkx, line 3717: I'd put this into a different commit, as it's a change unrelated to the problem.

src/mainwindow.h
```diff
@@ -166,7 +166,6 @@ public slots:
     int m_trashCounter;
     int m_layoutMargin;
     int m_shadowWidth;
-    int m_noteListWidth;
```
guihkx, line 169: Maybe I'm a little pedantic, but I'd also put this into a different commit. :p

guihkx (commented):
I have no comments on the fix itself because it's a little cryptic to my knowledge, but I'll say it seems to have fixed the issue (I've only tested on Linux with Qt 6.4.3).

I'd still love to figure out how to manually trigger it...

src/mainwindow.cpp
zjeffer: Ok, I made 2 commits fixing these problems.

zjeffer:
> I have no comments on the fix itself because it's a little cryptic to my knowledge

Basically, the QByteArray in the settings file is saved by m_splitter->saveState(). It is a list of 4 byte integers that looks something like this: `255, 1, 3, 222, 238, 459, 16777216, 65536, 65536`. The 4th, 5th and 6th values represent the size of the splitter panes. I have no idea what the others mean. 

When one of these 3 pane sizes is 0, the bug occurs. That's why this workaround will check if one of these 3 values is 0, and if so, assign the minimum width instead (185). 

In the issue reporter's case, their QByteArray consisted of these values: `255, 1, 3, 0, 191, 1010, 16777216, 65536, 65536`.  It's unclear why their folder width (4th value) is 0. This should never happen, as the saveState takes into account the minimum width of the folders. So if a splitter size is 0 (which happens when a widget is hidden), saving the splitter state will save it as 185 instead of 0. This somehow didn't happen once.

guihkx:
Thanks for the explanation. Isn't it possible that future Qt versions change how `saveState()` works, though? Wouldn't this change essentially bite us in the back?

zjeffer:
Perhaps, which is why I'm hesitant to even merge this in the first place. It's likely the issue will never come up again and the "fix" might be prone to future bugs.

What do you think, @bjorn ?

nuttyartist (approved):

bjorn:
We can see what `restoreState` does by just looking at the Qt code:

https://github.com/qt/qtbase/blob/8570e86fff183ffa4c1dff577c7fa14e3342daee/src/widgets/widgets/qsplitter.cpp#L1704-L1728

So, we see apart from splitter size, it stores whether they are collapsible, the handle width, whether opaque resize is enabled, the orientation, etc. It's rather more than I had expected to be honest, but then Qt can't make much assumptions about what state apps may want to store.

I think we should consider what state we really want to store, and consider calling only the necessary function, for example `QSplitter:setSizes` if the sizes are the only thing we're interested in storing (which seems likely?). Then it's also easier to apply some sanity bounds if necessary, rather then messing about in the byte array (which indeed could change with Qt version, even if it's unlikely QtWidgets will change much at this point).

bjorn (changes requested):
Forgot to give review in my last comment, so here it is just to remove the banner.

Alternative to changing what we're storing in the QSettings, as previously suggested, I guess another fix would be to check if we need to correct the sizes after calling `restoreState`? Or does that not address the issue we're running into?

zjeffer:
> Alternatively to changing what we're storing in the QSettings, as previously suggested, I guess another fix would be to check if we need to correct the sizes after calling restoreState? Or does that not address the issue we're running into?

I thought I tried this before and it didn't fix the bug, but I just tried it again and it seems to work fine now. This is indeed much simpler: simply setting the size to minimum if the width is 0 **after** calling restoreState.

bjorn (approved):
Looks fine, but I had two suggestions for simplification.

src/mainwindow.cpp
```diff
@@ -1282,9 +1282,17 @@ void MainWindow::restoreStates()
     m_splitter->setCollapsible(1, true);
     m_splitter->resize(width() - m_layoutMargin, height() - m_layoutMargin);
 
-    if (m_settingsDatabase->value(QStringLiteral("splitterSizes"), "NULL") != "NULL")
+    if (m_settingsDatabase->value(QStringLiteral("splitterSizes"), "NULL") != "NULL") {
```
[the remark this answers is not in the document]
zjeffer: I wasn't aware of the existance of that contains function, looks much better indeed.

src/mainwindow.cpp
```diff
@@ -1282,9 +1282,17 @@ void MainWindow::restoreStates()
...
         m_splitter->restoreState(
                 m_settingsDatabase->value(QStringLiteral("splitterSizes")).toByteArray());
+        // in rare cases, the splitter sizes can be zero, which causes bugs (issue #531)
+        auto splitterSizes = m_splitter->sizes();
+        splitterSizes[0] =
+                splitterSizes[0] == 0 ? m_foldersWidget->minimumWidth() : splitterSizes[0];
```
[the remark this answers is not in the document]
zjeffer: > Or do we need to keep sizes between 0 and the minimum size for some reason?

No, your change is better.

PR #566, opened 2023-04-04
Added alt+up/down shortcut to move current line

zjeffer:
This is something I've been meaning to add for a while, and it seems to work so far:

https://user-images.githubusercontent.com/4633209/229840697-82738395-0e3d-4f69-a398-99ca82f4af67.mp4

Basically, pressing alt + arrow up or down will move the current line up/down one line. It's especially useful when moving things around in a list of items. I use this feature a lot in VSCode, which is why I wanted to implement it here as well.

What do you guys think? Is this something useful for Notes? The implementation is kind of ugly but I don't see a better way to implement it, as the editor is kind of limited.

### DIFF
--- a/src/customdocument.cpp
+++ b/src/customdocument.cpp
@@ -53,6 +53,15 @@ bool CustomDocument::eventFilter(QObject *obj, QEvent *event)
             // check if mouse is over a link
             auto url = getUrlUnderMouse();
             viewport()->setCursor(url.isValid() ? Qt::PointingHandCursor : Qt::IBeamCursor);
+        } else if (keyEvent->modifiers().testFlag(Qt::AltModifier)) {
+            // alt + arrow up/down
+            if (keyEvent->key() == Qt::Key_Up) {
+                moveBlockUp();
+                return true;
+            } else if (keyEvent->key() == Qt::Key_Down) {
+                moveBlockDown();
+                return true;
+            }
         }
     } else if (event->type() == QEvent::MouseButtonRelease) {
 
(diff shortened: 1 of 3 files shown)


zjeffer:
I just noticed it has a bug when working with multiline blocks of text (because of word wrapping), needs to be fixed.

nuttyartist:
That's awesome! Of course it's useful, I, and am sure many other people use lists often. And yes, it works well but I also experience the multiline problem.

src/customdocument.cpp
```diff
@@ -293,3 +302,75 @@ QMap<QString, QString> CustomDocument::parseMarkdownUrlsFromText(const QString &
...
+        qDebug() << "Current horizontal position: " << currentHorizontalPosition;
+
+        cursor.movePosition(QTextCursor::StartOfBlock);
+        cursor.movePosition(QTextCursor::EndOfBlock, QTextCursor::KeepAnchor);
+        cursor.removeSelectedText();
+        // also remove empty line
+        cursor.deletePreviousChar();
+
+        // cursor is now at end of prev line, move to start
+        if (!cursor.movePosition(QTextCursor::StartOfLine)) {
```
nuttyartist, line 323: Chaing these lines to the following solved the issue for me. Can you test?

```cpp
// cursor is now at end of prev line, move to start of block
        if (!cursor.movePosition(QTextCursor::StartOfBlock)) {
            qDebug() << "Could not move to start of block";
        }
```

zjeffer: Yes! That fixed it. Thanks!

nuttyartist: Great. Did enough QTextCursor manipulations with the kanban feature to see the issue haha.

zjeffer:
I added the shortcut to the keyboard_shortcuts.md file. I also removed the unnecessary qDebug() statements.

nuttyartist:
Perfect. I'll need to remember to update the website shortcut keys upon release. Thanks! Merging...

PR #567, opened 2023-04-04
Overhaul styling: use .css files to set theme dynamically using CSS classes

zjeffer:
I'm looking into replacing all `setStyleSheet` calls to use .css files instead, and change the widget's theme by simply changing the css classes it has. This will make it much easier to change the styling of the widgets. We should try to move all color declarations to a css file. Lots of widgets have these colors hardcoded in their constructor.

* [x] About window
* [x] Style Editor window
* [ ] Main window
* [ ] Folder tree
* [ ] Tag tree

This will also make the mainwindow.cpp's setTheme() method much cleaner.


nuttyartist:
That's great, I'll test this very soon. This should allow us to iterate on styling much faster make things more organized.

src/mainwindow.cpp
```diff
@@ -593,13 +606,14 @@ void MainWindow::setupLine()
...
     ui->frameRightTop->installEventFilter(this);
+    setCSSThemeAndUpdate(ui->frameRight, m_currentTheme);
 
-    QString ss = QStringLiteral("QFrame{ "
-                                "  background-color: %1; "
-                                "  border: none;"
-                                "}")
-                         .arg(m_currentRightFrameColor.name());
-    ui->frameRight->setStyleSheet(ss);
+    // QString ss = QStringLiteral("QFrame{ "
```
zjeffer, line 611: Something strange going on over here: when I comment this block of code, the application segfaults when resizing. Specifically, it segfaults when calling `setLineWrapColumnOrWidth` (line 911) in current revision, which doesn't make much sense to me. My setCSSThemeAndUpdate function (which is what replaces the commented code) seems to work fine and changes the background color and border.

I don't get how this segfault is related to setting the stylesheet manually like how it's being done in the commented code. Any ideas?

zjeffer: Seems to only happen on Qt5 :(

nuttyartist:
This is a very welcome change. During my testing I found these issues:

The lines separating actions in the styleeditor here don't go far enough (more visible in dark mode):

Screen Shot 2023-04-08 at 7 28 42 PM

Screen Shot 2023-04-08 at 7 28 30 PM

Screen Shot 2023-04-08 at 7 28 35 PM

There's an unnecessary black rectangle in the Light theme:

Screen Shot 2023-04-08 at 7 29 07 PM

But I like it during Dark theme (good separation). BUT, it needs to disappear when the buttons currently disappear - when the user types something or scrolls. 

Screen Shot 2023-04-08 at 7 29 38 PM

We still have this issue where some areas during Dark theme (only then) not having the proper background color.

Screen Shot 2023-04-08 at 7 29 57 PM
Screen Shot 2023-04-08 at 7 31 04 PM

zjeffer:
@nuttyartist In the latest commit, I don't have those issues anymore, but I never had  the problem with the lines in the style editor not going far enough. Is it possible it looks differently on macOS than on Linux?

src/nodetreeview.cpp
```diff
@@ -16,38 +17,10 @@ NodeTreeView::NodeTreeView(QWidget *parent)
...
-            R"(QScrollBar::handle:vertical:hover { background: rgb(170, 170, 171); } )"
-            R"(QScrollBar::handle:vertical:pressed { background: rgb(149, 149, 149); } )"
-            R"(QScrollBar::handle:vertical { border-radius: 4px; background: rgb(188, 188, 188); min-height: 20px; }  )"
-            R"(QScrollBar::vertical {border-radius: 4px; width: 8px; color: rgba(255, 255, 255,0);} )"
-            R"(QScrollBar {margin: 0; background: transparent;} )"
-            R"(QScrollBar:hover { background-color: rgb(217, 217, 217);})"
-            R"(QScrollBar::add-line:vertical { width:0px; height: 0px; subcontrol-position: bottom; subcontrol-origin: margin; }  )"
-            R"(QScrollBar::sub-line:vertical { width:0px; height: 0px; subcontrol-position: top; subcontrol-origin: margin; })");
-#else
-    setStyleSheet(R"(QTreeView {)"
```
zjeffer, line 40: In the latest commit, I converted the NodeTreeView to also use my  `setCSSThemeAndUpdate` method, but I noticed that the CSS for macOS was different than the default. Why? Do you notice a difference in the latest build?

zjeffer:
Strange, I don't have half the issues you're having. Looks like I'm going to have to pull up my macOS VM...

Also needs to be tested on Windows.

zjeffer:
It looks even worse on my macOS VM: 

image

Not sure why this is happening...

zjeffer:
New test on macOS:

https://user-images.githubusercontent.com/4633209/232306921-e279ec07-4f38-4c37-b0b7-ddc83bc0e1b7.mp4

@nuttyartist Can you test again? Everything looks fine for me, except for the little arrow to change fonts in the StyleEditorWindow, but that seems to be an issue in the current version of Notes anyway, not something related to this PR...

zjeffer:
Can confirm it works on Windows as well.

nuttyartist:
Definitely better,

1. We should use native scroll bars on macOS for the folders and notes lists.
2. Only in Sepia there's this color trailing behind selected sub folders:
Window active:
Screen Shot 2023-04-16 at 5 13 35 PM
Window not active:
Screen Shot 2023-04-16 at 5 09 58 PM
3. The lines in the style editor still look like this for some reason:

Screen Shot 2023-04-16 at 5 10 06 PM

Screen Shot 2023-04-16 at 5 10 31 PM

nuttyartist:
1 & 2 are solved.

zjeffer:
Yep, currently trying to find the cause of the third issue. It's also present on Linux & Windows.

nuttyartist:
Let me know when to test

zjeffer:
All issues should be fixed now (except for the arrow position in the font buttons, but that's for another PR)

nuttyartist:
It's the same on my machine... But the arrows look good.

Screen Shot 2023-04-16 at 8 41 45 PM

zjeffer:
Even if you resize the Style editor window?

nuttyartist:
Yeah, maybe I didn't clone properly? 

I used `it clone --recursive --single-branch --branch feature/zjeffer/css-changes https://github.com/nuttyartist/notes.git`

zjeffer:
Strange, the arrow bug occurs on 2.1.0 and on my own branch.

zjeffer:
So I guess this can be merged?

nuttyartist:
Well, I don't get why the lines still look like this. The vertical size policy is expanding but they are not expanding...

nuttyartist:
If I simply do:

Screen Shot 2023-04-17 at 11 42 24 AM

It is set correctly:

Screen Shot 2023-04-17 at 11 42 07 AM

nuttyartist:
Okay I figured it.

This

```css
StyleEditorWindow QPushButton.btn {
	padding: 0px;
	border: none;
}
```

Should be:

```css
StyleEditorWindow QPushButton {
	padding: 0px;
	border: none;
}
```

Now it works.

zjeffer:
@nuttyartist Can you test the latest commit on your mac? I simplified the CSS a bit and implemented your fix.

zjeffer:
Alright, rebased and merging...

PR #569, opened 2023-04-10
Make toggling fullscreen possible with F11 & Esc

### LINKED ISSUE #568: CTRL+SHIFT+(+) and F11 doesn't work
I try to increase the size of the header, and after that, I try to exit fullscreen only to realize F11 and Esc don't work at all.
https://user-images.githubusercontent.com/40272709/230957973-87fa16e5-c2e8-424a-9d28-4e543489056f.mp4

EDIT: The scrolling suddenly started working after I turned the native window frame off and on again.

zjeffer:
(Partly) fixes #568


guihkx:
In addition to using <kbd>Esc</kbd> to leave full screen mode, I'm thinking it would probably make sense to <kbd>F11</kbd>, no?

Issue #563, opened 2023-03-26
A view that converts all TODOs into a Kanban


nuttyartist:
Say you have a note with 3 headers, each as its own TODOs. If the list is getting long, it can be hard to keep track of different TODOs types. This is where Kanban view is useful. We can easily scan  TODOs under headers and create Kanban. That is, remove/hide QTextEdit and switch to a kanban view back and forth.

Example:

QTextEdit:

```
This is some text we'll ignore

## TODO
- [ ] TODO 1
- [ ] TODO 2
- [ ] TODO 3

## In Progress
- [ ] In Progress 1
- [ ] In Progress 2
- [ ] In Progress 3

Another text we'll ignore

## Done
- [x] Done 1
- [x] Done 2
- [x] Done 3
```

Will be converted into:

| TODO        | In Progress           | Done  |
| ------------- |:-------------:| -----:|
| TODO 1      | In Progress 1 | Done 1|
| TODO 2     | In Progress 2      |   Done 2 |
| TODO 3 | In Progress 3      |    Done 3 |

Maybe we can use an existing Qt open-source project? Or maybe we should implement this ourselves using QML (easier to create drag and drop?)

nuttyartist:
Done in #574.

Issue #523, opened 2023-03-11
Editor settings improvements


zjeffer:
@darthagnon reported (https://github.com/nuttyartist/notes/issues/521) some issues with the editor settings window:

  - Aa settings x3, overlapping with font names, and multiple microscopic Next buttons (I'm guessing this is font settings, but it's broken for me)
  - Unlabelled +/- buttons (Trial and error: font size)
  - Unlabelled Full Width ➡⬅ buttons (uncertain purpose)
  - Reset all settings button is very prominent and large... there's not that many settings that would ever need resetting, and Reset All shouldn't be that easy to click in my opinion.

Darthagnon:
I'd also suggest merging the Editor Settings with the program settings `. . .` hamburger menu, under a Gear or hamburger icon.

zjeffer:
Proposals:

* Make the editor margins a slider instead of 3 buttons (https://doc.qt.io/qt-6/qslider.html)
* Make the font a spin box (https://doc.qt.io/qt-6/qspinbox.html)
* Make the font selection a Font Combo box (https://doc.qt.io/qt-6/qfontcombobox.html), but we would need to find a way to make it clear which type of font (sans, serif, mono) each font belongs to. A simple list of fonts in a combo box doesn't really show that. 
* Instead of light/dark/sepia, we show a label "Theme: " with the actual colors as buttons.
* Move all  these settings to a separate settings window, accessible with the gear icon.

What do you guys think?

nuttyartist:
> Make the editor margins a slider instead of 3 buttons (https://doc.qt.io/qt-6/qslider.html)

I don't like this. I hate sliders when you want preciosion. The buttons allow you to change the margins by one character each time.

> Make the font a spin box (https://doc.qt.io/qt-6/qspinbox.html)

Hmm make sense for a font. I like the current way because it shows you how the fonts look like. So either we use selection using a spinbox and keep the preview, or we improve the button to be more clickable?

> Instead of light/dark/sepia, we show a label "Theme: " with the actual colors as buttons.

That makes sense. I'll try to see how it will look if I incorporate that into the existing mockup.

> Move all these settings to a separate settings window, accessible with the gear icon. Move other relevant settings (like native window decorations) to this window as well.

That's a good idea. We can move more settings there.
- Start automatically.
- Hide to tray
What do you think?

I created a mockup. It's very much rough. 

- We probably are better off using the Tab widget than implementing a custom one?
[...]

PR #574, opened 2023-04-16
Create kanban view based on editor's text, solves #563

nuttyartist:
This PR creates a kanban view based on the detected tasks currently in the m_textEdit (text editor).

This is what this PR looks like currently:

test
     
Remaining tasks for this PR:  
- [x] Solve Qt5 issue
- [x] Option to reverse order of tasks
- [x] Figure out why linuxdeployqt isn't including QtQuick dependencies in the final AppImage
- [x] Rearranging tasks - rearrange text tasks
- [x] Rearranging columns - rearrange text from headers
- [x] Make sure reverse order works
- [x] Check/uncheck an item
- [x] Option to edit tasks' text
- [x] Option to add or remove tasks
- [x] Option to add/remove columns
- [x] Allow read-only mode
- [x] Don't allow editing when selecting multiple notes
- [x] Create a shortcut key to show/hide kanban
- [x] Dragging a task further from its position to scroll the task view can cause it to be destroyed due to window/virtual rendering
- [x] Complete architecural makeover. Focuses on changing the models and let Qt automatically handle the views by itself as much as possible. Also, doesn't need to re-render on every change, allows for a much smoother experience.
- [x] Switched to required properites rather than "model." which has many benefits:
https://discuss.kde.org/t/qml-tips-use-required-property-in-delegate-instead-of-relying-on-the-model-variable/135
- [x] Replaced using setText with QTextCursor so it's possible to use Ctrl-z/Ctrl-shift-z for ANY change that happens inside the kanban view when switching back to text mode.
- [x] The GUI is enough to create everything the user expects (columns/tasks, creating/adding/editing). Ofcourse, any changes in the underline text will be reflected in the kanban GUI.
- [x] Better dark mode
- [x] Progress bars
- [x] Read only mode
- [x] Cool animations with almost any interaction (apart from columns creation/deletion/rearrangement)
- [x] Checkbox for each task
- [x] Support Markdown **inside** all tasks
- [x] Kanban font changes based on the selected sans-serif font in editor settings
- [x] Borderless columns give more modern look. Option to show borders.
- [x] Many qmllint issues are gone (many are left but mostly related to the same problem I'm unsure how to fix).
- [x] Document `--qmldir` flag from `windeployqt` in `docs/build_on_windows.md`
- [x] Document `-qmldir` flag from `macdeployqt` in `docs/build_on_macos.md`
- [x] Document new Linux dependencies in `docs/build_on_linux.md`
- [ ] Add proper button in the new editor stylesheet button (tracked at issue #523)

Code Improvements::
- [ ] Replace `setContextPropery` with `qmlRegisterSingletonType`
	 https://raymii.org/s/articles/Qt_QML_Integrate_Cpp_with_QML_and_why_ContextProperties_are_bad.html#toc_4
- [ ] How to pass to delegates other objects aliases without getting Unqualified Access warning by qmllint? (Should make QML better compile to C++)
- [ ] Other qmllint warnings
- [ ] Should we migrate Qt Quick compononents to use native macOS and Windows styles (Although some look awful so we need to check) - Dialog (looks bad on macOS), CheckBox, etc...
- [ ] Figure how to render Pane componenets with rounded corners while keeping the Material elevation shadow.
- [ ] Replace programmatically drawn icons with font icons (As per documentation here https://doc.qt.io/qt-6/qtquickcontrols-texteditor-example.html Also a nice tool: https://github.com/mnesarco/ff-batch)

Known issues:
- [ ] Rearranging columns don't work from a certain height.

Future features (I'll open a new issue for this later):
- [ ] Option to view one task/card - opens the task in a big popup so users can focus on the task at hand.
- [ ] Keyboard support
- [ ] Tooltips for custom buttons (we'll need to create them from C++ since QQuickView can't render outside its box)
- [ ] Remembering what view each note is shown (text/kanban) via the database
- [ ] Make the kanban optionally on the right/left/bottom/up of TextEdit
- [ ] Tasks vertical scrollbar that shows only when scrolling
- [ ] Support sub tasks with collapsable feature
- [ ] When we implement synchrnoise blocks we could show in the kanban tasks from different notes


src/mainwindow.cpp
```diff
@@ -1895,6 +1924,11 @@ void MainWindow::setTheme(Theme theme)
     m_currentTheme = theme;
     switch (theme) {
     case Theme::Light: {
+        QJsonObject themeData {
```
zjeffer, line 1927: I don't like this way of changing themes, but I'll add this to my CSS themes PR when this is merged (or if my PR is merged first, I'll look into updating it here as well)

nuttyartist: What do you suggest?
Should I just send the theme type and in the .qml files read the appropriate properties from your .css files? (if it's possible)

nuttyartist: This is merely to signal to the .qml files that the theme changed with the data (type, colors...).

zjeffer: I  don't really know much about qml, but if that's possible that would be great. I think we should try putting all CSS in the same place.

nuttyartist: Maybe it's enough to just change the values I set in the lines you highlighted to use your .css file values and the .qml will just pick it up from there (therefore, no hard-coded styles in .qml)

zjeffer:
> First of, excuse my bad practice of not creating different commits during development to track changes.

Where I work, we create 1 commit per PR (or multiple commits if a PR contains multiple separate features), so I don't mind this at all. I don't know if there's a universal "best way" to do this.

src/noteeditorlogic.cpp
```diff
@@ -162,13 +194,202 @@ void NoteEditorLogic::onTextEditTextChanged()
...
+// }
+// ])
+void NoteEditorLogic::checkForTasksInEditor()
+{
+    QStringList lines = m_textEdit->toPlainText().split("\n");
+    QJsonArray data;
+    QJsonObject currentColumn;
+    QJsonArray tasks;
+    QString currentTitle = "Untitled";
+    QRegularExpression taskRegex("^\\s*-\\s*\\[(x| )\\]\\s*(.*)$");
```
zjeffer, line 299: This doesn't work for something like this:

```
* [ ] TODO 4
```

But instead of using regex, maybe there's a better way to check whether the line is a list item?

nuttyartist: Right, maybe regex is overkill here. That was ChatGPT's idea hah. Maybe a simple `find('- [ ]") / '- [x]")` should do?

Ah I didn't know that's another way people use that. It should be easy to support it.

zjeffer: Sure, but I think it would be more robust if this logic is tied together with the markdown highlighter. The user gets immediate feedback to know if it's a list item (because the color changes) and therefore knows for sure that it will show up in the kanban view. 

If something goes wrong with the regex for any reason (like the above bug), the user has to guess why it went wrong.

nuttyartist: Gotcha. Yes, sounds better.

nuttyartist: Now supporting other types of task expressions and dropping regex via https://github.com/nuttyartist/notes/pull/574/commits/406c6d2c47128203a78420899acf5d43fa8b4463

zjeffer:
Running the latest commit's AppImage (qt6) throws the following errors on startup:

```
qrc:/qt/qml/kanbanMain.qml:3:1: module "QtQuick.Controls.Material" is not installed 
     import QtQuick.Controls.Material 
     ^
qrc:/qt/qml/kanbanMain.qml:2:1: module "QtQuick.Controls" is not installed 
     import QtQuick.Controls 
     ^
qrc:/qt/qml/kanbanMain.qml: module "QtQml.WorkerScript" is not installed
qrc:/qt/qml/kanbanMain.qml:3:1: module "QtQuick.Controls.Material" is not installed 
     import QtQuick.Controls.Material 
     ^
qrc:/qt/qml/kanbanMain.qml:2:1: module "QtQuick.Controls" is not installed 
     import QtQuick.Controls 
     ^
qrc:/qt/qml/kanbanMain.qml: module "QtQml.WorkerScript" is not installed
qrc:/qt/qml/kanbanMain.qml:3:1: module "QtQuick.Controls.Material" is not installed 
     import QtQuick.Controls.Material 
     ^
qrc:/qt/qml/kanbanMain.qml:2:1: module "QtQuick.Controls" is not installed 
     import QtQuick.Controls 
     ^
qrc:/qt/qml/kanbanMain.qml: module "QtQml.WorkerScript" is not installed
```

and the entire editor window is white when clicking the Switch button.

guihkx:
Yeah, I'm still trying to figure out why `linuxdeploy` isn't including those automatically...

-------- Original Message --------

nuttyartist:
@guihkx @zjeffer So you both can't test the current branch?

zjeffer:
> @guihkx @zjeffer So you both can't test the current branch?

I think I can by compiling it myself but I didn't have time yesterday.

I might try it out tomorrow.

guihkx:
Okay, I believe I have fixed the AppImage issue on my last push, so please retest when possible.

@nuttyartist Unrelated to the AppImage itself, I'm getting this message on console whenever I enable the Kanban view:

```
<Unknown File>: columnStartLine is undefined. Adding an object with a undefined member does not create a role for it.
```

guihkx:
Great, now the openSUSE job is failing randomly. :P

It's probably temporary and most definitely unrelated to the change I just pushed, so please ignore that for now.

nuttyartist:
> @guihkx @zjeffer So you both can't test the current branch?

Just to clarify, I mean to ask if you are getting any errors and therefore unable to test (it was not some passive-aggressive question on why you're not testing haha).

nuttyartist:
> Okay, I believe I have fixed the AppImage issue on my last push, so please retest when possible.
> 
> @nuttyartist Unrelated to the AppImage itself, I'm getting this message on console whenever I enable the Kanban view:
> 
> ```
> <Unknown File>: columnStartLine is undefined. Adding an object with a undefined member does not create a role for it.
> ```

Weird, let's see if it still persists after I push some stuff.

guihkx:
>Just to clarify, I mean to ask if you are getting any errors and therefore unable to test (it was not some passive-aggressive question on why you're not testing haha).

Too late, I'm already offended. :(

Seriously though, just a reminder to myself that I (or we) should also check if the Windows installer (and the portable version too) are also including the necessary qml files.

I'll leave that one for later though, it's quite late already. :)

guihkx:
Pushed some fixes for the Windows build. Did a quick test and it everything seems to work now!

My macOS VM is borked again, so I won't be able to test the dmg until I fix it again... But it'd be great to get a confirmation that it works (especially the Qt 6 one, of course).

One small visibility issue I noticed while testing the AppImage: With the dark theme enabled, after you mark a specific task as completed, its text becomes a bit of hard to read:

image

The 'switch' button isn't themed as well.

nuttyartist:
I'll try to improve the colors and aesthetics later, but thanks for the heads-up.
I already deleted the Switch button and replaced it with a shortcut. It's CTRL+SHIFT+K, but maybe you can come up with a better one? I'll add a proper button once I take care of https://github.com/nuttyartist/notes/issues/523.

guihkx:
Ctrl+Shift+K seems fine to me. :)

We just have to make sure to document it in `docs/keyboard_shortcuts.md`.

nuttyartist:
So here are the changes of the latest commit (I hope I didn't break anything with my push, but everything looks fine):

- Complete architecural makeover. Focuses on changing the models and let Qt automatically handle the views by itself as much as possible. Also, doesn't need to re-render on every change, allows for a much smoother experience.
- Switched to required properites rather than "model." which has many benefits:
https://discuss.kde.org/t/qml-tips-use-required-property-in-delegate-instead-of-relying-on-the-model-variable/135
- Replaced using setText with QTextCursor so it's possible to use Ctrl-z/Ctrl-shift-z for ANY change that happens inside the kanban view when switching back to text mode.
- The GUI is enough to create everything the user expects (columns/tasks, creating/adding/editing). Ofcourse, any changes in the underline text will be reflected in the kanban GUI.
- Better dark mode
- Progress bars
- Read only mode
- Cool animations with almost any interaction (apart from columns creation/deletion/rearrangement)
- Checkbox for each task
- Support Markdown **inside** all tasks
- Kanban font changes based on the selected sans-serif font in editor settings
- Borderless columns give more modern look. Option to show borders.
- Many qmllint issues are gone (many are left but mostly related to the same problem I'm unsure how to fix).

Here's a GIF:

test

nuttyartist:
@zjeffer @guihkx @bjorn We should test this carefully to make sure there's no data loss as much as we can.

In my testing everything was fine, you should test this as well before we merge (on temporary notes, I would suggest).

zjeffer:
I think we should change the cursor to a hand when we hover over the grabbable area of the list items. I had a hard time figuring out how to move them because I couldn't find the grabbable area. And during the grab, the cursor should change as well, to the "closed fist" cursor.

nuttyartist:
I like both your suggestions, I will implement them. Thanks! Keep it coming.

nuttyartist:
Can you guys report if the app CPU usage gone up on idle in this PR? I see quite an increase: from 0.0/0.1 to 0.9 when the kanban is hidden. When kanban isn't hidden it's still very high so maybe there's some profiling I need to do.

guihkx:
>Can you guys report if the app CPU usage gone up on idle after this PR?

I'll check that on Linux.

Anyway, with my last two commits I've fixed the macOS build (by invoking `macdeployqt` with `-qmldir`).

And I also added an extra step to the CI to run `qmllint` for Qt 6 builds, just so you can check the warnings easier.

If you're not able to eventually fix all warnings, I can remove it before merging the PR.

Also, before you finally merge this, I'd like to split some of my changes into smaller commits... But I'll only do that when you feel this PR is ready to merge, just to avoid making you rebase your local branch every time. :P

guihkx:
Hmm, doesn't this feature work like this anymore?

https://github.com/nuttyartist/notes/assets/626206/158f654d-b9c1-43d7-b4be-12e5c537ede9

:thinking:

nuttyartist:
> Hmm, doesn't this feature work like this anymore?
> 
>  Video_2023-05-16_12-20-53_edit.mp4 
> 
> 🤔

Weird. It works for me. I'll look into that tomorrow.

nuttyartist:
I found out the culprit for the high CPU usage using the QML Profiler. I'll fix it.

guihkx:
Last commit (1ac09272f9dd44851619a3be58681168ded6bf6e) made the font size in popups a little too big on Linux.

Before:

image

After:

image

nuttyartist:
> Last commit ([1ac0927](https://github.com/nuttyartist/notes/commit/1ac09272f9dd44851619a3be58681168ded6bf6e)) made the font size in popups a little too big on Linux.
> 
> Before:
> 
> image
> 
> After:
> 
> image

Yes, you're right. I just tested on Ubuntu and noticed it. I will revert to the previous version. I also think the columns text are too big (I willl switch from point sizes to pixels, I think).

You don't see the previous error you reported (here https://github.com/nuttyartist/notes/pull/574#issuecomment-1549889467) anymore?

nuttyartist:
> > You don't see the previous error you reported (here [#574 (comment)](https://github.com/nuttyartist/notes/pull/574#issuecomment-1549889467)) anymore?
> 
> I still do, unfortunately. If you want proper steps to reproduce:
> 
> 1. Run Notes using the terminal (so you can see the messages)
> 2. Create a new note with this content:
>    ```
>     - [ ] task 1
>     - [ ] task 2
>     - [ ] task 3
>     - [ ] task 4
>     - [ ] task 5
>    ```
> 3. Hit Ctrl+Shift+K to enter Kanban view
> 
> You should now see this warning on the terminal:
> 
> ```
> <Unknown File>: columnStartLine is undefined. Adding an object with a undefined member does not create a role for it.
> qrc:/qt/qml/kanbanMain.qml:367:27: QML Component: Cannot create delegate
> qrc:/qt/qml/TodoColumnDelegate.qml:32:5: Required property columnStartLine was not initialized
> ```
> 
> Additionally, the Kanban view itself still looks 'empty':
> 
> image

Oh odd, I can reproduce it on my Mac as well! I'll get it fixed (: Good catch.

nuttyartist:
I want to merge this. @guihkx you want to do some squashing?

guihkx:
@nuttyartist Alrighty. But first, maybe you could sync this PR's branch against `master` (there are conflicts), and squash some of your recent commits as well?

For features I don't really mind a "one big commit", so feel free to squash everything into that 'Create kanban view based on editor text' commit, and then I'll split some of my changes.

It'd also buy me a little extra time to properly document the new build flags and dependencies... :sweat_smile:

guihkx:
@nuttyartist I just pushed the changes I promised:

- Fix the list of dependencies in deb/rpm packages generated by CPack
- Document new shortcut for toggling the Kanban view (<kbd>Ctrl</kbd> + <kbd>Alt</kbd> + <kbd>K</kbd>)
- Document new required flags for `windeployqt` and `macdeployqt`
- Document new QML-related dependencies on Linux

As usual, I added some more details in the commit's description.

Sorry for the long delay! :S

nuttyartist:
Awesome work @guihkx. Thanks!

PR #575, opened 2023-04-28
CI: Verify files included in the Windows installer

guihkx:
This extra step will ensure that the list of files deployed by `windeployqt`, will actually match the list of files we're including in the installer shipped to users.

If both lists of files don't match, the build job will fail with an error.

This helps us prevent:

- Forgetting to include files/directories necessary to make the app run correctly.
- Installing files/directories in the wrong path, like it's been done in the past by me. :)

---

Aside from the change above, I've also sorted the list of files included in the Windows installer.

Now, folders must come first, followed by regular files. Both are sorted naturally.


nuttyartist (approved):
Looking good.

PR #577, opened 2023-05-16
Aesthetic changes, improve dark mode

nuttyartist:
This PR create more "white space" in the app, makes it look slicker and clearer.

Changed:

- Dark background color from rgb(30,30,30) to rgb(25,25,25)
- Text color in dark mode from rgb(204,204,204) to rgb(212,212,212)
- Number of notes in folders pane to be half the alpha of m_titleColor
- Note highlighter color to rgb(35, 52, 69)
- Change the weight and size of 
`"Folders" and  "Tags"`
`"Pinned" and "Notes`
- Spacing between tag's color and text
- Separators colors - less strong in dark mode
- Spacing in notesList - more whitespace.

Before:

Screen Shot 2023-05-16 at 7 57 14 PM

After:

Screen Shot 2023-05-16 at 7 57 38 PM

Let me know what you think, and if you have any suggestions.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -327,6 +327,10 @@ void MainWindow::setupMainWindow()
     ui->listviewLabel1->setStyleSheet(m_styleSheet);
     ui->listviewLabel2->setStyleSheet(m_styleSheet);
 
+    // splitters
+    ui->verticalSplitterLine_left->setStyleSheet(m_styleSheet);
+    ui->verticalSplitterLine_middle->setStyleSheet(m_styleSheet);
+
     // buttons
     ui->toggleTreeViewButton->setStyleSheet(m_styleSheet);
     ui->styleEditorButton->setStyleSheet(m_styleSheet);
@@ -1806,6 +1810,8 @@ void MainWindow::setTheme(Theme theme)
     setCSSThemeAndUpdate(ui->styleEditorButton, theme);
     setCSSThemeAndUpdate(ui->listviewLabel1, theme);
     setCSSThemeAndUpdate(ui->searchEdit, theme);
+    setCSSThemeAndUpdate(ui->verticalSplitterLine_left, theme);
+    setCSSThemeAndUpdate(ui->verticalSplitterLine_middle, theme);
     setCSSThemeAndUpdate(ui->frameRight, m_currentTheme);
     setCSSThemeAndUpdate(ui->frameRightTop, m_currentTheme);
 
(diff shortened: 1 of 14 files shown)


nuttyartist:
Also, I only tested this on macOS as for now.

guihkx:
I noticed that on Linux, on both Qt 5 and Qt 6 AppImages, the panel separators are now transparent whenever I switch to the dark theme:

https://github.com/nuttyartist/notes/assets/626206/ea89ca46-6967-4f60-bf90-5debf230bdc6

I'm not sure if it's clear in the video, but with the dark theme chosen, you can see part of my desktop wallpaper in the vertical separators between panels. Sepia and white themes appear fine.

nuttyartist:
> I noticed that on Linux, on both Qt 5 and Qt 6 AppImages, the panel separators become transparent whenever I switch to the dark theme:
> 
>  transparent_separators_dark_theme.mp4 
> 
> I'm not sure if it's clear in the video, but with the dark theme chosen, you can see part of my desktop wallpaper in the vertical separators between panels. Sepia and white themes appear fine.

Oh, that's weird, It's probably because in dark mode I set rgba, with an alpha. I will try find a color without alpha but it's still weird that it works like that.

https://github.com/nuttyartist/notes/pull/577/files#diff-0c315e1afa0020651893a700f06d731ef709d72fae73e3cf63dfb81666effea8R142

guihkx:
>Oh, that's weird, It's probably because in dark mode I set rgba, with an alpha.

That too, but the underlying issue is actually caused by us setting the [translucency attribute](https://doc.qt.io/qt-5/qt.html#:~:text=Qt%3A%3AWA_TranslucentBackground) on the main window, and we do that only on Linux:

https://github.com/nuttyartist/notes/blob/314394fe8a3b6ff626d3d9e901476f0ce2d0afe7/src/mainwindow.cpp#L313-L315

If I comment out that line, the issue goes away, but then our own window decorations looks awful. :P

nuttyartist:
I guess since we will change make native window the default on Linux (https://github.com/nuttyartist/notes/issues/578) this won't be much of a problem?

Still tho, can you check if you just remove the alpha from the color, does it stops being translucent?

guihkx:
>I guess since we will change make native window the default on Linux (https://github.com/nuttyartist/notes/issues/578) this won't be much of a problem?

Probably. We just have to make sure to remove/add the translucency attribute after switching between decoration modes.

>Still tho, can you check if you just remove the alpha from the color, does it stops being translucent?

It does, but it looks like this:

image

nuttyartist:
@guihkx Can you test if the problem persists after the latest commit?

guihkx:
Yep, the transparency issue is gone:

image

Overall I think it looks pretty good!

guihkx (approved):

nuttyartist:
Awesome, merging...

PR #583, opened 2023-06-10
CI: Unify build jobs with reusable workflows

guihkx:
Instead of having individual workflow runs for every git operation, we can instead make them ['reusable'](https://docs.github.com/en/actions/using-workflows/reusing-workflows) and then bundle all of them into a single, central workflow (which we will call `CI`).

Having them unified into a single workflow like that has several advantages:

- Build jobs will only run if the `Code Linting` workflow passes first: Previously, build jobs would run in parallel with the linting workflow, which is probably undesired.

- Build artifacts for every OS will be all in the same job.

- This makes release automation much easier, which I intend to do next.

Finally, if for some reason we want to run a specific workflow individually (instead of running of all them), I've kept the [`workflow_dispatch` option](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow) in each workflow, so we can always dispatch them manually.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,31 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  format-lint:
+    name: Code Linting
+    uses: ./.github/workflows/format-lint.yml
+
+  linux:
+    name: Linux
+    needs:
+      - format-lint
+    uses: ./.github/workflows/linux.yml
+
+  macos:
+    name: macOS
+    needs:
+      - format-lint
+    uses: ./.github/workflows/macos.yml
+
+  windows:
+    name: Windows
+    needs:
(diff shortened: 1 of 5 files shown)


guihkx:
This setup looks kind of bad if you view it from the ['Checks' tab](https://github.com/nuttyartist/notes/pull/583/checks) in this pull request, but if you go to the [Actions page](https://github.com/nuttyartist/notes/actions/runs/5229236975) instead, it looks much better, IMO.

Anyway, let me know your thoughts.

zjeffer (approved):
I like it!

nuttyartist:
I like it too!

guihkx:
Awesome! :smile:

guihkx:
I just pushed a commit (https://github.com/nuttyartist/notes/commit/5b76489d2f8bb5d4feb782e7049645e7676c8572) directly into `master` to address the failing macOS build, which was caused by the lack of access to GitHub Secrets.

PR #584, opened 2023-06-10
chore: fix changeFont using repaint instead of update

zjeffer:
Looks like we only had one of these @bjorn .

I see no difference in functionality.


nuttyartist:
@guihkx Can you test if you still get 100% CPU usage with Adwaita Qt?

guihkx:
> @guihkx Can you test if you still get 100% CPU usage with Adwaita Qt?

Unfortunately it still does.

bjorn (approved):
I think it's still a good change. :-)

PR #586, opened 2023-06-12
Fix launching with Adwaita theme by removing unpolish() call when changing theme

zjeffer:
Solution found by @guihkx. 

Haven't noticed any issues when removing this line. I don't yet know why many examples online always call unpolish first.


guihkx (approved):
Yeah, I'm not sure about the correctness of this, but I tested the deb package on Ubuntu 18.04 (which still uses Qt 5.9), and theme switching still looks fine to me. I have not tested on any other OS, however.

And because this fixes [this Adwaita-Qt](https://github.com/FedoraQt/adwaita-qt/issues/193) issue, I say LGTM.

Issue #537, opened 2023-03-16
Bug allowing the creation of a note when the Trash folder is empty.


nuttyartist:
![test](https://user-images.githubusercontent.com/16375940/225730794-ad6c4987-9718-4d72-beaa-eb0223891ab8.gif)

PR #588, opened 2023-06-21
Fix bug allowing the creation of a note when the Trash folder is empty (#537)

rubenromani:
Issue https://github.com/nuttyartist/notes/issues/537:  Bug allowing the creation of a note when the Trash folder is empty

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -1940,6 +1940,10 @@ void MainWindow::onClearButtonClicked()
 void MainWindow::createNewNote()
 {
     m_listView->scrollToTop();
+    auto inf = m_listViewLogic->listViewInfo();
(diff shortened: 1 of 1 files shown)


nuttyartist:
Awesome, thanks for that! I'll wait for one more review and we can merge this.

zjeffer (changes requested):
This fixes the bug, but causes the console to print this log with every letter you type in that (invalid) note:

`NoteEditorLogic::onTextEditTextChanged() : m_currentNote is not valid`

It would be better if the text editor was not editable at all, if the trash folder is selected and empty. Maybe find a way to "disable" the editor?

nuttyartist:
Closing this due to #642. Thanks, @rubenromani!

Issue #347, opened 2022-09-17
UpdaterWindow doesn't have proper Window decorations on Linux


DesiOtaku:
On KDE, the regular window shows a proper border + shadow so it works fine on a white background. However, UpdaterWindow ignores the KDE Qt style and blends in to the white background.

Also the "Close" and "Update" button isn't consistent with the rest of the Qt style (it looks more like Android rather than KDE).

DesiOtaku:
![image](https://user-images.githubusercontent.com/410561/190856341-a4756193-f112-4770-af17-ac7a325690c7.png)

nuttyartist:
Thanks for reporting, we'll look into that.

guihkx:
Fixed by #469.

bjorn:
> Fixed by #469.

@guihkx Really? Did you test it? We're still setting `Qt::CustomizeWindowHint` on the updater window so it will not have a native decoration. It might be nice to follow the same configuration as for the main window.

> Also the "Close" and "Update" button isn't consistent with the rest of the Qt style (it looks more like Android rather than KDE).

@DesiOtaku Note that their style has been intentionally customized. There is currently no option for using the native widget style.

guihkx:
@bjorn It "works", but I just noticed that's maybe because frameless mode seems to be ignored altogether on Wayland now... :/

KDE Plasma

![image](https://user-images.githubusercontent.com/626206/218418138-b8f28293-f817-4e3f-b697-b4dab169f715.png)

GNOME

![Screenshot from 2023-02-13 09-15-28](https://user-images.githubusercontent.com/626206/218418214-0a2097f9-bf9c-4a8a-a160-7464e4380909.png)

bjorn:
> It "works", but I just noticed that's maybe because frameless mode seems to be ignored altogether on Wayland now... :/

On the KDE screenshot I don't see the update window.

Regarding frameless mode not working on Wayland, could that mean we need to put back the `Qt::FramelessWindowHint` on Linux?

guihkx:
> On the KDE screenshot I don't see the update window.

Dang it, uploaded the wrong one. But it matches GNOME's.

> Regarding frameless mode not working on Wayland, could that mean we need to put back the `Qt::FramelessWindowHint` on Linux?

Precisely. After I re-added that to `mainwindow.cpp` and `updaterwindow.cpp`, native decorations in frameless mode are now gone, at least:

[...]

PR #598, opened 2023-07-22
Linux: Fix Wayland window associations

ReillyBrogan:
Wayland compositors use the desktopFileName window property to associate native Wayland windows with their .desktop launcher.

This was previously set to "notes" when qt launched as a native Wayland window (the default in Wayland sessions), which could cause various inconsistencies.

Notably, KDE Plasma would display the generic Wayland icon in task view for Notes instead of the icon referenced in the .desktop file.

Fix this by propagating the APP_ID property (which is used to name the .desktop file during packaging) to main.cpp so that we can call setDesktopFileName with it.

After this change KDE Plasma correctly displays the Notes icon in task view.

Before:
image

After:
image

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -16,6 +16,10 @@ int main(int argc, char *argv[])
     app.setApplicationName("Notes");
     app.setApplicationVersion(APP_VERSION);
 
+#if defined(Q_OS_UNIX) && !defined(Q_OS_MACOS)
+    app.setDesktopFileName(QString(APP_ID) + ".desktop");
+#endif
+
 #if QT_VERSION >= QT_VERSION_CHECK(5, 10, 0) && QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
     app.setAttribute(Qt::AA_DisableWindowContextHelpButton);
 #endif
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -267,7 +267,7 @@ endif()
 
 target_compile_definitions(
   ${PROJECT_NAME}
-  PUBLIC APP_VERSION="${PROJECT_VERSION}${GIT_REV}"
+  PUBLIC APP_VERSION="${PROJECT_VERSION}${GIT_REV}" APP_ID="${APP_ID}"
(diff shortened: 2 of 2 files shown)


zjeffer (approved):
Awesome! Thanks for the fix.

ReillyBrogan:
Note that I have no experience with cmake or QT, so this solution may not be particularly idiomatic or clean. This should work for Flatpak but I have no idea how to test that.

I tested the following configurations after making this change and all worked as expected:
- x11 session
- Wayland session with native Wayland window
- Wayland session but forcing Notes to launch with xcb rendering

ReillyBrogan:
Lint issues should be fixed now

src/main.cpp
```diff
@@ -16,6 +16,10 @@ int main(int argc, char *argv[])
     app.setApplicationName("Notes");
     app.setApplicationVersion(APP_VERSION);
 
+#if defined(Q_OS_LINUX) && QT_VERSION >= QT_VERSION_CHECK(5, 7, 0)
```
guihkx, line 19: I think this check can be expanded to other *nixes too. And the Qt version check can be removed as well, as we already require at least Qt 5.12.0:

```suggestion
#if defined(Q_OS_UNIX) && !defined(Q_OS_MACOS)
```

guihkx (commented):
Thank you for the contribution!

Please allow me some time to test this on a Wayland session, though I'm fairly confident this won't break anything.

ReillyBrogan:
Note that this also fixes an issue with the native window frame. Before on Plasma it would display the generic Wayland icon in the top left, now it displays the Notes icon. This wasn't an issue on GNOME because GNOME Wayland reads the StartupWMClass property from the .desktop file and uses that to match windows as well (which is technically non-conformant with the Freedesktop/Wayland specs but GNOME devs do whatever they want to do). 

You can see this in the screenshots in this comment: https://github.com/nuttyartist/notes/issues/347#issuecomment-1427598576

nuttyartist:
Awesome! thanks, @ReillyBrogan! Let's wait for @guihkx to test on his end as well before merging (:

guihkx:
Thanks!

PR #602, opened 2023-07-25
Linux: Set app id without `.desktop` suffix

guihkx:
Silences this startup warning given by Qt 6.5.2:

`QGuiApplication::setDesktopFileName: the specified desktop file name
ends with .desktop. For compatibility reasons, the .desktop suffix will
be removed. Please specify a desktop file name without .desktop suffix`

Here's the rationale, according to a Qt dev:

>The desktop file name should not contain ".desktop" suffix, but some applications still specify it anyway because of the ambiguity in the documentation that was fixed in [0c5135a9dfa6140d23d86b001c3054891c22dcb9](https://codereview.qt-project.org/c/qt/qtbase/+/477797).

Source: https://codereview.qt-project.org/c/qt/qtbase/+/480688

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -17,7 +17,7 @@ int main(int argc, char *argv[])
     app.setApplicationVersion(APP_VERSION);
 
 #if defined(Q_OS_UNIX) && !defined(Q_OS_MACOS)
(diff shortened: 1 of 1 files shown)


nuttyartist:
👍

PR #616, opened 2023-07-28
Add ubuntu23.04 build to GitHub actions

zjeffer:


zjeffer:
Damn, there is no runner yet for ubuntu-23.04 :(

guihkx:
> Damn, there is no runner yet for ubuntu-23.04 :(

There's no official GitHub runner, no. But they allow us to use any other container o,age.

Anyway, I think this is a good time to replace our deb package made for Ubuntu 22.10 by one made for Ubuntu 23.04, since the former [will reach EOL](https://wiki.ubuntu.com/Releases) at the end of this month.

If you want to give it another shot, I think you'll have to only change this line (hopefully):

https://github.com/nuttyartist/notes/blob/67e724fb70101a59a20b862ba440ea548deadab7/.github/workflows/linux.yml#L28

PR #626, opened 2023-08-01
CI: Replace Ubuntu 22.10 by 23.04

guihkx:
Ubuntu 22.10 (Kinetic Kudu) [has reached end-of-life](https://wiki.ubuntu.com/Releases) in July 2023, so let's replace it by Ubuntu 23.04 (Lunar Lobster), which will be supported by Canonical until January 2024.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -16,6 +16,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
+          # Ubuntu's release cycle: https://wiki.ubuntu.com/Releases
           - os: ubuntu-20.04
             qt-version-major: 5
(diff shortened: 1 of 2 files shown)


.github/workflows/linux.yml
```diff
@@ -25,7 +26,7 @@ jobs:
             build-type: release
 
           - os: ubuntu-20.04
-            container: ubuntu:22.10
+            container: ubuntu:23.04
```
zjeffer, line 29: I don't fully understand this. What's the difference between os and container in this case? Why are we still setting os to ubuntu-20.04 while the container is 23.04?

nuttyartist (approved):

PR #627, opened 2023-08-01
CI: Replace Fedora 36 by 37

guihkx:
Fedora 36 [has reached end-of-life](https://docs.fedoraproject.org/en-US/releases/eol/#_unsupported_fedora_linux_releases) in May 2023, so let's replace it by version 37.


guihkx:
Pushed this last commit to the wrong branch, my bad. I'll fix it.

nuttyartist (approved):

PR #636, opened 2023-08-15
Fixed loading of custom font icons

### LINKED ISSUE #620: Icons not showing properly
Screenshot_20230729_215044
This happens if compiled from source, and also in the Appimages (both Qt5 and Qt6).
Running on KDE Plasma 5.27.6, Xorg, Arch Linux.

zjeffer:
Fixes #620 

Apparently the QFont constructor doesn't load fonts from the application's database. We have to use `QFontDatabase::font`, which does load the fonts properly when `otf-font-awesome` is installed.

I replaced every occurence of QFont when loading our custom fonts, with the QFontDatabase::font constructor.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -148,6 +148,8 @@ set(SOURCES
     ${PROJECT_SOURCE_DIR}/src/elidedlabel.h
     ${PROJECT_SOURCE_DIR}/src/foldertreedelegateeditor.cpp
     ${PROJECT_SOURCE_DIR}/src/foldertreedelegateeditor.h
+    ${PROJECT_SOURCE_DIR}/src/fontloader.cpp
+    ${PROJECT_SOURCE_DIR}/src/fontloader.h
     ${PROJECT_SOURCE_DIR}/src/labeledittype.cpp
     ${PROJECT_SOURCE_DIR}/src/labeledittype.h
     ${PROJECT_SOURCE_DIR}/src/listviewlogic.cpp
--- a/src/allnotebuttontreedelegateeditor.cpp
+++ b/src/allnotebuttontreedelegateeditor.cpp
@@ -6,6 +6,7 @@
 #include "nodetreeview.h"
 #include "notelistview.h"
 #include "editorsettingsoptions.h"
+#include "fontloader.h"
 
 AllNoteButtonTreeDelegateEditor::AllNoteButtonTreeDelegateEditor(QTreeView *view,
                                                                  const QStyleOptionViewItem &option,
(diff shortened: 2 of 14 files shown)


zjeffer:
Looks like the QML fonts are not yet fixed (top right):

image

image

I don't know much about QML, would you know if there's a similar way of loading the application database's fonts instead of the system's @nuttyartist ?

nuttyartist:
Thanks for this!

> I don't know much about QML, would you know if there's a similar way of loading the application database's fonts instead of the system's @nuttyartist ?

I'm quite sure the qml loads the app fonts and not the system fonts since the path to the fonts is `qrc:/fonts/...` which are embedded to the binary. So I'm not sure what's happening.

nuttyartist:
I guess it won't hurt to merge this even tho we still didn't figure why the QML fonts won't load?

guihkx:
> I guess it won't hurt to merge this even tho we still didn't figure why the QML fonts won't load?

As long as all build jobs pass, I'm fine with that.

nuttyartist:
> Seems like it doesn't work for qt5 at the moment but I don't really have the time to figure out why at the moment.

I'll look into it.

nuttyartist:
The reason (per doc and error) is that in Qt 6 QFontDatabase::font is a static function and in Qt 5 it's not (so we need to istantiate an object for it).

Screen Shot 2023-08-23 at 7 22 59 PM

Screen Shot 2023-08-23 at 7 24 03 PM

nuttyartist:
I actually don't know why calling QFontDatabase::font should be needed, the docs don't seem to say that this is any different than loading via QFont's constructor. But I didn't notice we need to load two more fonts in main.cpp

```
if (QFontDatabase::addApplicationFont(":/fonts/fontawesome/fa-regular-400.ttf") < 0)
        qWarning() << "FontAwesome Regular cannot be loaded !";

    if (QFontDatabase::addApplicationFont(":/fonts/fontawesome/fa-brands-400.ttf") < 0)
        qWarning() << "FontAwesome Brands cannot be loaded !";        
```

Tho I'm I don't think these are the fonts that don't load for you.

nuttyartist:
I do see now that there are both `QFontDatabase::font` and `QFontDatabase::systemFont` so you're probably right.

zjeffer:
> Should we just create an object for both Qt 6 and 5 or do everything with #if?

I think maybe creating a separate class that loads fonts might be the best way. In that class, we can use a single #if. I'll look into trying that.

nuttyartist:
> > Should we just create an object for both Qt 6 and 5 or do everything with #if?
> 
> I think maybe creating a separate class that loads fonts might be the best way. In that class, we can use a single #if. I'll look into trying that.

Ahaha sounds promising.

zjeffer:
Alright lets see if this works

nuttyartist:
Did it solve the problem?

zjeffer:
All checks seem to have passed and it seems to work on my system (linux, qt6), but we need to test it on other platforms as well. Can someone test on Windows & macOS?

nuttyartist:
I tested on macOS and seems to be fine. I can test Windows tonight/tomorrow morning.

nuttyartist:
Tested on Windows, working well. Thanks, @zjeffer! Merging...

PR #639, opened 2023-08-24
Option to disable the confetti in Kanban

### LINKED ISSUE #619: Option to disable the confetti upon completing all tasks in a column in kanban mode
It's very distracting and i can imagine that there can be some accessibility issues regarding it

nuttyartist:
Fixes #619

### DIFF
--- a/src/qml/TodoColumnDelegate.qml
+++ b/src/qml/TodoColumnDelegate.qml
@@ -44,7 +44,7 @@ MouseArea {
     pressAndHoldInterval: 200
 
     function checkTasksForConfetti () {
-        if (numberOfCompletedTasks === tasksView.model.items.count) {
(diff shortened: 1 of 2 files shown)


guihkx (approved):
This works, but I wonder why Kanban-specific options aren't permanent?

nuttyartist:
Because we'll need to save these options into the database, which I don't know how to do. In the next version, we'll save these type of metadata in the text itself, which will allow the portability of files (using plain text rather than a databse).
That'll look something like this:

```
{{kanban showProgressBars=true showConfetti=false}}
#### TODO
- [ ] Task 1
- [ ] Task 2
#### In Progress
- [ ] Task 1
- [ ] Task 2
{{/kanban}}
```

guihkx:
Oh yeah, you spoke about that before. Nice.

PR #642, opened 2023-08-25
Disable editing inside textEdit when trash folder is selected

nuttyartist:
Following @rubenromani PR https://github.com/nuttyartist/notes/pull/588, I decided it's better to just disable editing when the trash folder is selected. It only makes sense that a user should restore his data before editing it. This fixes issue #537.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -630,12 +630,9 @@ void MainWindow::setupKeyboardShortcuts()
     connect(new QShortcut(QKeySequence(Qt::CTRL | Qt::SHIFT | Qt::Key_K), this),
             &QShortcut::activated, this, [=]() {
                 if (m_kanbanWidget->isHidden()) {
-                    emit m_noteEditorLogic->showKanbanView();
-                    updateSelectedOptionsEditorSettings();
+                    setKanbanVisibility(true);
                 } else {
-                    emit m_noteEditorLogic->hideKanbanView();
-                    m_textEdit->setFocus();
-                    updateSelectedOptionsEditorSettings();
+                    setKanbanVisibility(false);
                 }
             });
 #endif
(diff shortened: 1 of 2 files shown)


guihkx:
This solves one way to create a note inside the Trash folder, yes, but you can still create/modify one through the Kanban interface.

And perhaps another improvement would be disabling the "New Note" option from context menu (it shows up when right-clicking a note that's in the Trash folder). Though right now this option does nothing.

Side note: The first commit you added here is already present in #641.

nuttyartist:
> Side note: The first commit you added here is already present in https://github.com/nuttyartist/notes/pull/641.

Oops, what should I do about it?

> This solves one way to create a note inside the Trash folder, yes, but you can still create/modify one through the Kanban interface.

Ahah! Good catch! I had a feeling I'm missing something.

> And perhaps another improvement would be disabling the "New Note" option from context menu (it shows up when right-clicking a note that's in the Trash folder). Though right now this option does nothing.

Another good catch. I'll see what I can do.

guihkx:
Since that PR is now merged into `master`, you can simply sync your local `master` branch (`git checkout master && git pull`), and then rebase this PR's branch against `master` (`git checkout trash-behavior && git rebase -i master`) and then force-push.

Sent from Proton Mail mobile

-------- Original Message --------

nuttyartist:
Thanks for the exaplanation!

zjeffer (approved):

nuttyartist:
> This solves one way to create a note inside the Trash folder, yes, but you can still create/modify one through the Kanban interface.

Latest commit should address this, but If the Kanban is visible and an empty folder is selected, the kanban will still be visible. I couldn't figure this out.

> And perhaps another improvement would be disabling the "New Note" option from context menu (it shows up when right-clicking a note that's in the Trash folder). Though right now this option does nothing.

I couldn't fix this problem from the same reason as above. I just need to get a signal in mainWindow that the user is inside the Trash folder, but I couldn't make it to work (#in_qml_it's_easier).

guihkx:
>I couldn't fix this problem from the same reason as above.

I think hiding the "New Note" context menu option while we're in the Trash folder is actually easy to do. Can you check if my last commit fixes it?

Also, I think your fix that prevents the Kanban button from working while in the Trash folder is good enough for now, but in a future PR we can figure out how to display the Kanban view in read-only mode instead.

nuttyartist:
Indeed! It works, thanks!

> Also, I think your fix that prevents the Kanban button from working while in the Trash folder is good enough for now, but in a future PR we can figure out how to display the Kanban view in read-only mode instead.

Yes, I agree.

guihkx (approved):
Thanks for testing! :)

PR #649, opened 2023-09-05
Linux: Remove low resolution icons

guihkx:
This is just an aesthetical change.

Long story short:

AppImages have a default icon. For some reason, our AppImage-making tool always chooses the lowest resolution icon we have (16x16) as the default one, and there's no way to specify a different icon (AFAIK).

End result: Our AppImage icon looks really bad on file managers that support previewing AppImage files via libappimage, like Dolphin.

Quick and dirty solution: Keep only icons that are 128x128 and above.

Comparison:

image


nuttyartist:
Won't we need the lower ones for the system tray?

guihkx:
No. We actually use the [1024x1024 icon](https://github.com/nuttyartist/notes/blob/70ee81ff09c082c4179713d6a07bda1151c30188/src/images/notes_system_tray_icon.png) for the tray icon here:

https://github.com/nuttyartist/notes/blob/70ee81ff09c082c4179713d6a07bda1151c30188/src/mainwindow.cpp#L522-L535

And I think Qt takes care of resizing the icon for us.

zjeffer (approved):

nuttyartist:
Okay 👍

PR #653, opened 2023-11-07
CI/Windows: Use OpenSSL 1.x binaries from FireDaemon

guihkx:
This fixes the Windows builds.

Now that [Qt 6.5+ relies on OpenSSL 3.x](https://www.qt.io/blog/moving-to-openssl-3-in-binary-builds-starting-from-qt-6.5-beta-2), Qt (the company) has deleted from their servers the OpenSSL 1.x binaries they used to provide for Windows.

However, since we still use Qt 5.15 / 6.4, we still need those OpenSSL 1.x binaries, so we have to download them from elsewhere.

I decided to go with [OpenSSL binaries built by FireDaemon](https://kb.firedaemon.com/support/solutions/articles/4000121705), as they're free, have no external dependencies (i.e. MSVC), work on Windows 7, and have the additional benefit of being digitally signed.


zjeffer (approved):

nuttyartist:
Awesome. I'm going to test this tomorrow. What Windows version did you test this on?

guihkx:
>I'm going to test this tomorrow

No problem.

>  What Windows version did you test this on?

Only on Windows 7, via QEMU.

They have [this compatibility table](https://kb.firedaemon.com/support/solutions/articles/4000121705) which mentions Windows 7 being "compatible, but not supported". Only Windows 10 and 11 are actually "supported".

I also just force-pushed to adjust the commit message.

nuttyartist:
Seems to work well on Windows 10 as well, thanks @guihkx!

nuttyartist:
I see there are 2 failed builds. Why is that?

zjeffer:
They seemed to have timed out (they were running for more than 3 hours). I managed to restart the macOS homebrew qt5 build, let's see how that goes.

guihkx:
Homebrew seems to have stopped distributing Qt binaries for the macOS version we're currently using (Big Sur), which means they will be built from source, which means it will take ages and GitHub Actions will eventually time out.

See #654

nuttyartist:
Oh damn, it was still going. Alrighty, merging...

PR #655, opened 2023-11-11
CI/macOS: Update to macOS 12 runner for Homebrew build jobs

### LINKED ISSUE #654: CI: Upgrade version of macOS runner
We currently use macOS 11, but Homebrew apparently dropped support for it, which means that installing Qt will take ages, since it has to compile it from source.

Investigate moving to GitHub's `macos-12` runner and its implications.

I'm also not that familiar with macOS/homebrew, so I'm accepting other suggestions here.

guihkx:
macOS 11 isn't [officially supported by Homebrew](https://docs.brew.sh/Installation#macos-requirements) anymore, which means they have stopped distributing Qt binaries for that version.

This made some of our macOS jobs to timeout in the "Install Qt" step, because Qt was being built from source...

Please note that this change will not have any impact on macOS users, since we distribute official binaries from Qt, not from Homebrew.

Closes #654

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -15,11 +15,11 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - os: macos-11
+          - os: macos-12
             qt-version-major: 5
             build-type: debug
 
-          - os: macos-11
+          - os: macos-12
             qt-version-major: 6
             build-type: debug
     steps:


zjeffer (approved):

nuttyartist:
Awesome job again, @guihkx. Thanks.

PR #657, opened 2023-12-01
Fix #652: option window doesnt work

### LINKED ISSUE #652: Editor settings window is broken with Qt 6.6
Qt 6.6 was released just a few days ago, and while Notes builds just fine with it, the Editor Settings window won't open anymore:

https://github.com/nuttyartist/notes/assets/626206/fabae2bd-3973-4039-a3f5-6e9c723b891e

This is the output I get when I open the app using the terminal:

```
$ notes
qrc:/qt/qml/EditorSettings.qml:429:21: Type OptionItemButton unavailable 
                         OptionItemButton { 
                         ^
qrc:/qt/qml/OptionItemButton.qml:15:5: Cannot override FINAL property 
         property bool enabled: true 
         ^
qrc:/qt/qml/kanbanMain.qml:615:33: Type OptionItemButton unavailable 
                                     OptionItemButton { 
                                     ^
qrc:/qt/qml/OptionItemButton.qml:15:5: Cannot override FINAL property 
         property bool enabled: true 
         ^
Database: connection ok
```

This was tested on commit 10bed3427934456711a7e49b05ecb47fff617666 and on Linux. I haven't tested on other platforms.

zjeffer:
The issue was that `enabled` is a built-in QML property, and we were overwriting it. It seems that since 6.6, this overwriting actually causes an error, as opposed to previous versions. 

Fixed by not redefining the `enabled` property, but simply setting to `true` in the constructor.

Thanks to @dmytrovoytko for pointing us closer to the solution!

Tested and works on Qt 6.6.1

### DIFF
--- a/src/qml/OptionItemButton.qml
+++ b/src/qml/OptionItemButton.qml
@@ -12,9 +12,9 @@ MouseArea {
     property bool isContainingMouse: optionItemMouseArea.containsMouse
     property bool checked: false
     property var themeData: {{thme: "Light"}}
-    property bool enabled: true
     property int pointSizeOffset: -4
 
+    enabled: true
     hoverEnabled: true
     width: innerRectangle.width
     height: innerRectangle.height


zjeffer:
Appimage 5.15.2 works

zjeffer:
Appimage 6.4.3 also works

guihkx (approved):
Works for me too! But we should probably wait for @nuttyartist's review.

nuttyartist:
Works just fine (: Thanks, @zjeffer.

PR #659, opened 2023-12-08
CI/Linux: Update to Fedora 38

guihkx:
Fedora 37 reached end-of-life in December 5th, 2023.

https://lists.fedoraproject.org/archives/list/devel-announce@lists.fedoraproject.org/thread/BOMYDF5U3FVAQSOGM6NZEZCTRBMQQJ56/

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -122,7 +122,7 @@ jobs:
         include:
           - os: ubuntu-20.04
             # Fedora's release cycle: https://docs.fedoraproject.org/en-US/releases/lifecycle/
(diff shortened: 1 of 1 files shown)


zjeffer (approved):

zjeffer:
Maybe we should immediately bump it to 39? It's the latest release: https://en.wikipedia.org/wiki/Fedora_Linux#Releases

guihkx:
We could, but I personally think we should try supporting the oldest (yet still supported) release of a distro. 

Since version 38 is still supported by Fedora, people are still using it.

PR #660, opened 2023-12-09
Added .clang-tidy file

zjeffer:
I also added this for the Waybar repo (see discussion [here](https://github.com/Alexays/Waybar/discussions/2591) for more info on what it is).

Later we could also add a GitHub action that checks whether new code adheres to these clang-tidy checks (I'm testing this here: https://github.com/zjeffer/Waybar/pull/1/files)


.clang-tidy
```diff
@@ -0,0 +1,28 @@
...
+  -fuchsia-trailing-return,
+  -readability-magic-numbers,
+  -modernize-use-nodiscard,
+  -modernize-use-trailing-return-type,
+  -readability-braces-around-statements,
+  -readability-redundant-access-specifiers,
+  -readability-redundant-member-init,
+  -readability-redundant-string-init,
+  -readability-identifier-length,
+CheckOptions:
```
zjeffer, line 18: I also added naming checks, so we all use the same naming conventions. I chose these based on what we currently use most often.
Let me know if any of these need to change.

guihkx:
This looks very promising!

Can I test this locally? If so, what commands should I use exactly?

zjeffer:
In vscode, with the `clangd` extension, you can see this in the editor itself:

https://github.com/nuttyartist/notes/assets/4633209/c958a964-713d-4406-a28e-537df957141a

You can also manually run clang-tidy in a terminal:

`find src -name '*.cpp' -o -name '*.hpp' | xargs clang-tidy`. 

You need to install clang-tidy on your system (using your package installer of choice)

guihkx:
The VSCode integration looks neat!

> `find src -name '*.cpp' -o -name '*.hpp' | xargs clang-tidy --config-file=./.clang-tidy`.

Thanks, but I think that's probably not the "right way" to do it?

Because I'm getting a *lot* of errors running it that way, including nonsensical ones like:

```
/home/gui/dev/notes/src/aboutwindow.h:4:10: error: 'QDialog' file not found [clang-diagnostic-error]
#include <QDialog>
         ^~~~~~~~~
```

zjeffer:
You can pass the compile_commands.json file:

`find ./src -regex ".*\.\(c\|cpp\|h\|hpp\)$" | xargs clang-tidy --config-file ./.clang-tidy -p ./build/compile_commands.json`

guihkx:
> You can pass the compile_commands.json file:

Alright, so I didn't know how to generate that file, but simply invoking CMake with `-DCMAKE_EXPORT_COMPILE_COMMANDS=ON` did the trick.

Perhaps we can add it to `CMakeLists.txt` so it can be generated by default?

>It's a lot slower when doing this, though, because it has to parse all included Qt header files as well. But it looks a lot better.

It too a while to finish, indeed, but the amount of warnings and errors I got at the end still seem too high...?

```
$ find ./src -regex '.*\.\(c\|cpp\|h\|hpp\)$' | xargs clang-tidy --config-file ./.clang-tidy -p ./build/compile_commands.json
[... very long output ...]
1139005 warnings and 41 errors generated.
Error while processing /home/gui/dev/notes/src/nodedata.h.
error: too many errors emitted, stopping now [clang-diagnostic-error]
```

Or is that expected?

zjeffer:
> Perhaps we can add it to CMakeLists.txt so it can be generated by default?

This should indeed be added to CMakeLists.txt by default. I think the vscode CMake Tools extension already does that for me when I configure, which is why I didn't notice it wasn't included.

zjeffer:
> It too a while to finish, indeed, but the amount of warnings and errors I got at the end still seem too high...?
> 
> $ find ./src -regex '.*\.\(c\|cpp\|h\|hpp\)$' | xargs clang-tidy --config-file ./.clang-tidy -p ./build/compile_commands.json
> [... very long output ...]
> 1139005 warnings and 41 errors generated.
> Error while processing /home/gui/dev/notes/src/nodedata.h.
> error: too many errors emitted, stopping now [clang-diagnostic-error]
> 
> Or is that expected?

I also get roughly that amount of warnings, but it doesn't stop with `too many errors emitted`, that's odd. I'll run it again.

zjeffer:
Nevermind, I do get the `too many errors emitted, stopping now [clang-diagnostic-error]` error, here's my full output:

[clang-tidy-warnings.txt](https://github.com/nuttyartist/notes/files/13637049/clang-tidy-warnings.txt)

zjeffer:
@guihkx Found two ways to make it much faster:

* the `clang` package includes both `/usr/bin/clang-tidy` and `/usr/bin/run-clang-tidy`. The latter runs clang-tidy in parallel.
* instead of finding first and then running clang-tidy on every file using `xargs`, you can run find in $()

It still outputs the warning counts like:

```
1139005 warnings and 41 errors generated.
Error while processing /home/gui/dev/notes/src/nodedata.h.
error: too many errors emitted, stopping now [clang-diagnostic-error]
```
but you can disable that with `-quiet`.


---

The command then becomes:

```sh
run-clang-tidy -p ./build/ -config-file ./.clang-tidy $(find ./src -regex '.*\.\(cpp\|hpp\|h\|c\)$') -quiet
```

New output: [clang-tidy-warnings2.txt](https://github.com/nuttyartist/notes/files/13660181/clang-tidy-warnings2.txt)

guihkx (approved):
>Either way, clang-tidy is not really supposed to be used like this, you're supposed to set up your IDE with clang as a language server and then enable clang-tidy in the clang options.

Yeah, using it like makes more sense. But it's still nice to know there's a way to run with only a command.

PR #662, opened 2023-12-21
CI: Only download required build artifacts to make the Windows installer

guihkx:
Our Windows CI has a separate build job with the sole purpose of creating an installer, by joining three different Windows builds:

- Qt5: x86 and x86_64
- Qt6: x86_64

To accomplish that, this job needs to download these build artifacts from previous jobs.

Until now, we would download all build artifacts, including ones from unrelated jobs such as Linux and macOS, which doesn't make sense, as it's wasteful and slow.

Thankfully, version [4.1.0](https://github.com/actions/download-artifact/releases/tag/v4.1.0) of `actions/download-artifact` now allows us to specify which artifacts to download, which is what this pull request does.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -96,7 +96,7 @@ jobs:
           lintian '${{ steps.deb.outputs.path }}'
 
       - name: Upload deb package
-        uses: actions/upload-artifact@v3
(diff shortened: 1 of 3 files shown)


zjeffer (approved):
Cool!

nuttyartist:
👍

PR #666, opened 2024-02-18
Use Docker files for Linux workflows

zjeffer:
This PR converts the rest of the linux workflows to use docker images, speeding up the CI/CD.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -103,34 +103,25 @@ jobs:
           path: ${{ steps.deb.outputs.path }}
 
   rpm:
-    name: rpm (${{ matrix.build-type }}, Qt ${{ matrix.qt-version-major }}, ${{ matrix.container != '' && matrix.container || matrix.os }})
-    runs-on: ${{ matrix.os }}
-    container: ${{ matrix.container }}
+    name: rpm (${{ matrix.build-type }}, Qt ${{ matrix.qt-version-major }}, ${{ matrix.image }})
+    runs-on: ubuntu-latest
+    container: zjeffer/notes:${{ matrix.image }}
     strategy:
       fail-fast: false
       matrix:
         include:
-          - os: ubuntu-20.04
-            # Fedora's release cycle: https://docs.fedoraproject.org/en-US/releases/lifecycle/
-            container: fedora:38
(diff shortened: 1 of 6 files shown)


nuttyartist:
Awesomee, thanks, @zjeffer!

nuttyartist (approved):

.github/workflows/linux.yml
```diff
@@ -307,6 +262,11 @@ jobs:
         env:
           VERBOSE: 1
         run: |
+          echo "LD_LIBRARY_PATH=${LD_LIBRARY_PATH}"
```
guihkx, line 265: Is this a leftover from debugging? The lines below this one as well.

guihkx (commented):
This is really nice work, thank you!

I have a few questions, that I should've asked in the previous PR, actually.

I'm not 100% sure how Docker works, but from I understand, ideally, these images have to be kept constantly up to date, right?

For example, in a scenario where a package in Ubuntu 20.04 (that we ship to users in the AppImage) has a serious CVE, do you have to rebuild and republish these images to the Docker website? Or, when we pull these images, the base image is automatically updated?

Based on a local test, I think it's the first option: I pulled `zjeffer/notes:ubuntu-22_04` (which was last updated two months ago), and there were 43 packages out of date.

If that's the case, can we somehow automate the updating and publishing of these images?

I understand GitHub itself also has it's own OCI Registry, correct? So, it'd be really awesome if we could integrate that here, instead of having to dump all this maintenance work on you.

.github/workflows/linux.yml
zjeffer: Yes, nicely spotted. Will remove.

zjeffer:
> If that's the case, can we somehow automate the updating and publishing of these images?

Yes, that should be possible. We should be able to automatically build these every week or so, for example.
Will look into this maybe next weekend when I have more time.


> I understand GitHub itself also has it's own OCI Registry, correct? So, it'd be really awesome if we could integrate that here, instead of having to dump all this maintenance work on you.

In my experence, the GitHub container registry is extremely slow to upload/download from. That's why I always use the official Docker registry. It should still be possible to automate the builds with this, so there won't be any maintanance work from my side.

guihkx (approved):
Gotcha, thanks for clarifying!

PR #670, opened 2024-05-08
add notes with scoop x64

ghost:


guihkx:
Hey, thanks for working on this! :)

I haven't tested this properly yet, as I don't usually review draft pull requests, because they're technically not ready...

But anyway, you should probably update the `notes.json` file to use the Qt 6 build instead of the Qt 5 one, since the latter doesn't have the Kanban feature.

guihkx:
I'm not familiar with how Scoop works, but after we install Notes using it, how do we open it?

I assume it could only be launched through the command line and it won't be available in Windows' start menu. Or am I mistaken?

guihkx:
Thanks for clarifying things!

In my opinion, since Notes is obviously a graphical-only program, having a shortcut in the start menu to open it is a must. If this is indeed doable with Scoop, I'd prefer having that before we could merge it.

>I was wondering where I should store notes.json, or if I should create notesx86.json

I'm not sure what other contributors think, but in my opinion, having a single `notes.json` file only for the x64 Qt 6 build is preferable.

We pretty much only offer the x86 Qt 5 build because, apparently, some of our users are still on Windows 7, where Qt 6 isn't officially supported.

guihkx:
>I was asking if I should submit the pull request to the scoop community or here

I definitely think it would make more sense to submit to the Scoop community instead.

Based on a quick look I took at how Scoop works, they seem to have their own official repositories, so installing Notes from one of those would be much better, IMO.

However, if they have a way to "import" the Scoop file from third-party repositories (i.e., from here into their own repositories), I think it would be even better.

Just so we can "control" the release process better, directly from here.

Issue #316, opened 2022-09-11
Import/Export .md and .txt files to/from the database


KhaaL:
As a enduser, in order to migrate to Notes I'd need to migrate my already made text-files. The only alternative is to copy-paste them, which would be very time consuming.

nuttyartist:
I agree that's a good suggestion. Thanks!

KhaaL:
Let me know if you'd like to have a more fleshed out user-story, I wrote the issue quickly just to get it off my head 😃

nuttyartist:
I think it's a pretty understandable request. If you think there's something you should point out, please do.

nuttyartist:
@batsatt You're right. We will ship this on our next (minor) release. See: https://github.com/nuttyartist/notes/projects/8#card-85711502

badrow:
Very important feature,I need it too ! I have 1700+ notes in txt format, in 10 folders, and I don't know how to import them in this app. 
Thank you for your job, very interesting project !

passionate2023:
Any update on this option?

passionate2023:
Any chance of implementing this?

nuttyartist:
I hope we'll manage to make it happen when releasing v2.3.0, fingers crossed.

passionate2023:
Any update on this?

nuttyartist:
Nothing new yet. No need to keep asking here, please. I'll keep you posted.

passionate2023:
i meant when the 2.3.0 is expected - next year or this year? and not update on the about feature

nuttyartist:
I'm not sure. I estimate it will take around 4 months. It's not going to be 2.3.0 but a major new version.

passionate2023:
Is this feature is available now @nuttyartist ?

nuttyartist:
Hey, not yet. Please ask again in 2 months from now.

nuttyartist:
@passionate2023 I'm happy to finally say I implemented importing .txt/.md files in Plume, and it will see its way into Notes as well. In a few days I should implement exporting to .txt/.md as well.

passionate2023:
Thank you so much @nuttyartist

zjeffer:
Closing as completed.

PR #678, opened 2024-06-22
Version 2.3.0

### LINKED ISSUE #673: Two warnings in `SubscriptionWindow.qml` with Qt 6.7.1
I just built Notes with Qt 6.7.1 on Linux, and upon running Notes from the terminal, I get the following two warnings:

```
$ notes
Database: connection ok
m_subscriptionStatus: SubscriptionStatus::Active
qrc:/qt/qml/SubscriptionWindow.qml:1583: May 22, 2024 should be coerced to void because the function called is insufficiently annotated. The original value is retained. This will change in a future version of Qt.
qrc:/qt/qml/SubscriptionWindow.qml:1564: May 22, 2024 should be coerced to void because the function called is insufficiently annotated. The original value is retained. This will change in a future version of Qt.
```

nuttyartist:
This branch adds the following improvements:

- Fixes #673 
- Remove unnecessary missing Ubuntu font - improves startup time.
- Export/import to/from .txt/.md - solves #316
- Much more efficient notesList load - by improving the `NoteEditorLogic::getSecondLine` function - which also convert Markdown formatted text to plain text, helping create a cleaner text for the notesList pane
- Improve the update mechanism to not just check for update on startup but once every hour as well

There are some more changes I want to make before merging this. This is why it's labeled WIP.


vlasky:
Very happy to see this progress! I am eagerly awaiting encrypted sync functionality between multiple instances on different PCs. For me, that's the killer feature that would make this perfect!

nuttyartist:
Thanks, @vlasky! I don't have a plan to support built-in sync for Notes at this time. If anyone would like to contribute such a thing, I will happily merge it. But I do have plans to support sync for my note-taking app Plume (https://www.get-plume.com/) that shares the same database with Notes.

nuttyartist:
Hey everyone,

I want to announce some changes that will go live in a few days. 

- I made a new website for Notes FOSS: https://www.notes-foss.com/. Open sourced here: https://github.com/nuttyartist/notes-foss-website
- This GitHub repo (https://github.com/nuttyartist/notes) will remain unchanged.
- The Discord will remain dedicated to Notes FOSS.
- The current Ubuntu Snap and FlatHub versions will remain unchanged (Notes FOSS).
- I'll soon rebrand Notes previous website (https://www.get-notes.com/) under the name Daino Notes. It'll basically be Plume (https://www.get-plume.com/) with a different name (the name of my U.S. company (https://www.daino.io/) + Notes).
- The current user base of Notes will get an update to Daino Notes. The new Notes FOSS app will use a different update file. There will be a link inside Daino Notes for notes-foss.com, if users wish to use the FOSS version.

The reason for the rebranding is to streamline all my product offerings (current and future) under one brand name (Daino).

Let me know if you have any questions/suggestions.

guihkx:
>The current user base of Notes will get an update to Daino Notes.

Just to be sure, you mean paid users of current "Notes PRO", correct?

>The new Notes FOSS app will use a different update file.

I'm not sure I understand the reason behind this change. Won't it completely break the update checking feature?

By the way, I personally think these changes warrant a `$minor` version bump, instead of just a `$patch` one.

nuttyartist:
> By the way, I personally think these changes warrant a $minor version bump, instead of just a $patch one.

Yes, you're right. I'll update this.

> Just to be sure, you mean paid users of current "Notes PRO", correct?
> I'm not sure I understand the reason behind this change. Won't it completely break the update checking feature?

No, all current users of Notes will get an update to Daino Notes (apart from those for whom it's disabled, like those from the Snap Store or FlatHub, if I remember correctly). It will provide links to notes-foss.com if users wish to continue using the FOSS version (i.e., this repo).

I'm releasing Daino Notes in the coming days (maybe even tomorrow) so that will take effect very soon.

> Lastly, can we provide feedback on the changes already, or do you prefer us to wait until the WIP tag is removed?

Yes, you can provide feedback, for sure.

Current TODO:

- Linting
- Bump to minor version
- New update file

guihkx:
>No, all current users of Notes will get an update to Daino Notes (apart from those for whom it's disabled, like those from the Snap Store or FlatHub, if I remember correctly). It will provide links to notes-foss.com if users wish to continue using the FOSS version (i.e., this repo).

Got it, but isn't Daino Notes fully proprietary, though?

If that's the case, I don't think it'd be ethical from us to sneak in an update that will replace a FOSS app by a proprietary one...

zjeffer:
> If that's the case, I don't think it'd be ethical from us to sneak in an update that will replace a FOSS app by a proprietary one...

I fully agree with this: I think many users use Notes *because* it's open-source. Auto-updating to a proprietary version would piss off a lot of people.

nuttyartist:
Current users who will update from Notes to Daino Notes will be greeted with a screen that will explain about the change and will allow them to download the FOSS version at any time. Since Daino Notes and Notes FOSS (this repo) share exactly the same database, no change to their data will take effect.

Also, Daino Notes will comply with the MPL license of Notes so all the code that is used from Notes will be uploaded to https://github.com/nuttyartist/daino-notes-public.

Even though Notes is currently open source, it has always been and still is a project that I've been trying to make a living from. This proved difficult. So, making a clear distinction between the FOSS vision and the non-FOSS will probably be better in the long run.

81% of the visits to the website comes from Google after searching for the keyword "notes". So most users of Notes probably don't know what open source means and don't even understand that Notes is open source at all. They are lay people who are looking for a place to jot down their thoughts. So while it's true there are **_many_** who use Notes **_because_** it's open source, it's quite certain that **_most_** users **_don't_**.

I do understand that it's in a tricky ethical category for users who downloaded Notes **because** it's open source, but **they always have the option to switch to the FOSS version**, and **they are most certainly a minority in the user base**, as it stands.

guihkx:
> Even though Notes is currently open source, it has always been and still is a project that I've been trying to make a living from. This proved difficult. So, making a clear distinction between the FOSS vision and the non-FOSS will probably be better in the long run.

There's no problem in trying to make money off of open source, I just think the current approach is not the best.

I'd actually be more okay with a pop up nagging Notes FOSS users to try out Daino Notes, than actually forcing it on them without them knowing or consenting to it first.

> Also, Daino Notes will comply with the MPL license of Notes so all the code that is used from Notes will be uploaded to https://github.com/nuttyartist/daino-notes-public.

That's great, but it's precisely the proprietary, kept-private code that will concern people. As opposed to Notes FOSS, people obviously can't know what Daino Notes does because not all the code is auditable.

> So while it's true there are **_many_** who use Notes **_because_** it's open source, it's quite certain that **_most_** users **_don't_**.

I don't know, even if that's true - we can only know for sure if we actually ask them -, I still think doing this is really shady and a breach of trust, regardless of the target being your average Joe or a FOSS expert.

I can even imagine the more technical people eventually noticing this was done, and being very vocal about it in places like Reddit. That would inevitably inflict heavy damage on this project and likely your other projects and brands, too...

nuttyartist:
> That's great, but it's precisely the proprietary, kept-private code that will concern people. As opposed to Notes FOSS, people obviously can't know what Daino Notes does because not all the code is auditable.

Most (if not all) of the code that connects to the internet will be open source, if that's the part of the code that you think will concern people.

> I don't know, even if that's true - we can only know for sure if we actually ask them -, I still think doing this is really shady and a breach of trust, regardless of the target being your average Joe or a FOSS expert.

Highly likely the average user didn't even know that the app was open source, so I don't agree with your statement in that scenario - they are most likely not operating on different information than before the update. I do admit that this decision is ethically not optimal for users who downloaded Notes because it's FOSS. But these users can easily download the FOSS version with completely no affect to their data.

nuttyartist:
BTW, commit https://github.com/nuttyartist/notes/pull/678/commits/df7e4db766fd318c320de467a114565b7bcd47ee is necessary to make the license for Notes FOSS work (since I'm the old Notes Lemon Squeezy store for Daino Notes). So the endpoints at daino.io/notes_foss testing the licenses on my own DB (hosted on Supabase). To be fully FOSS, I'll of course open source these end points as well, soon enough.

PR #691, opened 2024-08-31
Fix HideToTray checkbox not being synced properly with database at boot

zjeffer:
Fixes a bug where at startup, the 'checked' state of the hideToTray option would always be false. Checking the box would then also disable the tray icon, while unchecking the box would enable it.
This happened because `m_hideToTray` would only be loaded from the database in the `restoreState()`, which is called after the `hideToTrayAction->setChecked(m_hideToTray);` in `setupGlobalSettingsMenu()`. The initial state would therefore always be false, even if the user has `true` in their database.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -138,6 +138,7 @@ MainWindow::MainWindow(QWidget *parent)
       m_userLicenseKey(QStringLiteral("")),
       m_mainMenu(nullptr),
       m_buyOrManageSubscriptionAction(new QAction(this)),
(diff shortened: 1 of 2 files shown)


guihkx (approved):
Yep, the issue exists here on both 2.2.1 and 2.3.0, and your patch fixes it.

Thanks!

nuttyartist:
That was an annoying issue I forgot to document (or maybe I thought it's a bug in Qt). Thanks for that @zjeffer!

PR #698, opened 2024-09-29
CI/Linux: Build/publish Docker images used to create RPM packages

guihkx:
This follows what was done in PRs #695 and #697, but now for RPM packages.

PR #703, opened 2024-09-30
CI/Linux: Include Linux distro version in RPM's artifact name

guihkx:
This is a minor follow-up fix to #698.

After that PR, build artifacts for Fedora and openSUSE lacked the distro version at the end of their file names, i.e.:

- `notes-2.3.0-1.x86_64-qt6-fedora-38` wrongly became `notes-2.3.0-1.x86_64-qt6-fedora`
- `notes-2.3.0-1.x86_64-qt6-opensuse-leap-15.5` wrongly became `notes-2.3.0-1.x86_64-qt6-opensuse-leap`

---

This PR also includes an important change that detects changes to build scripts used in Docker images, and rebuilds these images when necessary.


guihkx:
Force-pushed to adjust the name of a build step more accurately...

It went from **Check if Dockerfile has been modified** to **Check if Dockerfile or build script has been modified**.

PR #712, opened 2024-10-20
CI/macOS: Run `create-dmg` as root

### LINKED ISSUE #711: Do something about flaky macOS builds
Upstream issue: https://github.com/actions/runner-images/issues/7522

Despite the failure being technically random, our macOS CI/CD has been failing quite a lot recently, which requires us to manually trigger rebuilds until it passes, and that's quite annoying.

We trigger that failure because we use [`create-dmg`](https://github.com/create-dmg/create-dmg) to create DMGs for macOS, and they invoke `hdiutil` internally, which then might or might not trigger the error: `hdiutil: create failed - Resource busy`.

The [suspicion](https://github.com/actions/runner-images/issues/7522#issuecomment-1556766641) is that a macOS 13+ service called "XProtectBehaviorService" is causing the failure.

People have been suggesting some workarounds on the issue I linked above:

- Trying to kill the `XProtectBehaviorService` service before invoking `create-dmg` or `hdiutil`
- Invoking `hdiutil` or `create-dmg` as root

What I gathered from the comments is that the first option does not work reliably. The second option I haven't seen any negative comments about, so I might try that first.

There's a third option, which is far more drastic, that would be stop distributing Notes as DMG, and simply distribute the bundle (i.e., the `.app` folder) compressed as a zip file.

But I'm not a macOS user, so it's not obvious to me if the average macOS user would know how to install Notes like that.

guihkx:
Fixes #711 (hopefully).

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -187,7 +187,9 @@ jobs:
       - name: Build dmg (${{ matrix.build-type }})
         run: |
           cd build
-          create-dmg \
+          # FIXME: No, we're not crazy. We invoke create-dmg as root to work around this issue:
+          #       https://github.com/actions/runner-images/issues/7522
+          sudo create-dmg \
               --no-internet-enable \
               --format ULFO \
               --background ../packaging/macos/dmg-background.png \


guihkx:
No errors in this particular CI run on macOS. This "fix" looks promising, but only time will tell...

@nuttyartist Are you okay with this?

nuttyartist:
Let's give it a shot. Is there a major downside to using `root` here?

guihkx:
> s there a major downside to using `root` here?

Well, we already give `create-dmg` full control over our macOS binaries even without root, so I don't think root will have any meaningful impact here.

nuttyartist:
Gotcha 👍

PR #718, opened 2024-10-25
[qt5 branch] CI: Only generate Qt 5 build artifacts

guihkx:
We're on the "qt5" branch, it doesn't make sense to generate Qt 6 build artifacts anymore.

This also drops the x64 build for Windows, leaving only the x86 build.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,9 +3,9 @@ name: CI
 on:
   push:
     branches:
-      - master
+      - qt5
     tags:
-      - v*
+      - v2.3.*
   pull_request:
   workflow_dispatch:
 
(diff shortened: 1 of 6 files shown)


guihkx:
Hmm, I should also add a restriction to prevent the Qt 5 CI/CD from running when pushing tags...

I think we can still allow it to run when `v2.3.*` tags are pushed, right?

nuttyartist:
> We're on the "qt5" branch, it doesn't make sense to generate Qt 6 build artifacts anymore.

👍 

> This also drops the x64 build for Windows, leaving only the x86 build.

👍 

> I think we can still allow it to run when v2.3.* tags are pushed, right?

Yep.

zjeffer (approved):

PR #719, opened 2024-10-25
Avoid infinite repaints upon hovering the folders/tags panels

### LINKED ISSUE #689: High CPU usage when hovering over folders & tags
On latest git version, running with Qt 6.7.2 on Hyprland. Happens on both native Wayland and XWayland.

Watch the CPU usage of Notes on the right.
* On idle, notes uses 0% CPU.
* Hovering over text in the editor uses 5-10%
* Hovering over the list of notes uses 5-10%
* Hovering over `All notes` and `Trash` uses 0-5%
* Hovering over the folders and tags uses 30-50%

https://github.com/user-attachments/assets/f42009fb-358d-4e17-a685-4e4806fd1989

guihkx:
We were calling `setStyleSheet()` unconditionally inside the paint event handlers of the folders/tags panels, which would then trigger a new paint event, causing an infinite repaint recursion.

Now, we only call `setStylesheet()` if the stylesheet of a widget actually needs to be updated.

Fixes #689


nuttyartist:
Good catch @zjeffer, and thanks @guihkx for the fix!

PR #720, opened 2024-10-25
`[qt5]` Avoid infinite repaints upon hovering the folders/tags panels

guihkx:
Same commit of PR #719, cherry picked for the `qt5` branch.


nuttyartist:
👍

PR #723, opened 2024-11-03
refactor: use arg() to concatenate strings for better readability

Integral-Tech:


zjeffer (commented):
Perhaps QStringBuilder can be used in some of these cases? https://doc.qt.io/qt-6/qstring.html#more-efficient-string-construction

Integral-Tech:
> Perhaps QStringBuilder can be used in some of these cases? https://doc.qt.io/qt-6/qstring.html#more-efficient-string-construction

Both `QStringBuilder()` and `QStringLiteral().arg()` are efficient I think

> More efficient string construction
Many strings are known at compile time. The QString constructor from C++ string literals will copy the contents of the string, treating the contents as UTF-8. This requires memory allocation and re-encoding string data, operations that will happen at runtime. If the string data is known at compile time, you can use the [QStringLiteral](https://doc.qt.io/qt-6/qstring.html#QStringLiteral) macro or similarly operator""_s to create QString's payload at compile time instead.

nuttyartist:
> Perhaps QStringBuilder can be used in some of these cases? https://doc.qt.io/qt-6/qstring.html#more-efficient-string-construction

> These allocations can be optimized by an internal class QStringBuilder. This class is marked internal and does not appear in the documentation, because you aren't meant to instantiate it in your code. Its use will be automatic, as described below. The class is found in src/corelib/tools/qstringbuilder.cpp if you want to have a look at it.

Seems like it's not needed to instantiate it? Probably Qt optimize this automatically during compilation?

zjeffer:
> Seems like it's not needed to instantiate it? Probably Qt optimize this automatically during compilation?

Indeed, I think you're supposed to use the `%` operator as an alternative to `+`, like it says in the example:

```cpp
#include <QStringBuilder>

QString hello("hello");
QStringView el = QStringView{ hello }.mid(2, 3);
QLatin1StringView world("world");
QString message =  hello % el % world % QChar('!');
```

But I'm not sure what the difference is with using `.args()`. Maybe `%` is more readable? 
I'll approve anyway.

zjeffer (approved):

Issue #215, opened 2018-04-27
Add command line arguments to get information about the app.


theshadowx:
Sometimes the user need to have information about the the app as the current version of the app, where the note.db is stored, how many note store or in trash ...

Using the command line can help to get these information without needing to open the app.

theshadowx:
example : google-chrome --> in terminal type ```google-chrome -h``` and you'll have how to use chrome from terminal

just ideas
```
notes -v   --> give the current version
notes --count
notes --trash-count
notes --list-notes
...
```

nuttyartist:
Oh ok. It won't affect the UX in any way so that's not a problem. And it might be a good thing for power users so that's good.

nuttyartist:
@theshadowx Can you please join the slack group again? [Invite link](https://join.slack.com/t/notes-app-team/shared_invite/enQtODI3ODQ5NDk4OTAwLWU1OTI3NDNhYzE5ODczZGYxZGI2NGRmN2M4MWJmYzUzNzBhMGViODQ3OWI2MDQ4YzM4ZDM1NDc4NjRiMGZkODE).

PR #734, opened 2025-01-20
Commandline parsing to get basic information about the app

zjeffer:
Implements #215 

```
> ./build/notes --help
Usage: ./build/notes [options]
Notes is a simple note-taking application.

Options:
  -h, --help     Displays help on commandline options.
  --help-all     Displays help, including generic Qt options.
  -v, --version  Displays version information.
  --count        Count the number of notes
  --trash-count  Count the number of notes in the trash
  --list-notes   List all notes
```
```
> ./build/notes --version
Notes 2.3.1
```

Creating a draft PR to get initial review and more ideas on what to add to the parser.
I haven't yet implemented the `count`, `trash-count` and `list-notes` options, because I'm not sure how we should best handle those. It would require setting up the DBManager in `main.cpp`, I guess. We could also just merge without those 3 options, so we can at least provide a `--version` argument.


guihkx:
I really like this!

I can't think of any other command line switches we could add at the moment, but I like the ones you came up with.

For the `--version` switch ,I'd follow what most command line apps seem to output on Linux, which would be simply `APP_NAME APP_VERSION` (like you showed in your post, but the code seems to display it differently?).

I'd also add the Qt version there (build version + runtime version), similarly to what we already do in the about window, e.g.:

```
./notes --version
Notes 2.3.1
Qt 6.8.1 (built with Qt 6.8.0)
```

I like showing both Qt versions because since we unfortunately rely on private Qt headers, having mismatched versions can be a source of problems sometimes...

zjeffer:
I think we should also change the CMakeLists file to add the git commit to the app version, only in case of a manual compile (not releases). Otherwise we won't see which commit a user built from.

guihkx:
That should be doable, yeah.

We actually do provide that info, but only if we build with `-DGIT_REVISION=ON`, which is turned off by default.

We could check if the current commit is tagged and prefixed with `v`, and then set `GIT_REVISION=ON` automatically if it's not - unless the user specifically turns it off.

nuttyartist:
Nice! I like where this is going. @zjeffer there are some lint errors, I believe?

zjeffer:
Fixed the lint errors by rebasing after merging my fixes PR. 

I'll look into making the APP_VERSION dynamic when I have some more time.

PR #744, opened 2025-02-09
Cppcheck & clang-tidy fixes

zjeffer:
Command used:
```
cppcheck --enable=all --inconclusive --force --inline-suppr --library=qt --project=build/compile_commands.json --suppress="*:3rdParty*" --suppress="*:build/*" -i 3rdParty/ -i build --suppress=missingIncludeSystem --suppress=missing
Include --suppress=useStlAlgorithm --quiet -j$(nproc) --disable=unusedFunction --check-level=exhaustive  2>&1 | tee cppcheck.txt
```
All warnings are now fixed. 

The command has 3 suppresions:
* `--suppress=missingIncludeSystem`: because it can't find any Qt system libraries. Not sure how to fix this, we already pass `--library=qt`...
* `--suppress=missingInclude`: same reason
* `--suppress=useStlAlgorithm`: it often suggests changing a simple for-loop to a much less readable std::transform call.

### DIFF
--- a/src/allnotebuttontreedelegateeditor.cpp
+++ b/src/allnotebuttontreedelegateeditor.cpp
@@ -76,8 +76,8 @@ void AllNoteButtonTreeDelegateEditor::paintEvent(QPaintEvent *event)
 #else
     int iconPointSizeOffset = -4;
 #endif
-    painter.setFont(FontLoader::getInstance().loadFont("Material Symbols Outlined", "",
-                                                       16 + iconPointSizeOffset));
+    painter.setFont(
+            font_loader::loadFont("Material Symbols Outlined", "", 16 + iconPointSizeOffset));
     painter.drawText(iconRect, iconPath); // folder
 
     if (m_view->selectionModel()->isSelected(m_index)) {
--- a/src/allnotebuttontreedelegateeditor.cpp
+++ b/src/allnotebuttontreedelegateeditor.cpp
@@ -8,21 +8,15 @@
 #include "editorsettingsoptions.h"
 #include "fontloader.h"
 
-AllNoteButtonTreeDelegateEditor::AllNoteButtonTreeDelegateEditor(QTreeView *view,
-                                                                 const QStyleOptionViewItem &option,
-                                                                 const QModelIndex &index,
-                                                                 QListView *listView,
-                                                                 QWidget *parent)
+AllNoteButtonTreeDelegateEditor::AllNoteButtonTreeDelegateEditor(QTreeView *view, const QStyleOptionViewItem &option, const QModelIndex &index,
+                                                                 QListView *listView, QWidget *parent)
     : QWidget(parent),
       m_option(option),
       m_index(index),
 #ifdef __APPLE__
-      m_displayFont(QFont(QStringLiteral("SF Pro Text")).exactMatch()
-                            ? QStringLiteral("SF Pro Text")
-                            : QStringLiteral("Roboto")),
+      m_displayFont(QFont(QStringLiteral("SF Pro Text")).exactMatch() ? QStringLiteral("SF Pro Text") : QStringLiteral("Roboto")),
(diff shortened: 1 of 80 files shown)


src/allnotebuttontreedelegateeditor.cpp
```diff
@@ -76,8 +76,8 @@ void AllNoteButtonTreeDelegateEditor::paintEvent(QPaintEvent *event)
 #else
     int iconPointSizeOffset = -4;
 #endif
-    painter.setFont(FontLoader::getInstance().loadFont("Material Symbols Outlined", "",
-                                                       16 + iconPointSizeOffset));
+    painter.setFont(
+            font_loader::loadFont("Material Symbols Outlined", "", 16 + iconPointSizeOffset));
```
nuttyartist: Why use snake_case here?

nuttyartist:
Wow so many fixes...The branch seems to work in my testing. The `lqtutils_enum` file is failing during lint like always (I usually discard clang-format changes to this file since the lint doesn't like its formatting for some reason).

src/allnotebuttontreedelegateeditor.cpp
zjeffer: That's a setting in our .clang-tidy file: namespaces should be snake case. We can change this if you want, but personally I prefer that because it makes a clear distinction between namespaces and classes, without having to look at the implementation. If we would call it `FontLoader` instead, you'd think we're calling a static method in a class, and not a standalone function from a namespace.

zjeffer:
In https://github.com/nuttyartist/notes/pull/744/commits/13a1f6b9207cf26fbdc48fe759e8bde75cc28058, I opted to replace the SpecialNodeId namespace + Value enum with simpler `auto constexpr variables` (see nodedata.h). I think it's cleaner this way.

zjeffer:
Added some clang-tidy fixes in latest commits (sorry :smile:)

src/allnotebuttontreedelegateeditor.cpp
nuttyartist: Gotcha 👍

guihkx:
This seems like a great thing to have on our CI/CD as well! Do you plan to do that in this PR? 

On a side note, the openSUSE job is spewing a few compiler warnings in the build step. Can you a look at those, please?

nuttyartist:
> This seems like a great thing to have on our CI/CD as well! Do you plan to do that in this PR?
> 
> On a side note, the openSUSE job is spewing a few compiler warnings in the build step. Can you a look at those, please?

That would be a really great addition to our CI indeed.

zjeffer:
> On a side note, the openSUSE job is spewing a few compiler warnings in the build step. Can you a look at those, please?

Fixed it. Not sure why my system and the other jobs don't show that warning though. `-Wpedantic` should be enabled when using GCC or Clang, I'm using clang and I didn't get the error :thinking:

guihkx:
> Fixed it. Not sure why my system and the other jobs didn't show that warning though.

Must be some quirk/bug of GCC 10 — which only openSUSE is using. Thanks for fixing it, though. :)

As for all the code changes, I didn't go through *all* of them, but even if I did, I wouldn't understand most of them anyway, because most C++ concepts are still voodoo to me. :sweat_smile: 

Anyway, I did at least test the AppImage on Linux and it seems to be working okay.

I also went to test the Windows build, and got hit by [a regression](https://github.com/nuttyartist/notes/issues/745) unrelated to this PR, but after reverting the offending commit, I  verified that the Windows build also seems to be working okay.

Still, I'm deferring the actual review to @nuttyartist. But if I may nitpick, I'd just squash up some of these commits for a cleaner merge.

nuttyartist (approved):
Seems to work well on my machine (macoS). @zjeffer thanks for this! Ready to merge imo, your call on whether to squash.

PR #749, opened 2025-04-24
Add macOS-specific keyboard shortcuts for note navigation

### LINKED ISSUE #186: Command + arrows doesn't work on Mac
The keyboard shortcuts to scroll between notes no matter where you're - 'Command + arrows', doesn't work on Mac.

Ctrl + L does work.

ishaan-builds:
## Description
This PR adds macOS-specific keyboard shortcuts for note navigation and improves the overall keyboard navigation experience. The changes ensure that keyboard shortcuts work consistently across different platforms and match user expectations on macOS.

### Changes Made
1. Added macOS-specific keyboard shortcuts:
   - `Control + ↑` to select note above   
   - `Control + ↓` to select note below
   - `Control + L` to focus on the text editor

2. Updated keyboard shortcuts documentation to reflect these changes.

## Testing
- Tested on macOS Sequoia (15.3)
- Verified that shortcuts work when:
  - Text editor is focused
  - List view is focused
  - Search bar is focused

## Related Issue
Fixes #186  - Command + arrows doesn't work on Mac

## Notes
The changes should maintain backward compatibility while improving the user experience for macOS users. The implementation uses the Control (called META in code) key, which is the standard approach for cross-platform applications.

This fix currently requires that the user disable the default macOS behavior of using `Control + ↑` for Mission Control and `Control + ↓` to show Application Windows by doing the following:
1. Open System Preferences -> Keyboard -> Keyboard Shortcuts... ->  Mission Control
2. Untick the `Mission Control` and `Application Windows`

### DIFF
--- a/docs/keyboard_shortcuts.md
+++ b/docs/keyboard_shortcuts.md
@@ -24,10 +24,10 @@ These shortcuts are meant to be used while the Notes window is currently focused
 | <kbd>Ctrl</kbd> + <kbd>E</kbd>                        | Clear the search bar                                          |
 | <kbd>Ctrl</kbd> + <kbd>↓</kbd> (or just <kbd>↓</kbd>) | Select note below                                             |
 | <kbd>Ctrl</kbd> + <kbd>↑</kbd> (or just <kbd>↑</kbd>) | Select note above                                             |
-| <kbd>Ctrl</kbd> + <kbd>↓</kbd> (macOS)             | Select note below                                             |
-| <kbd>Ctrl</kbd> + <kbd>↑</kbd> (macOS)             | Select note above                                             |
+| <kbd>Ctrl</kbd> + <kbd>↓</kbd> _(macOS)_              | Select note below                                             |
+| <kbd>Ctrl</kbd> + <kbd>↑</kbd> _(macOS)_              | Select note above                                             |
 | <kbd>Ctrl</kbd> + <kbd>Enter</kbd>                    | Focus on the text editor                                      |
-| <kbd>Ctrl</kbd> + <kbd>L</kbd> (macOS)             | Focus on the text editor                                      |
+| <kbd>Ctrl</kbd> + <kbd>L</kbd> _(macOS)_              | Focus on the text editor                                      |
 | <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>F</kbd>     | Toggle full screen mode                                       |
 | <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>K</kbd>     | Toggle kanban view                                            |
 | <kbd>F11</kbd>                                        | Toggle full screen mode                                       |
--- a/src/listviewlogic.cpp
+++ b/src/listviewlogic.cpp
@@ -191,7 +191,7 @@ void ListViewLogic::selectNoteUp()
 {
     // Ensure the list view has focus
     m_listView->setFocus();
-    
+
     auto currentIndex = m_listView->currentIndex();
     if (currentIndex.isValid()) {
         int currentRow = currentIndex.row();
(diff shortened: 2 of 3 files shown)


docs/keyboard_shortcuts.md
```diff
@@ -24,7 +24,10 @@ These shortcuts are meant to be used while the Notes window is currently focused
 | <kbd>Ctrl</kbd> + <kbd>E</kbd>                        | Clear the search bar                                          |
 | <kbd>Ctrl</kbd> + <kbd>↓</kbd> (or just <kbd>↓</kbd>) | Select note below                                             |
 | <kbd>Ctrl</kbd> + <kbd>↑</kbd> (or just <kbd>↑</kbd>) | Select note above                                             |
+| <kbd>Ctrl</kbd> + <kbd>↓</kbd> (macOS)             | Select note below                                             |
```
guihkx, line 27: Nitpick: Please align the `|` character with the lines above and below, i.e.:

```suggestion
| <kbd>Ctrl</kbd> + <kbd>↓</kbd> (macOS)                | Select note below                                             |
```

src/listviewlogic.cpp
guihkx, line 1: This file needs to be clang-formatted. Please run `clang-format -style=file -i src/listviewlogic.cpp`.

src/mainwindow.cpp
guihkx, line 1: This file also needs to be clang-formatted. Please run `clang-format -style=file -i src/mainwindow.cpp`.

guihkx (changes requested):
I just have a few nitpicks....

As for the changes themselves, I'll defer the review to @nuttyartist who is an actual Mac user.

docs/keyboard_shortcuts.md
zjeffer, line 27: I don't use a mac but shouldn't it be Command instead of Ctrl?

src/mainwindow.cpp
```diff
@@ -532,14 +532,22 @@ void MainWindow::setupKeyboardShortcuts()
     new QShortcut(QKeySequence(Qt::CTRL | Qt::Key_D), this, SLOT(deleteSelectedNote()));
     new QShortcut(QKeySequence(Qt::CTRL | Qt::Key_F), m_searchEdit, SLOT(setFocus()));
     new QShortcut(QKeySequence(Qt::CTRL | Qt::Key_E), m_searchEdit, SLOT(clear()));
+    
+    // Add macOS-specific shortcuts for note navigation
```
zjeffer, line 536:
```suggestion
    // shortcuts for note navigation
```
These are not just macOS specific ;)

docs/keyboard_shortcuts.md
ishaan-builds: Since Command + ↑ or ↓ is used to jump to the top or bottom of a document in almost every app, using Control makes more sense imo. Only problem is that it requires changing a setting through System Preferences on the user's end to work without MacOS taking the keystrokes first.

zjeffer: Yes, but in your code you used Meta + Up/Down, that's what I'm confused about. Isn't `Meta` `Command`?

nuttyartist: @zjeffer, @ishaan-builds is correct - `META` on macOS means `control` (`^`).

nuttyartist:
Thanks for contributing!

I must note that I've tested your branch but it doesn't work - probably since built in default shortcuts override your implementation. Here's a video of me pressing control + arrow_down/arrow_up.

https://github.com/user-attachments/assets/31d5ac12-af64-49c8-b095-f0a727f24848

I'm running macOS 15.4.1.

nuttyartist:
I'm happy you chose our project for your first time (:

It's important that our users will not need to change such settings for using our app. So I believe our current implementation (allow focusing the notes list using cmd + shift + L and then use arrow keys) is better. But in the future we'll have a proper settings window that will allow changing keyboard shortcuts, so that's where people could use an approach like yours.

PR #755, opened 2025-08-31
CI/Windows: Update from Windows 2019 to Windows 2025

guihkx:
The Windows 2019 runner has been deprecated: https://redirect.github.com/actions/runner-images/issues/12045

This also installs Inno Setup manually, because it's not included in the Windows 2025 image.

I've tested the installer on Windows 10, things seem fine.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -14,7 +14,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - os: windows-2019
+          - os: windows-2025
             qt-version: 6.4.3
             build-type: release
     outputs:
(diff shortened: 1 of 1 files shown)


guihkx:
>BTW, does this change affect the versions of Windows we support?

Nope!

This essentially just updates the old redistributables from Visual Studio 2019 (V142) to newer ones, from Visual Studio 2022 (V143).

The V143 redistributables are even supported on Windows 7 (which we've stopped supporting after switching to Qt-6-only), so we should be good.

More links if you're interested:

- https://github.com/actions/runner-images/blob/main/images/windows/Windows2025-Readme.md#visual-studio-enterprise-2022
- https://learn.microsoft.com/en-us/lifecycle/faq/visual-c-faq#what-are-the-support-dates-for-visual-c---redistributable--msvc-runtime-libraries--and-msvc-toolset-versions-2015--2017--2019--2022--and-later-
- https://learn.microsoft.com/en-us/cpp/windows/latest-supported-vc-redist?view=msvc-170#notes:~:text=Current%20versions%20of%20the%20Visual%20C%2B%2B%20Redistributable%20for%20Visual%20Studio%202015%2D2022%20only%20support%20Windows%207%2C%208.1%2C%2010%2C%20and%2011

PR #761, opened 2025-10-17
CI/macOS: Update to macOS 15

guihkx:
The macOS 13 runner is getting removed by GitHub in December 4th, 2025:

https://github.blog/changelog/2025-09-19-github-actions-macos-13-runner-image-is-closing-down/

I decided to go with the `-intel` variant of macOS 15, because I think it makes sense to build for x86_64 for as long as we can, even though Apple is also retiring the architecture.

While I don't have a Mac to actually test the resulting binary, I've extracted the `Notes` binary from the DMG image on Linux and ran the `file` command on it, which looks okay:

```sh
$ 7z e Notes.dmg 'Notes/Notes Better.app/Contents/MacOS/Notes'
$ file Notes
Notes: Mach-O universal binary with 2 architectures: [x86_64:\012- Mach-O 64-bit x86_64 executable, flags:<NOUNDEFS|DYLDLINK|TWOLEVEL|WEAK_DEFINES|BINDS_TO_WEAK|PIE>] [\012- arm64]
```

Still, it'd be great if @nuttyartist could test the resulting binary on both x86_64 and ARM64 Macs (if possible).

Thanks!

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -25,7 +25,7 @@ jobs:
 
       - name: Install Qt 6 (homebrew)
         run: |
-          brew install ninja qt@6 
+          brew install qt@6
 
       - name: Setup CLang problem matcher
         # Technically, this action only supports GCC, but it seems to work well for Clang too.
@@ -96,7 +96,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          brew install create-dmg ninja
+          brew install create-dmg
 
       - name: Build (${{ matrix.build-type }})
         env:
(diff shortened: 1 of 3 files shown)


nuttyartist:
@guihkx thanks!

Works on ARM macOS. My Intel MacBook got its batteries swelled up so I'm afraid of using it haha.

guihkx:
Hahah, no problem. Thanks for testing on ARM, though.

I managed to test it on an virtualized macOS 11 on x86-64, and it also works:

image

zjeffer (approved):

guihkx:
Thanks, I forgot about this one. lol

PR #769, opened 2026-05-11
Fix note list rendering at fractional display scaling

### LINKED ISSUE #164: Bug - notes are not aligned well on some reported Windows machines

Shriram-Vasudevan:
## Summary
- set Qt high-DPI scale factor rounding to pass through before creating the application
- create note-list delegate paint buffers with the owning widget/window device-pixel ratio
- use logical buffer geometry while painting cropped row backgrounds/labels, so Windows fractional scaling does not mix physical pixmap pixels with logical item rects

Fixes #164.

## Verification
- `git diff --check`
- `c++ -std=c++17 -fsyntax-only -include arm_acle.h ... src/notelistdelegate.cpp src/notelistdelegateeditor.cpp` against locally installed Qt 6.9.2 headers
- syntax-only check for `QGuiApplication::setHighDpiScaleFactorRoundingPolicy(...)` against Qt 6.9.2 headers

I could not run the full CMake build locally because `cmake` is not installed in this environment, and `clang-format` is also unavailable locally.

### DIFF
--- a/src/notelistdelegate.cpp
+++ b/src/notelistdelegate.cpp
@@ -294,11 +294,11 @@ QTimeLine::State NoteListDelegate::animationState()
 void NoteListDelegate::paintBackground(QPainter *painter, const QStyleOptionViewItem &option, const QModelIndex &index) const
 {
     auto bufferSize = bufferSizeHint(option, index);
-    QPixmap buffer{ bufferSize };
+    QPixmap buffer = utils::makeDevicePixelRatioPixmap(bufferSize, m_view);
     buffer.fill(Qt::transparent);
     QPainter bufferPainter{ &buffer };
     bufferPainter.setRenderHints(QPainter::Antialiasing | QPainter::SmoothPixmapTransform);
-    QRect bufferRect = buffer.rect();
+    QRect bufferRect{ QPoint{}, bufferSize };
     auto isPinned = index.data(NoteListModel::NoteIsPinned).toBool();
     auto const *model = static_cast<NoteListModel *>(m_view->model());
     if (model->hasPinnedNote() && model->isFirstPinnedNote(index) && static_cast<NoteListView *>(m_view)->isPinnedNotesCollapsed()) {
@@ -419,7 +419,7 @@ void NoteListDelegate::paintLabels(QPainter *painter, const QStyleOptionViewItem
 {
     if (m_animatedIndexes.contains(index)) {
         auto bufferSize = bufferSizeHint(option, index);
-        QPixmap buffer{ bufferSize };
+        QPixmap buffer = utils::makeDevicePixelRatioPixmap(bufferSize, m_view);
         buffer.fill(Qt::transparent);
         QPainter bufferPainter{ &buffer };
         bufferPainter.setRenderHints(QPainter::Antialiasing | QPainter::SmoothPixmapTransform);
--- a/src/notelistdelegate.cpp
+++ b/src/notelistdelegate.cpp
@@ -298,7 +298,7 @@ void NoteListDelegate::paintBackground(QPainter *painter, const QStyleOptionView
     buffer.fill(Qt::transparent);
     QPainter bufferPainter{ &buffer };
     bufferPainter.setRenderHints(QPainter::Antialiasing | QPainter::SmoothPixmapTransform);
-    QRect bufferRect{ QPoint{}, bufferSize };
+    QRect bufferRect{ QPoint{ }, bufferSize };
     auto isPinned = index.data(NoteListModel::NoteIsPinned).toBool();
     auto const *model = static_cast<NoteListModel *>(m_view->model());
     if (model->hasPinnedNote() && model->isFirstPinnedNote(index) && static_cast<NoteListView *>(m_view)->isPinnedNotesCollapsed()) {
--- a/src/notelistdelegateeditor.cpp
+++ b/src/notelistdelegateeditor.cpp
@@ -135,11 +135,11 @@ NoteListDelegateEditor::~NoteListDelegateEditor()
 void NoteListDelegateEditor::paintBackground(QPainter *painter, const QStyleOptionViewItem &option, const QModelIndex &index) const
 {
     auto bufferSize = rect().size();
-    QPixmap buffer{ bufferSize };
+    QPixmap buffer = utils::makeDevicePixelRatioPixmap(bufferSize, this);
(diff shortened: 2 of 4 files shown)


zjeffer:
@nuttyartist Do we have an AI-policy? Personally I'm not a big fan of these fully AI-generated PRs, I mean the OP didn't even bother to let the AI install cmake:

> I could not run the full CMake build locally because cmake is not installed in this environment, and clang-format is also unavailable locally.

I recently added one in a personal repo if you want an example: https://github.com/zjeffer/split-monitor-workspaces/blob/main/CONTRIBUTING.md

Shriram-Vasudevan:
I installed the missing local build tooling and pushed follow-up commit 404e7cf with the clang-format adjustments found by the repo format gate.

Verification now run locally:
- `/opt/homebrew/opt/llvm/bin/clang-format -style=file --dry-run --Werror --color=false src/main.cpp src/notelistdelegate.cpp src/notelistdelegateeditor.cpp src/utils.h` -> passed
- `find src/ -iregex '.*\.\(cpp\|h\|mm\)$' -print0 | xargs -n 1 -0 /opt/homebrew/opt/llvm/bin/clang-format -style=file --dry-run --Werror --color=false` -> passed
- `cmake . --warn-uninitialized --warn-unused-vars -G Ninja -B build -DCMAKE_BUILD_TYPE=debug -DCMAKE_PREFIX_PATH="$(brew --prefix qt@6)" -DPRO_VERSION=OFF` -> passed with Qt 6.9.2
- `cmake --build build` -> passed
- `ctest --test-dir build --output-on-failure` -> no tests registered
- `git diff --check` -> passed

The build produced only existing/project-wide Qt deprecation and macOS deployment-target linker warnings, not errors in the touched note-list files.

src/utils.h
```diff
@@ -1,8 +1,13 @@
 #pragma once
 
+#include <QGuiApplication>
 #include <cmath>
-#include <QString>
 #include <QDateTime>
+#include <QPixmap>
```
nuttyartist, line 6: This helper will also need `QRect` for converting logical source rects to pixmap/device-pixel source rects.

```suggestion
#include <QPixmap>
#include <QRect>
```

src/utils.h
```diff
@@ -44,4 +49,29 @@ inline QString parseDateTime(const QDateTime &dateTime)
...
+    return qreal(1);
+}
+
+inline QPixmap makeDevicePixelRatioPixmap(const QSize &logicalSize, const QWidget *widget)
+{
+    const qreal scale = devicePixelRatioForWidget(widget);
+    QPixmap pixmap(logicalSize * scale);
+    pixmap.setDevicePixelRatio(scale);
+    return pixmap;
+}
```
nuttyartist, line 75: Could we pair the DPR pixmap allocator with a small source-rect conversion helper? Painting into the pixmap stays in logical coordinates, but `QPainter::drawPixmap(..., sourceRect)` reads `sourceRect` in pixmap/device pixels.

```suggestion
inline QPixmap makeDevicePixelRatioPixmap(const QSize &logicalSize, const QWidget *widget)
{
    const qreal scale = devicePixelRatioForWidget(widget);
    QPixmap pixmap(logicalSize * scale);
    pixmap.setDevicePixelRatio(scale);
    return pixmap;
}

inline QRect devicePixelRatioPixmapSourceRect(const QRect &logicalRect, const QPixmap &pixmap)
{
    const qreal scale = pixmap.devicePixelRatio();
    const QRectF scaledRect(logicalRect.x() * scale, logicalRect.y() * scale, logicalRect.width() * scale, logicalRect.height() * scale);
    return scaledRect.toAlignedRect();
}
```

src/notelistdelegate.cpp
```diff
@@ -294,11 +294,11 @@ QTimeLine::State NoteListDelegate::animationState()
 void NoteListDelegate::paintBackground(QPainter *painter, const QStyleOptionViewItem &option, const QModelIndex &index) const
 {
     auto bufferSize = bufferSizeHint(option, index);
-    QPixmap buffer{ bufferSize };
+    QPixmap buffer = utils::makeDevicePixelRatioPixmap(bufferSize, m_view);
```
nuttyartist, line 297: Once this buffer is DPR-aware, the `drawPixmap(..., buffer, QRect{...})` calls lower in `paintBackground()` still need to convert their source rects before drawing. One compact way is to add this near the draw section and route those calls through it:

```cpp
auto drawBuffer = [&buffer, painter](const QRect &targetRect, const QRect &sourceRect) {
    painter->drawPixmap(targetRect, buffer, utils::devicePixelRatioPixmapSourceRect(sourceRect, buffer));
};
```

Then replace each `painter->drawPixmap(..., buffer, QRect{...})` in this function with `drawBuffer(..., QRect{...})`. The same pattern applies to the animated label buffer in `paintLabels()`.

src/notelistdelegateeditor.cpp
```diff
@@ -135,11 +135,11 @@ NoteListDelegateEditor::~NoteListDelegateEditor()
 void NoteListDelegateEditor::paintBackground(QPainter *painter, const QStyleOptionViewItem &option, const QModelIndex &index) const
 {
     auto bufferSize = rect().size();
-    QPixmap buffer{ bufferSize };
+    QPixmap buffer = utils::makeDevicePixelRatioPixmap(bufferSize, this);
```
nuttyartist, line 138: Same coordinate-space issue here: the buffer is now DPR-aware, but the later source rect passed to `drawPixmap()` is still logical. The final draw should convert that source rect first, for example:

```cpp
auto rowHeight = rect().height();
const QRect sourceRect{ 0, bufferSize.height() - rowHeight, rect().width(), rowHeight };
painter->drawPixmap(rect(), buffer, utils::devicePixelRatioPixmapSourceRect(sourceRect, buffer));
```

nuttyartist (commented):
Thanks for the fix. I left a few inline suggestions for the remaining DPR source-rect issue. The helper approach looks good; the main thing is making the later drawPixmap source rectangles use the same coordinate space as the DPR pixmaps.

nuttyartist:
@zjeffer Not really yet, and I agree, we should have. Yours seems reasonable enough, I'll add them to the repo.

@Shriram-Vasudevan Thanks for your PR.
I've added some comments using GPT 5.5, please look at them.
Can you please share before and after screenshots after you make changes?

PR #770, opened 2026-05-13
Stabilize note list text alignment on Windows

### LINKED ISSUE #164: Bug - notes are not aligned well on some reported Windows machines

MizuCiaossu:
## Summary
- use stable font line metrics for note list title, date, content, and parent rows
- stop laying out note rows from per-string `boundingRect()` heights, which vary with the actual glyphs rendered
- keep `NoteListDelegate` and `NoteListDelegateEditor` in sync so the persistent editor matches the normal row renderer

Fixes #164.

## Why
The Windows alignment bug appears to come from using text-specific bounding boxes to position stacked note row labels. On some Windows font/DPI combinations, that can make row spacing depend on the glyphs in each note title/date/content, which leads to visually inconsistent alignment.

## Testing
- Not run locally in this environment because the required Qt/CMake Windows toolchain is not installed here.
- The change is limited to replacing layout height calculations with `QFontMetrics::height()` in the two note-list paint paths.


zjeffer:
@nuttyartist another AI account that has the exact same issues as https://github.com/nuttyartist/notes/pull/769. What should we do with these?

> Not run locally in this environment because the required Qt/CMake Windows toolchain is not installed here.

like seriously wtf

PR #773, opened 2026-06-14
Fix note list row alignment on Windows display scaling

### LINKED ISSUE #164: Bug - notes are not aligned well on some reported Windows machines

AJ0070:
#### Changes

- Added font-metric-based minimum height helpers for note list rows
- Applied those minimum heights in the note list delegate sizing paths
- Updated tagged-note editor sizing to use the same metric-based floor
- Replaced fixed tag-list vertical placement with metric-based placement

#### Why

The Windows issue appears to come from a mismatch between:

- fixed pixel row heights in the custom note list layout
- scaled font rendering on Windows, especially with `Segoe UI` and non-`100%` display scaling

When scaling increases text height, the old fixed values could leave the content taller than the row allocated for it, causing visible misalignment.

#### Tested for these settings

Built and tested locally on Windows.

Verified the note list manually at multiple display scaling levels:

- `100%`
- `125%`
- `150%`
- `175%`

Checked the following scenarios at each scaling level:

- regular note rows
- tagged note rows
- `All Notes` view
- pinned/unpinned note sections

Verified that:

- note titles, dates, and preview text remain aligned
- text does not overlap or get clipped
- tagged rows expand enough to contain tag chips cleanly
- pinned section headers and first rows below them remain properly spaced
- `All Notes` rows still leave enough room for the folder line

Fixes #164

### DIFF
--- a/src/notelistdelegate.cpp
+++ b/src/notelistdelegate.cpp
@@ -98,10 +98,10 @@ int NoteListDelegate::minimumContentHeight() const
     return note_list_constants::TOP_OFFSET_Y + titleHeight + dateHeight + contentHeight + note_list_constants::DATE_DESC_SPACE;
 }
 
-int NoteListDelegate::minimumRowHeight(bool isInAllNotes) const
+int NoteListDelegate::minimumRowHeight() const
 {
     int result = minimumContentHeight() + note_list_constants::LAST_EL_SEP_SPACE;
-    if (isInAllNotes) {
+    if (m_isInAllNotes) {
         const int folderLineHeight = qMax(QFontMetrics(m_titleFont).height(), 16) + note_list_constants::DESC_FOLDER_SPACE;
         result += folderLineHeight;
     }
@@ -243,7 +243,7 @@ QSize NoteListDelegate::sizeHint(const QStyleOptionViewItem &option, const QMode
     }
 
     int yOffsets = secondYOffset + thirdYOffset + fourthYOffset + fifthYOffset;
-    const int minimumHeight = minimumRowHeight(m_isInAllNotes) + yOffsets;
+    const int minimumHeight = minimumRowHeight() + yOffsets;
     if (m_isInAllNotes) {
         result.setHeight(result.height() - 2 + note_list_constants::LAST_EL_SEP_SPACE + yOffsets);
     } else {
@@ -299,7 +299,7 @@ QSize NoteListDelegate::bufferSizeHint(const QStyleOptionViewItem &option, const
     }
 
     int yOffsets = secondYOffset + thirdYOffset + fourthYOffset;
-    const int minimumHeight = minimumRowHeight(m_isInAllNotes) + yOffsets;
+    const int minimumHeight = minimumRowHeight() + yOffsets;
     if (m_isInAllNotes) {
         result.setHeight(result.height() - 2 + note_list_constants::LAST_EL_SEP_SPACE + yOffsets);
     } else {
--- a/src/notelistdelegate.cpp
+++ b/src/notelistdelegate.cpp
@@ -334,7 +334,16 @@ void NoteListDelegate::paintBackground(QPainter *painter, const QStyleOptionView
     QRect bufferRect = buffer.rect();
     auto isPinned = index.data(NoteListModel::NoteIsPinned).toBool();
     auto const *model = static_cast<NoteListModel *>(m_view->model());
-    if (model->hasPinnedNote() && model->isFirstPinnedNote(index) && static_cast<NoteListView *>(m_view)->isPinnedNotesCollapsed()) {
+    const bool isCollapsedPinnedHeader = m_view->isPinnedNotesCollapsed() && model->isFirstPinnedNote(index);
+    const int headerOffset = sectionHeaderHeight(index, *model);
+    if (headerOffset > 0 && !isCollapsedPinnedHeader) {
+        int fifthYOffset = 0;
+        if (!m_view->isPinnedNotesCollapsed() && model->isFirstUnpinnedNote(index)) {
+            fifthYOffset = note_list_constants::LAST_PINNED_TO_UNPINNED_HEADER;
+        }
+        bufferRect.setY(bufferRect.y() + headerOffset + fifthYOffset);
+    }
+    if (model->hasPinnedNote() && isCollapsedPinnedHeader) {
         bufferPainter.fillRect(bufferRect, QBrush(m_defaultColor));
     } else if ((option.state & QStyle::State_Selected) == QStyle::State_Selected) {
         if (qApp->applicationState() == Qt::ApplicationActive) {
--- a/src/notelistdelegate.cpp
+++ b/src/notelistdelegate.cpp
@@ -108,6 +108,15 @@ int NoteListDelegate::minimumRowHeight() const
     return result;
 }
 
+int NoteListDelegate::sectionHeaderHeight(const QModelIndex &index, const NoteListModel &model)
+{
+    if (model.hasPinnedNote() && (model.isFirstPinnedNote(index) || model.isFirstUnpinnedNote(index))) {
+        return note_list_constants::SECTION_HEADER_HEIGHT;
+    }
+
+    return 0;
+}
+
 void NoteListDelegate::setState(NoteListState NewState, QModelIndexList indexes)
 {
     if (animationState() != QTimeLine::NotRunning) {
@@ -197,9 +206,8 @@ QSize NoteListDelegate::sizeHint(const QStyleOptionViewItem &option, const QMode
         if (m_state == NoteListState::MoveIn) {
             result.setHeight(rowHeight);
         } else {
-            double rate = m_timeLine->currentFrame() / (m_maxFrame * 1.0);
-            double height = rowHeight * rate;
-            result.setHeight(int(height));
+            const double rate = m_timeLine->currentFrame() / (m_maxFrame * 1.0);
+            result.setHeight(int(rowHeight * rate));
         }
     } else {
         result.setHeight(rowHeight);
@@ -211,18 +219,18 @@ QSize NoteListDelegate::sizeHint(const QStyleOptionViewItem &option, const QMode
         auto isPinned = note.isPinnedNote();
         if (isPinned) {
             if (noteListModel->isFirstPinnedNote(index)) {
-                result.setHeight(25);
+                result.setHeight(note_list_constants::SECTION_HEADER_HEIGHT);
                 return result;
             }
             result.setHeight(0);
             return result;
         }
         if (noteListModel->hasPinnedNote() && noteListModel->isFirstUnpinnedNote(index)) {
-            result.setHeight(result.height() + 25);
+            result.setHeight(result.height() + note_list_constants::SECTION_HEADER_HEIGHT);
         }
     } else {
         if (noteListModel->hasPinnedNote() && (noteListModel->isFirstPinnedNote(index) || noteListModel->isFirstUnpinnedNote(index))) {
-            result.setHeight(result.height() + 25);
+            result.setHeight(result.height() + note_list_constants::SECTION_HEADER_HEIGHT);
         }
     }
     int secondYOffset = 0;
@@ -243,13 +251,15 @@ QSize NoteListDelegate::sizeHint(const QStyleOptionViewItem &option, const QMode
     }
 
     int yOffsets = secondYOffset + thirdYOffset + fourthYOffset + fifthYOffset;
-    const int minimumHeight = minimumRowHeight() + yOffsets;
+    const int minimumHeight = minimumRowHeight() + sectionHeaderHeight(index, *noteListModel) + yOffsets;
     if (m_isInAllNotes) {
         result.setHeight(result.height() - 2 + note_list_constants::LAST_EL_SEP_SPACE + yOffsets);
     } else {
         result.setHeight(result.height() - 10 + note_list_constants::LAST_EL_SEP_SPACE + yOffsets);
     }
-    result.setHeight(qMax(result.height(), minimumHeight));
+    if (!m_animatedIndexes.contains(index)) {
+        result.setHeight(qMax(result.height(), minimumHeight));
+    }
     return result;
 }
 
@@ -278,7 +288,7 @@ QSize NoteListDelegate::bufferSizeHint(const QStyleOptionViewItem &option, const
         auto isPinned = index.data(NoteListModel::NoteIsPinned).value<bool>();
         if (isPinned) {
             if (noteListModel->isFirstPinnedNote(index)) {
-                result.setHeight(25);
+                result.setHeight(note_list_constants::SECTION_HEADER_HEIGHT);
                 return result;
             }
             result.setHeight(0);
@@ -299,7 +309,7 @@ QSize NoteListDelegate::bufferSizeHint(const QStyleOptionViewItem &option, const
     }
 
     int yOffsets = secondYOffset + thirdYOffset + fourthYOffset;
-    const int minimumHeight = minimumRowHeight() + yOffsets;
+    const int minimumHeight = minimumRowHeight() + sectionHeaderHeight(index, *noteListModel) + yOffsets;
     if (m_isInAllNotes) {
         result.setHeight(result.height() - 2 + note_list_constants::LAST_EL_SEP_SPACE + yOffsets);
     } else {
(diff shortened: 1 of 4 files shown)


AJ0070:
@nuttyartist Can I get review on this?

zjeffer:
Can you add before & after screenshots?

AJ0070:
100% Scale: 

image


125% Scale:

image


150% Scale:

image


175% Scale:

image

src/notelistdelegate.cpp
```diff
@@ -89,6 +89,25 @@ NoteListDelegate::NoteListDelegate(NoteListView *view, TagPool *tagPool, QObject
...
+    const int dateHeight = QFontMetrics(m_dateFont).height();
+    const int contentHeight = titleHeight;
+
+    return note_list_constants::TOP_OFFSET_Y + titleHeight + dateHeight + contentHeight + note_list_constants::DATE_DESC_SPACE;
+}
+
+int NoteListDelegate::minimumRowHeight(bool isInAllNotes) const
+{
+    int result = minimumContentHeight() + note_list_constants::LAST_EL_SEP_SPACE;
+    if (isInAllNotes) {
```
zjeffer: the isAllNotes parameter could be removed, and simply use m_isInAllNotes

AJ0070: done

AJ0070:
The windows failure looks unrelated to my changes.

zjeffer:
Indeed, I just ran the ci/cd on master and it failed with the same issues https://github.com/nuttyartist/notes/actions/runs/27494239891/job/81265236753

zjeffer (approved):

AJ0070:
@nuttyartist gentle ping!

nuttyartist:
Hi there @AJ0070! Thanks for contributing. Going over your PR now!

nuttyartist (changes requested):
Thanks again for the changes. There are some issues. 

### 1. Animation rows no longer collapse/expand correctly

`src/notelistdelegate.cpp`

First, this PR breaks the some of the animations of notes list delegates:

Before:

https://github.com/user-attachments/assets/87e7e353-3664-4320-a2e3-8e8aa6bbb828

After:

https://github.com/user-attachments/assets/cedc052b-b8a2-44df-b151-66f0f5bcf063

**The reason**
The new minimum-height clamp is applied even when `sizeHint()` is returning timeline-scaled heights for insert/remove/move animations. That means an animated row can be clamped back up to `minimumRowHeight()` instead of shrinking toward zero or growing smoothly.

**How to test**
Create, delete, pin, or unpin a note and watch the note-list row animation. The affected row should smoothly grow or collapse. With this change, the row can remain at least `minimumRowHeight()` during the animation.

### 2. Pinned/Notes section header height is missing from the minimum row height

`src/notelistdelegate.cpp`  
`src/notelistdelegateeditor.cpp`

Before:
Image

After:
Image

The painted content is shifted down by 25px for the first pinned row or first unpinned row under the `Pinned` / `Notes` section headers, but the new minimum-height calculation does not include that 25px. At higher Windows scaling, those first rows can still clip or overlap because the row floor only accounts for text/folder/tag content, not the section header above it.

**How to test**

On Windows, use display scaling or text scaling high enough to reproduce the original issue, ideally `150%` or `175%`.

Create notes so there is:

- a first pinned note
- a first unpinned note
- tagged rows
- the `All Notes` view, so the folder line is visible

Then inspect the first pinned and first unpinned rows. These are the rows whose content is shifted down by the section header. If the 25px header is missing from the minimum height, they are the most likely places to see clipped text, clipped tag chips, or overlap.

src/notelistdelegate.cpp
```diff
@@ -224,11 +243,13 @@ QSize NoteListDelegate::sizeHint(const QStyleOptionViewItem &option, const QMode
     }
 
     int yOffsets = secondYOffset + thirdYOffset + fourthYOffset + fifthYOffset;
+    const int minimumHeight = minimumRowHeight() + yOffsets;
     if (m_isInAllNotes) {
         result.setHeight(result.height() - 2 + note_list_constants::LAST_EL_SEP_SPACE + yOffsets);
     } else {
         result.setHeight(result.height() - 10 + note_list_constants::LAST_EL_SEP_SPACE + yOffsets);
     }
+    result.setHeight(qMax(result.height(), minimumHeight));
```
nuttyartist: This clamp also runs for animated rows. In the animated branch above, `result.height()` is intentionally scaled by the timeline so insert/remove/move rows can grow or collapse, but the final `qMax(..., minimumHeight)` raises it back to at least the full minimum height.

Suggested fix: only apply the minimum-height floor when the row is not in `m_animatedIndexes`, or clamp the target row height before applying the timeline rate.

How to test: create, delete, pin, or unpin a note and watch the affected row. It should smoothly grow/collapse. You can also log `result.height()` near `currentFrame == 0`: it is computed near zero, then raised again by this `qMax`.

src/notelistdelegate.cpp
```diff
@@ -224,11 +243,13 @@ QSize NoteListDelegate::sizeHint(const QStyleOptionViewItem &option, const QMode
     }
 
     int yOffsets = secondYOffset + thirdYOffset + fourthYOffset + fifthYOffset;
+    const int minimumHeight = minimumRowHeight() + yOffsets;
```
nuttyartist: This minimum height does not include the 25px section header for the first pinned row / first unpinned row, even though `paintLabels()` shifts those rows' content down by that header. Once font metrics exceed the old fixed budget, those first rows can still clip/overlap. The same issue applies to the editor floor in `NoteListDelegateEditor::recalculateSize()` because `tagListTop()` includes the 25px header but `minimumRowHeight() + tag height` does not.

Suggested fix: include the same header offset in the minimum-height calculation for first pinned/first unpinned rows, preferably via a shared helper so delegate paint, tag placement, and row sizing stay in sync.

How to test: on Windows at 150% or 175% scaling, create a first pinned note and a first unpinned note, with tags, and also check `All Notes` so the folder line is visible. Inspect those first rows for clipped text, clipped tag chips, or overlap.

nuttyartist (commented):
Thanks for tackling the Windows scaling issue. I left two spots where the sizing math still needs to account for existing behavior.

AJ0070:
@nuttyartist Thank you for the comprehensive review! I've addressed both issues.

nuttyartist:
I found one follow-up visual regression after the latest sizing fix: when the first pinned note is selected, the selected background now bleeds upward into the `Pinned` section header.

Before:
Screenshot 2026-06-19 at 14 30 54
After:
Screenshot 2026-06-19 at 14 31 53

The likely cause is that `sizeHint()` now correctly includes `sectionHeaderHeight(...)`, but `NoteListDelegate::paintBackground()` still fills the entire `bufferRect` for selected/hover/default rows before drawing the pixmap back. The editor path already handles this correctly by moving `bufferRect` down before filling:

```cpp
bufferRect.setY(bufferRect.y() + note_list_constants::SECTION_HEADER_HEIGHT + fifthYOffset);
```

Suggested fix: mirror that behavior in `NoteListDelegate::paintBackground()` for first pinned / first unpinned rows, while preserving the collapsed pinned-header-only case. In other words, the background fill rect should start below the `Pinned` / `Notes` header when the row includes both a section header and note content.

A shape like this should be close:

```cpp
int headerOffset = sectionHeaderHeight(index, *model);
if (headerOffset > 0 && !(m_view->isPinnedNotesCollapsed() && model->isFirstPinnedNote(index))) {
    int fifthYOffset = 0;
    if (!m_view->isPinnedNotesCollapsed() && model->isFirstUnpinnedNote(index)) {
        fifthYOffset = note_list_constants::LAST_PINNED_TO_UNPINNED_HEADER;
    }
    bufferRect.setY(bufferRect.y() + headerOffset + fifthYOffset);
}
```

How to test:

- Expand pinned notes and select the first pinned note. The blue selection background should begin below the `Pinned` label, not behind it.
- Select the first unpinned note when the `Notes` header is visible. The `Notes` header should also keep its normal background.
- Collapse pinned notes and make sure the pinned header-only row still paints normally.

AJ0070:
@nuttyartist I have fixed the follow-up visual regression.

AJ0070:
Let me know if any further changes are required.

AJ0070:
@nuttyartist I have addressed all of the reviews, please let me know if any further changes are required.

nuttyartist:
@AJ0070 Sorry for the delay, I'll have available time to properly review your changes only at the weekend. I did find multiple issues with your changes already, so I want to make sure to test it again properly. Stay tuned.

AJ0070:
@nuttyartist did you get the chance to review it?

nuttyartist:
One issue still looks off around the first unpinned row when pinned notes are expanded.

`sizeHint()` accounts for `LAST_PINNED_TO_UNPINNED_HEADER`, but `bufferSizeHint()` does not, and `paintBackground()` now also shifts `bufferRect` by `sectionHeaderHeight + fifthYOffset`. Since the pixmap is later drawn bottom-aligned into the row, the first unpinned row can end up with the background/body region shifted relative to the actual row height.

I can see this in hover/selection/drag states, and it also affects the no-tags path: an unpinned note without tags directly under the `Notes` header has incorrect top padding.

Repro:
1. Have at least one pinned note and one unpinned note.
2. Keep pinned notes expanded.
3. Make the first unpinned note sit directly under the `Notes` header.
4. Test that first unpinned row both with and without tags, then hover/select/drag over it.

Expected: the row background, top padding, separator, text, and tag editor all start consistently below the header/gap.
Actual: the first unpinned row can get a shifted/extra top gap, especially visible in background states and also on no-tag rows.

Suggested direction: centralize the header/gap/body offset calculation so `sizeHint()`, `bufferSizeHint()`, `paintBackground()`, and the editor path all agree. In particular, avoid mixing a full-row `sizeHint()` with a differently-sized/shifted paint buffer for the first unpinned row.

nuttyartist:
There's still a problem with the first unpinned note when there are pinned notes.

Before:
Screenshot 2026-07-03 at 12 51 46
After:
Screenshot 2026-07-03 at 12 52 31

It takes me a long time to spin up a VM every time and test your changes, so, please, if you can test your commits carefully next time, I'll appreciate that. I'm not going to accept PRs that haven't been carefully tested. in the end, people are going to use this, it should look and behave well.

I'm leaving comments on the PR using an agent.

src/notelistdelegate.cpp
```diff
@@ -278,11 +309,13 @@ QSize NoteListDelegate::bufferSizeHint(const QStyleOptionViewItem &option, const
     }
 
     int yOffsets = secondYOffset + thirdYOffset + fourthYOffset;
+    const int minimumHeight = minimumRowHeight() + sectionHeaderHeight(index, *noteListModel) + yOffsets;
```
nuttyartist: This path needs the same first-unpinned gap that `sizeHint()` uses. Without it, the paint buffer can be shorter than the actual row rect, which is what makes the first unpinned row's background/top padding drift, including the no-tag case.

```suggestion
    int fifthYOffset = 0;
    if (noteListModel->hasPinnedNote() && !m_view->isPinnedNotesCollapsed() && noteListModel->isFirstUnpinnedNote(index)) {
        fifthYOffset = note_list_constants::LAST_PINNED_TO_UNPINNED_HEADER;
    }

    int yOffsets = secondYOffset + thirdYOffset + fourthYOffset + fifthYOffset;
    const int minimumHeight = minimumRowHeight() + sectionHeaderHeight(index, *noteListModel) + yOffsets;
```

src/notelistdelegate.cpp
```diff
@@ -301,7 +334,16 @@ void NoteListDelegate::paintBackground(QPainter *painter, const QStyleOptionView
...
     auto const *model = static_cast<NoteListModel *>(m_view->model());
-    if (model->hasPinnedNote() && model->isFirstPinnedNote(index) && static_cast<NoteListView *>(m_view)->isPinnedNotesCollapsed()) {
+    const bool isCollapsedPinnedHeader = m_view->isPinnedNotesCollapsed() && model->isFirstPinnedNote(index);
+    const int headerOffset = sectionHeaderHeight(index, *model);
+    if (headerOffset > 0 && !isCollapsedPinnedHeader) {
+        int fifthYOffset = 0;
+        if (!m_view->isPinnedNotesCollapsed() && model->isFirstUnpinnedNote(index)) {
+            fifthYOffset = note_list_constants::LAST_PINNED_TO_UNPINNED_HEADER;
+        }
+        bufferRect.setY(bufferRect.y() + headerOffset + fifthYOffset);
```
nuttyartist: This offset should stay paired with the size-hint gap above. Naming the first-unpinned gap here makes the dependency easier to spot; ideally this calculation would become a tiny shared helper used by `sizeHint()`, `bufferSizeHint()`, and painting.

```suggestion
    const bool isCollapsedPinnedHeader = m_view->isPinnedNotesCollapsed() && model->isFirstPinnedNote(index);
    const bool isFirstUnpinnedWithGap = model->hasPinnedNote() && !m_view->isPinnedNotesCollapsed() && model->isFirstUnpinnedNote(index);
    const int bodyTopOffset = sectionHeaderHeight(index, *model)
            + (isFirstUnpinnedWithGap ? note_list_constants::LAST_PINNED_TO_UNPINNED_HEADER : 0);
    if (bodyTopOffset > 0 && !isCollapsedPinnedHeader) {
        bufferRect.setY(bufferRect.y() + bodyTopOffset);
    }
```

AJ0070:
Thanks for the review, pushed a fix for the first-unpinned-row spacing, properly tested this time.

**Why not the inline suggestion verbatim:** it grows `bufferSizeHint()` *and* shifts `bufferRect` by the same offset, but the buffer is drawn bottom-aligned by `drawPixmap()`, so the two overlap. I kept the buffer body-only and let the existing bottom-alignment reserve the header/gap, which preserves `rowHeight − bufferHeight == sectionHeaderHeight + firstUnpinnedGap`. `sizeHint()`, `bufferSizeHint()` and `paintBackground()` now all derive that offset from the same two helpers.

**Measured** — highlight band top/height in logical px, first unpinned row without tags, pinned expanded (expected offset 35 = 25 header + 10 gap):

| build | 100% | 150% | 175% |
|---|---|---|---|
| `master` | 132 / 72 | 132 / 72 | 132 / 72 |
| `6919e46` (what you reviewed) | 148 / 68 | 148 / 68 | 148 / 68 |
| this PR | 138 / 78 | 138 / 78 | 138 / 78 |

The regression started 10px low and lost 10px of height. Fixed, and identical at every scaling level.

comparison_175

comparison_150

comparison_aligned

Also verified: first pinned (25), ordinary rows (0), pinned collapsed (25, no gap), All Notes, hover/selected, and the insert animation still collapses to height 0 — so the minimum-height floor is correctly skipped for animated rows. Not tested: drag, since it enters a modal `QDrag::exec()` loop; its indicator rects derive from the same `bufferRect`.

**On #164 itself:** display scaling alone can't reproduce it on Qt 6 — `dpr` changes but `logicalDpi=96` and `titleFontHeight=17` at 100/150/175%, so rows and text scale together. The real trigger is font metrics exceeding the hardcoded 70/106px row budget (font substitution, DPI-unaware path, larger system text). Raising the font size reproduces it on `master` — the pinned note's preview line overflows and collides with the `Notes` header — and this PR fixes it. That's what the metric-based `minimumRowHeight()` floor addresses, whichever cause is at play.

issue164

AJ0070:
Heads up that the red X on Windows here is pre-existing and unrelated to this PR:

```
FAILED: CMakeFiles/Notes.dir/src/customapplicationstyle.cpp.obj
Qt\6.4.3\msvc2019_64\include\QtCore/qvarlengtharray.h(377): error C3861: 'stdext': identifier not found
Qt\6.4.3\msvc2019_64\include\QtCore/qvarlengtharray.h(909): error C2065: 'stdext': undeclared identifier
```

It fails inside Qt's own header while compiling `src/customapplicationstyle.cpp` — a file this PR doesn't touch. The build stops before it ever reaches the files I changed. `master` fails identically ([run 27494239891](https://github.com/nuttyartist/notes/actions/runs/27494239891)), same job, same error.

Cause: the `windows-2025` runner image now ships Visual Studio 18 / MSVC 14.51, whose STL dropped the `stdext` namespace, and Qt 6.4.3's `qvarlengtharray.h` still uses it. Bumping the Windows job off Qt 6.4.3 (the macOS dmg job already uses 6.5.2) or pinning `runs-on: windows-2022` should clear it.

Happy to open a separate PR for that if it'd help — kept it out of this one to avoid mixing a CI change into the layout fix.

AJ0070:
@nuttyartist gentle ping!

